PR #59, opened 2016-02-11
Updated email to be a link

### LINKED ISSUE #58: New social account user can't be edited due to no name
When a user first joins through a social account, they cannot be edited from the list of users because they do not yet have a name.  It might be good to allow their email to also be a link just to avoid this.

nwalke:
Fixes #58


ssddanbrown:
Awesome, First pull request, Thanks! I'll double check and merge it in later today.

PR #60, opened 2016-02-11
Fixes #45

### LINKED ISSUE #45: Book navigation auto-scroll
It would be great if the Book Navigation pane maintained its position on the right as you scrolled down the page.

nwalke:
This might not be the best way to do this, but it certainly gets the job done.

Here's a gif of it in action: https://i.imgur.com/9mgpwHC.gif

### DIFF
--- a/resources/assets/js/pages/page-show.js
+++ b/resources/assets/js/pages/page-show.js
@@ -72,4 +72,19 @@ window.setupPageShow = module.exports = function (pageId) {
         goToText(text);
     }
 
+    // Get current tree's width
+    var bookTreeWidth = $(".book-tree").width();
+    // Get header height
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks again for the contribution! I changed the tree-list to be more of a full-screen sidebar as there were some issues when the tree list was taller than the page. I also changed some of the JS so it took into account the mobile widths and other breaking factors.

Here's a video of it now:
https://gfycat.com/TintedLastingEuropeanpolecat

I'm not fully set on it but I'll see how it feels in daily use.

nwalke:
Ah, perfect.  I also realized my version was setting the CSS on every scroll action... probably not the best.

PR #63, opened 2016-02-20
Update PageRepo.php

AkibaWolf:
Fix encoding problem.

By default DOMDocument::loadHTML treats a string as being encoded with ISO-8859-1. This causes a problem with saving cyrillic pages' text that becomes completely unreadable (like ÐÑÐ¾Ð²ÐµÑÐºÐ° instead of normal symbols).


ssddanbrown:
Awesome, Sorry about this issue and thanks for fixing it :+1:

AkibaWolf:
You're welcome!

Issue #54, opened 2016-02-04
User Groups and Permissions


cpressland:
It'd be nice to be able to define groups of users and then define on a Page or Chapter level if they're able to view said Page/Chapter.

For example, I have two groups, "Infrastructure" & "Developers"
I have a page called "Super Secret Information"

I may want "Infrastructure" to be able to view and edit "Super Secret Information" but "Developers" won't even be aware the page exists.

However, if a member of the "Developers" group is an Admin they should always be able to see everything. This should only apply to Viewers and Editors.

tuudik:
Yeah, currently ACL is missing. Would love it! :+1:

litewhatever:
:+1:

armouredking:
Piggybacking - It'd be nice for it to also have the ability to have multiuser controls. The way Bookstack is made, it'd be great for say a fan story site. However, atm there is only the three levels Viewer / Editor / Admin. Viewers can't create books, and Editors can write to any book. It'd be nice if there was a fourth level, between Viewers and Editor ( Author? ) where they can create books and edit their own books but nobody else's.

ssddanbrown:
The current permissions system is currently database driven so adding an interface to allow custom roles is definitely possible. The tricky part will be checking permissions every time an entity is shown but I'm sure I can figure it out. I've marked this up as my next major feature to work on.

@armouredking In implementing the base issue this should be covered as you'll be able to create custom roles. I'll ensure to have one of the assignable permissions to be something like 'Can edit own content' so you'll be able to create an 'Author' user as that's a great idea.

ssddanbrown:
As of the latest release all the above is now possible. Details on the new roles & restriction systems can be found here: https://www.bookstackapp.com/blog/beta-release-v0-7-6/

Due to large changes there still may be some holes in this system. If any problems are found please open a new issue.

tuudik:
:clap: Nice job! 

Does it work also with LDAP? :)

ssddanbrown:
@tuudik No :disappointed: 

I've just opened #75 specifically for this.

Since I'm not too familiar with LDAP, what would be the ideal way of mapping LDAP groups to roles? Would assigning group DN's on each role be sufficient?

tuudik:
Maybe, if just adding possibilty to add custom filter to each role admin creates would work? @litewhatever what do you think?

litewhatever:
[...]

PR #72, opened 2016-03-03
Closes #70.

### LINKED ISSUE #70: Make search terms in quotes an exact match, not split out
Currently the search terms are split on every space. Text in quotes should not be split.

nwalke:
Added the ability to search by multi-word terms using double quotes.

### DIFF
--- a/app/Entity.php
+++ b/app/Entity.php
@@ -100,10 +100,14 @@ abstract class Entity extends Model
      */
     public static function fullTextSearchQuery($fieldsToSearch, $terms, $wheres = [])
     {
-        $termString = '';
-        foreach ($terms as $term) {
-            $termString .= htmlentities($term) . '* ';
+        foreach ($terms as $key => $term) {
+            $term = htmlentities($term);
+            if (preg_match('/\s/', $term)) {
+                $term = '"' . $term . '"';
+            }
+            $terms[$key] = $term . '*';
         }
+        $termString = "'" . implode(' ', $terms) . "'";
         $fields = implode(',', $fieldsToSearch);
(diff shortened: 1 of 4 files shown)


ssddanbrown:
@nwalke Hi again, Thanks again for all the pull requests :+1: , I'll review & merge over the weekend.
I've seen you've been fixing a lot of the issues marked for the next release, Don't worry about #54 since I've pretty much finished it on the [custom_role_system](https://github.com/ssddanbrown/BookStack/compare/custom_role_system) branch. Just got to write a good amount of tests and clean it up a little before merging it back it.

nwalke:
Yeah, I saw!  Should be good stuff.

nwalke:
Wow, I just realized this is horribly broken.  Sorry about that...   

The logic used to turn terms into strings kind of ruins what I was hoping I could easily do.  I'll fix that up.

nwalke:
Awesome!

ssddanbrown:
@nwalke Yeah, I saw the issue, That's a right pain. Since using PDO quoting for escaping doesn't seem to play well with full text searches I've added some custom escaping. I'm a little nervous about it but I've tested loads of combinations of symbols and characters against it (Trying to inject SQL) and it appears to be work fine. I'll still need to test it against mariaDB before release to ensure there's no differences there.

I also learnt that MySQL won't fulltext search exactly against a phrase. Searching `"the tree"` would match against `the tree` and `the old tree`. Any number of words could be between them. I've added some `LIKE` queries which are used if any quoted terms are in the search query to resolve this. I found that doing a fulltext match then a `LIKE` query has minimal performance impact and is still way faster than only using a `LIKE` query.

PR #73, opened 2016-03-03
Closes #55.  Allows you to set the primary color.

### LINKED ISSUE #55: Ability to update the site Primary Colour
Provide a preference on the settings page that allows admins to update the primary colour on the site, which is currently Blue by default.

nwalke:
This certainly should not be seen as the final implementation of this.  I imagine that requests like this will happen for many other colors in the system, it might be good at some point to allow them to upload their own stylesheet to be used.  But then again, maybe this suffices.


ssddanbrown:
@nwalke Thanks again for the pull request. I switched out the color picker library since jscolor is GPLv3 licensed which I believe is not compatible with BookStack's MIT licence. I also moved the color rgb logic from the controller to be calculated on the front-end since the new color picker library had that built in and it keeps the controller cleaner. Now the lighter color shade is sent through as a hidden setting.

I totally agree with being able to use own stylesheets but for now the current solution will be very nice for people that don't want to get involved with editing css. At some point soon I'm planning to have a setting to insert any custom content in the head (For styles or things like analytics scripts).

Issue #48, opened 2016-01-19
Custom page attributes & values to add an extra dimension of organisation


Di2g10:
Currently you have Books, Chapters and Pages with a chapter only able to sit in one book and pages only able to sit in one chapter. However this leads to the requirement for a rigid and constricting structure as you can only break up articles on one dimension. A Dimension being say Client or Platform or Article type. However in a ideal world a page would want to be split by all these dimensions which are then standardised and enforced such that you can filter by any of these for all the related articles or you can search for combinations.

I can suggest two methods to achieve this
You could introduce the ability for the same page to be a member of multiple chapters potentially across different books. (simpler to impliment harder to enforce)
Redefine how pages are organised so that you have tag groups and tags such that every article should have a tag in each group. it may be that a tag group has a hierachical structer in itself. I.E. you may have internal vs client, then have to select specific client if applicable. Or is you select the platform then only relevant to clients on that platform.

ssddanbrown:
Rather than a simple tagging/labelling system I think it will be worthwhile implementing a full attribute => value system so that pages can be assigned multiple attributes and each has it's own value. This way people can categorize content in a way that also give extra value to the page it's put upon rather than just a generalised tag.

For example, you might have a book of 'meetings' and then an attribute => value structure like this:
```
Meeting 1 (Page in BookStack)
-- Meeting Subject: Efficiency
-- Meeting Leader: Dennis
-- Meeting Location: Conference Room A

Meeting 2 (Page in BookStack)
-- Meeting Subject: Customer Service
-- Meeting Leader: Dee
-- Meeting Location: Conference Room A
```

I'm will alter this title of this issue to better reflect this system.

nwalke:
It would be pretty neat to be able to require certain attributes on all pages under a book.  So in your example, I would require "Meeting Subject" at the book level and then every page under it would have to have that.

Issue #95, opened 2016-04-06
Support more sublevel in chapters


ducheneo:
It should be interesting if we could use at least one more level of sub-chapter in a book.
Collapsible sections are also interesting.

ssddanbrown:
Hi @ducheneo, I agree with collapsible sections, They're already open and requested on #78.

The extra level of grouping is an interesting topic. I **would rather not** add this feature and here's why:

Initially when building BookStack I allowed a system of highly nested pages. I was following the system that confluence used. From a development perspective this was tricky to support, Not impossible by any means but added extra complication whenever there was a listing or when working out relation trees. 

From a user perspective things were fairly hard to navigate. Allowing a high level of nesting means you can only show part of a grouping, or tree structure, at a time which meant you only see a snapshot of a whole and it becomes easier for content to stay hidden. It also means there was no standard grouping system or clear separation of levels. 

At that point I changed BookStack to a 2 tier grouping, Books & Chapters. I very much like this level of grouping as it can be related to the way we store content in books, It's familiar and simple while allowing a good level of organisation. It also forces you to keep the wider organisation of the content from going out of control. I feel that adding an extra layer of chapters would break this simple relation to books. On top of that there will always be someone that wants an extra layer, no matter how many layer are implemented.

I do understand the need for an extra dimension of organisation/categorisation through which will be added by the attribute-value system proposed in #48.

I'll keep this issue open and also I'm open to further discussion as the community may have a very different opinion to me but I do really like the core Book/Chapter/Page structure of BookStack.

nwalke:
We definitely like the Book/Chapter/Page structure.  This is why we're using BookStack.

ducheneo:
Hi Dan, I can understand your point of view.
The point is that the nature of our documentation is deeply hierarchical, and 3 levels of grouping would be more appropriate. The display of the navigation tree could remain unchanged, by replacing the collapsible with "number of page" with the sub-chapter title. And if most people find the actual structure fine, adding the extra level as an option could satisfy everyone (but certainly tricky to implement).

cpressland:
Well, this is an awkward one. While I love the idea of being able to have chapters within chapters, I imagine it'll quickly become as overly complex as Sharepoint or a Filesystem. The current system forces the people and teams adding content to BookStack to be creative, rather than ending up with one page in a large number of chapters.

```
[book]Products > [chapter]SomeApp > [page]SomeContent
[...]

PR #110, opened 2016-05-06
Attribute System

ssddanbrown:
As per #48. To be created as an AJAX-based system and controlled with angular in the page editor.
Attributes will only be able to be assigned to pages initially but future Chapter/Book support will be considered in build.
### Build task list:
- [x] Initial API build
- [x] API testing in place
- [x] Angular-based UI on page editor screen
- [x] Attribute/value search support


KnightNine:
Is there documentation on how to use this feature?
There's no mention of it on the [v0.10.0 beta release page](https://www.bookstackapp.com/blog/beta-release-v0-10-0/).

ssddanbrown:
@KnightNine it's mentioned quite a lot under the "Page Tagging" title

KnightNine:
oh, I thought this was a separate feature to tagging that would help with categorization in the "folder" view (where you navigate through shelf>book>chapter) as you offered this as a potential solution to #95 . Tags aren't viewable unless you're already viewing the page/chapter/book the tags are attached to.

I would suggest to...:
tags

But it wouldn't really help with easily seeing which pages/chapters/books use the same tags without searching for the specific tag, and likely becoming less and less easily viewable the more tags that are added among your pages.


I feel like Tagging only really helps if you already know the tag you're looking for and you have a lot of pages which relate to each other on certain topics. It doesn't really help when trying to organize the branches of a subject and have it be easily navigated.

> I do understand the need for an extra dimension of organisation/categorisation through which will be added by the attribute-value system proposed in https://github.com/BookStackApp/BookStack/issues/48.

True that this is an extra dimension of categorization but not for the folder view which #95 is addressing.

PR #118, opened 2016-05-25
Making sure MyISAM is set

rovox:
This will help fix the bug where some people get the message FULLTEXT is not supported. until mysql 5.6 innodb does not support FULLTEXT by using MyISAM for the tables that require FULLTEXT this issue will be fixt. (This will only work for new installations because you can't change the type after a table is created.

### DIFF
--- a/database/migrations/2015_07_12_114933_create_books_table.php
+++ b/database/migrations/2015_07_12_114933_create_books_table.php
@@ -13,6 +13,7 @@ class CreateBooksTable extends Migration
     public function up()
     {
         Schema::create('books', function (Blueprint $table) {
+	    $table->engine = 'MyISAM';
             $table->increments('id');
             $table->string('name');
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Hi @rovox, Thanks for the pull request.
I'm not sure forcing a DB engine is the correct way to go about this. This is only a workaround for MySQL 5.5 which is getting fairly old now. I don't think a specific engine should be forced for everyone else.
MySQL 5.6 onwards (Or MariaDB 10+) is fairly accessible on the majority of operating systems now.

rovox:
Hmm right now centos 7 is using a mariadb version that does not support fulltext. Maybe forcing is a little bit overkill. But a warning might be handy. :)

ssddanbrown:
I've now added some checks to see whether the MySQL version is 5.5 before forcing the database engine. Code now merged in.

PR #139, opened 2016-07-07
html diff in revision view

younes0:
Result:
image

### DIFF
--- a/app/Http/Controllers/PageController.php
+++ b/app/Http/Controllers/PageController.php
@@ -12,6 +12,7 @@ use BookStack\Repos\ChapterRepo;
 use BookStack\Repos\PageRepo;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Views;
+use Icap\HtmlDiff\HtmlDiff;
 
 class PageController extends Controller
 {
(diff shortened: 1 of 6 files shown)


ssddanbrown:
@younes0 Thanks for this. Unfortunately the library used, `icap/html-diff`, is licensed under GPL2+ which, as far as I'm aware, is not compatible with the MIT licence used in BookStack :(

younes0:
@ssddanbrown Ooops sorry for that. I'll push with this lib instead https://github.com/gathercontent/htmldiff 

Note that this one will require the php-tidy ext

younes0:
done!

ssddanbrown:
Hey @younes0,
Thanks again for this feature. Just want to explain why I haven't merged this in yet. It's just the additional `php-tidy` extension requirement that's stopping me merging at the moment. I didn't want to introduce a breaking update right away so I'm seeing if there's some other things that can be added to BookStack that will require potential breaking direct update compatibility.

younes0:
Hello @ssddanbrown 
As far as I remember, I couldn't find a MIT licenced package that didn't require `php-tidy`

ssddanbrown:
Yeah, I also had a look around but could not find a better library. Also tried making the diff code myself but failed pretty miserably.

There's going to be some compatibility changes/breakages in the BookStack release anyway so now is a good time to sneak in the `php-tidy` requirement. Should be handy to have tidy in other areas of BookStack anyway.

Thanks again for your work on this.

PR #140, opened 2016-07-07
edit summary

younes0:


ssddanbrown:
@younes0 Again, Thank you for this. Due to the silly way I was saving revisions the summaries were being attached to the previous page versions so I've updated how revisions are handled on save. 

I changed the wording from 'Summaries' to 'Changelog' on the user side just to be a little more descriptive of what they need to enter.

younes0:
@ssddanbrown I didn't even notice that behavior! just pulled changes, that looks neat, thank you

PR #143, opened 2016-07-09
shorter amazon S3 url

younes0:


ssddanbrown:
@younes0 Thanks again! I had to change this up a little. When a s3 bucket has periods in, the `'https://' . $storageDetails['bucket'] . '.s3.amazonaws.com'` url format would break due to https issues so I added a simple check to see if the bucket contains periods.

younes0:
@ssddanbrown nice catch, I'm surprised that a bucket name can contain dots

PR #144, opened 2016-07-09
set uploaded images visibility public (relevant for S3 storage)

younes0:


ssddanbrown:
@younes0 Awesome :smile:, Thanks for this, Did not know flysystem could control visibility like this.

PR #154, opened 2016-07-21
Added issue template

nwalke:
I would love to help out troubleshooting BookStack issues, but often times there is a clear lack of information available in the bug reports.  Simple things such as PHP and MySQL versions aren't listed which creates noise on the issue for someone to have to ask.  

This is probably by no means "complete" and you should feel free to edit this to include/exclude whatever information you want.

Just trying to help out ❤️


ssddanbrown:
@nwalke Great idea, Thanks

PR #183, opened 2016-09-08
[Enhancement] - Installation via Docker / Docker Compose

dylanrhysscott:
Hi Dan @ssddanbrown ,

I've forked this project to dockerise it on the recommendation of Jonathan Tawn. I thought I'd chuck a pull request for it. This is for the release branch due to the need for a manifest.json. However these changes will allow Bookstack to be built and deployed in a single command making the process a lot easier

## With Docker

Docker - [https://www.docker.com/](https://www.docker.com/)

* Build the image by running ```docker build -t bookstack .```
* Create a DB container by running ```docker run -d --name mysql -e MYSQL_DATABASE=bookstack -e MYSQL_ROOT_PASSWORD=root mysql```
* Create a Bookstack container by running ```docker run -d --name bookstack -e DBNAME=bookstack -p 80:80 --link mysql:mysql bookstack

The install will be available on ```http://localhost```

## With Docker Compose

Docker Compose - [https://docs.docker.com/compose/](https://docs.docker.com/compose/)

* Run ```docker-compose up```

## Configuration

* All of the envs for Bookstack can be configured by editing the example ```.env``` file prior to build. With the exception of **the database details**. These are handled at run time
* The database names and passwords can be configured my passing different env variables from the commands or by editing ```docker-compose.yml```. They must be consistent between the two containers to work correctly

## Base Images used

* MySQL - [https://hub.docker.com/_/mysql/](https://hub.docker.com/_/mysql/)
* PHP w/ Apache Tag [https://hub.docker.com/_/php/](https://hub.docker.com/_/php/)

## Autobuilds

Autobuilds can be configured for your project by leveraging Dockerhub - [https://hub.docker.com/](https://hub.docker.com/)

Looking forward to your feedback!

Cheers

Dylan

PR #184, opened 2016-09-08
[Enhancement] [Reopen] - Installation via Docker / Docker Compose

dylanrhysscott:
Reopen of PR #183 - See for further details


ssddanbrown:
Hi @dylanrhysscott,
Thanks for your work on this! It's really appreciated and awesome to see people contributing to BookStack.

Unfortunately I won't be merging this pull-request and here's why:

1. I don't want to add install-specific files into the core BookStack repo. This would add a few extra root-level files that don't really need to be in this repo. Ideally I'm trying to keep all dev-ops type of code separated from the BookStack app code.
2. There's already a [pretty solid docker implementation](https://github.com/SolidNerd/docker-bookstack) of BookStack on the go.
3. I won't accept pull requests to the release branch since it's reserved for only versioned releases, No development.

Let me know if I've got the wrong end of the stick and your docker implementation works differently to the one mentioned above or adds to it. Sorry it's not clear that an implementation currently exists. I've been meaning to update the installation docs for a while.

dylanrhysscott:
Hi Dan @ssddanbrown,

Thanks for getting back to me - being a big fan of open source I do like to contribute where I can. Really appreciate the feedback - wanted to address the points as we could open a discussion about it:

1. Fair enough on this point I have added some extra files for the docker build and I could probably abstract this further / clean this up. However with regards to install specific files I'd argue that due to the interoperable and immutable nature of docker containers they aren't quite install specific as in theory the container will run on any platform. I probably should abstract the apache confs for example into the container away from the source repo to keep it clean and not tied to a given web server / DevOps tools. 
2. My approach to dockerising Bookstack appears to deliver the same level of the repo you mentioned. However that one is a lot cleaner and I could add the git clone as part of the build process to slim things down. I'll take a look at this for some inspiration.
3. I chose the release branch as the install would not complete successfully without a build manifest.json and doing the above with Git would likely solve this issue.

As an extra thought it would be brilliant to have a Docker install baked into your official repo for coherence and giving more options through the official repo. Also by including the Docker stuff directly into your repo and linking this with automated builds it would streamline the update process a lot more and allow easier upgrades with the added bonus of when you push a new version a self contained build is handled for you. I'd love to help make it happen directly in the official repo and would gladly take your advice on how to meet the standard for you to merge it. That is of course if it's something your interested in!

Thanks

Dylan

PR #197, opened 2016-09-23
Database port changeable via DB_PORT env variable

### LINKED ISSUE #196: Database port env variable
### For Feature Requests

Desired Feature:
Change the database port via env variable.

Expected Behavior:
Be able to change the database port via DB_PORT variable

Actual Behavior:
This env variable is not implemented

dionysiusmarquis:
Hey,

I needed to change the database port and this changes will make it possible :)

fixes #196

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -71,6 +71,7 @@ return [
         'mysql' => [
             'driver'    => 'mysql',
             'host'      => env('DB_HOST', 'localhost'),
+            'port'      => env('DB_PORT', 3306),
             'database'  => env('DB_DATABASE', 'forge'),
             'username'  => env('DB_USERNAME', 'forge'),
             'password'  => env('DB_PASSWORD', ''),
@@ -95,6 +96,7 @@ return [
         'pgsql' => [
             'driver'   => 'pgsql',
             'host'     => env('DB_HOST', 'localhost'),
+            'port'     => env('DB_PORT', 5432),
             'database' => env('DB_DATABASE', 'forge'),
             'username' => env('DB_USERNAME', 'forge'),
             'password' => env('DB_PASSWORD', ''),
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @dionysiusmarquis,
Thanks for raising this issue.

The database port can currently be set by specifying it on the `DB_HOST` env variable, For example:

```
# Connect to database locally on port 6060
DB_HOST=127.0.0.1:6060
```

dionysiusmarquis:
In my special case I wanted to use it with https://hub.docker.com/r/solidnerd/bookstack/ and the whole docker start routine got messed up with the host:port approach. The [laravel port option](https://laracasts.com/discuss/channels/laravel/how-to-specify-a-different-database-port?page=1) was the only one working for me in this scenario.

dionysiusmarquis:
The latest changes for https://hub.docker.com/r/solidnerd/bookstack/ (https://github.com/solidnerd/docker-bookstack/pull/6) depending on this pull request. Any chance to get it merged?

ssddanbrown:
@dionysiusmarquis Yeah but I just want to test why using host:port was breaking on docker first. Also, once merged, it will be a small while before the next release is put together anyway.

This pull request could also break the configuration of existing setups so I need to test that out to find any consequences.

dionysiusmarquis:
It doesn't work because the shell script waits for the database using `${DB_HOST}:${DB_PORT}`. If `DB_HOST` is `127.0.0.1:6060` it would end up with `127.0.0.1:6060:3306`. https://github.com/solidnerd/docker-bookstack/blob/master/docker-entrypoint.sh#L87

ssddanbrown:
Hi @dionysiusmarquis,
From looking at the issue here I don't think a new `.env` variable needs to be made available and no changes are needed to the BookStack core code. This issue is really only specific to the docker script and how it handles the environment variables.

I'll take a look at updating the docker entrypoint script.

ssddanbrown:
Since solidnerd/docker-bookstack#9 has now been merged and made it into a release this is no longer required as the port can be defined via the host and is now compatible via the docker solution.

PR #232, opened 2016-11-12
Added initial translation into German (formal)

robertlandes:
This translation is based on the current english language files. It is a
transaltion into German (formal) language.  A language selector on the
Settings page would be a nice addition.

### DIFF
--- a/resources/lang/de/activities.php
+++ b/resources/lang/de/activities.php
@@ -1,0 +1,40 @@
+<?php
+
+return [
+
+    /**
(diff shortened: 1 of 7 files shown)


ssddanbrown:
@robertlandes Thanks again for this. Once the languages are a bit more complete then I'll add a language selector to the settings area (And ideally to a user's preferences as well) but for now, while in development, I have made the language configurable via the `APP_LANG` .env variable:

```
# .env
APP_LANG=de
```

PR #236, opened 2016-11-15
Update Ldap.php to allow LDAP over SSL/TLS connections

fredericmohr:
This is a very crude workaround to allow LDAPS connections. A better solution would be the one explained in the comments.

### DIFF
--- a/app/Services/Ldap.php
+++ b/app/Services/Ldap.php
@@ -18,6 +18,24 @@ class Ldap
      */
     public function connect($hostName, $port)
     {
+        /*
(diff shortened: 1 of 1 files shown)


ssddanbrown:
@fredericmohr Thanks for this and sorry it's taken me a while to getting round to merging your pull request.

I have followed your comments and updated the ldap in 598b07b53d5e3aa5ed43dad419ce2499c5493943 so that a protocol can be optionally specified without breaking. Port 636 or 389 will then be used, depending on protocol type, unless a port is specified by the user.

PR #255, opened 2016-12-19
Add Carbon localization support

olexus:
Synchronize localization for Carbon with main app


ssddanbrown:
@olexus Awesome 😄 

Thanks for this, Will merge in now.

PR #279, opened 2017-01-18
Brazilian Portuguese Localization

NakaharaL:
Translated files for messages from English to Brazilian Portuguese


ssddanbrown:
@NakaharaL All looks good! Once again, Thank you so much for doing these translations. I'll merge this in now and it will be in the next release. 👍

NakaharaL:
You very welcome, Dan. 

Can I ask you something? Is the version 0.13.1 already capable of using all the localization files? I ask this because the version that I have installed on my server, although is showing that is the 0.13.1, won't show the system messages translated. 

I wonder what is going on, since I have seen the code within github and all is correct.

Thanks!

ssddanbrown:
@NakaharaL v0.13.1 does not include the translation text unless you've copied your work into the lang folder. You can then make language to show by setting the `APP_LANG` variable in the `.env` file, For example:

```
APP_LANG=ptbr
```

Setting the above would tell BookStack to look in the `resources/lang/ptbr` folder for translations. Again, That will work on v0.13.1 if you add in the translations yourself.

When v0.14 of BookStack is released (Hopefully this weekend) the language files will be included in the update and a language preference dropdown will be available in the 'Update profile' area of BookStack.

PR #356, opened 2017-04-06
[WIP] views: add support of highlight on live preview

Kdecherf:
This PR adds the support of code highlight on the live preview.

### DIFF
--- a/resources/views/pages/edit.blade.php
+++ b/resources/views/pages/edit.blade.php
@@ -22,5 +22,6 @@
     
     @include('components.image-manager', ['imageType' => 'gallery', 'uploaded_to' => $page->id])
     @include('components.entity-selector-popup')
+    @include('partials/highlight')
 
 @stop


ssddanbrown:
@Kdecherf Thanks again, Will merge before next release. Easier solution to that than i expected 😃

ssddanbrown:
Hi @Kdecherf, Just gave this a test. Seems like this will work in only the markdown editor on initial load, Any updates will then remove highlighting. Also requires the double 'document.ready' as per your other pull request. 

I think it would be best to either have it fully highlighted or not at all otherwise the change is a bit jarring.

Kdecherf:
Hello @ssddanbrown, I managed to fix auto-refresh of hljs when typing text. This feature is still limited to the markdown editor and could have a performance impact on huge documents (_not tested_).

Do you still want to merge this PR? (_I did not push my changes yet_)

Kdecherf:
Well… In the current configuration it has a real performance impact on documents with tens of code blocks.

This feature could be improved by redrawing only modified parts of text in the markdown live display. Currently BookStack redrawns the whole document on each keystroke.

ssddanbrown:
@Kdecherf Feel free to push your changes, I'd like to see how it performs. Maybe we could add a throttle for hljs renders to prevent re-rendering on every small change.

I was planning at some point soon to update the markdown editor from a textarea to something like codemirror so maybe then we can do something way more efficient as we'll be able to update the preview 'per block'.

Kdecherf:
This PR does not seem relevant anymore as the Markdown editor has changed in the recent releases, closing it.

PR #358, opened 2017-04-11
Add Slovak translation

jendrol:


ssddanbrown:
@jendrol Thank you so much for these translations. I will merge in shortly before the next release.

PR #362, opened 2017-04-14
Queue confirmation emails.

DaneEveritt:
Implements Laravel's queue abilities into the email notification job. Should not affect existing installations that are not using queues as the environment file defaults to `sync`.


ssddanbrown:
@DaneEveritt Thanks for this. I'm not really supporting any queue functionality right now but I'm happy for any Laravel experienced users to set up a queue if they want.

PR #363, opened 2017-04-17
Add APP_LOGGING

solidnerd:
This will add an variable for logging types to make it easier to define outside via .env.

Signed-off-by: solidnerd <niclas@mietz.io>


ssddanbrown:
@solidnerd Good idea, Thanks

PR #365, opened 2017-04-19
Enhanced Book Name slugification

ecappa:
I had some issues with accented characters in Book name. Here is a stronger slugification routine.
Perhaps should be more callable generic service ?

### DIFF
--- a/app/Repos/EntityRepo.php
+++ b/app/Repos/EntityRepo.php
@@ -675,17 +675,247 @@ class EntityRepo
         $this->permissionService->buildJointPermissionsForEntities($collection);
     }
 
-    /**
+/**
      * Format a name as a url slug.
      * @param $name
      * @return string
      */
     protected function nameToSlug($name)
     {
+        $name = $this->remove_accents($name);
+
         $slug = str_replace(' ', '-', strtolower($name));
         $slug = preg_replace('/[\+\/\\\?\@\}\{\.\,\=\[\]\#\&\!\*\'\;\:\$\%]/', '', $slug);
         if ($slug === "") $slug = substr(md5(rand(1, 500)), 0, 5);
         return $slug;
+    }
+
+    /**
+     * Checks to see if a string is utf8 encoded.
+     *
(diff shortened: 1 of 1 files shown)


Kdecherf:
Which issue did you encounter?

What about using `iconv()` to convert strings into `ASCII/TRANSLIT`?

ssddanbrown:
Hi @ecappa, Thanks for the pull request but I cannot merge this as the code seems to be from Wordpress which is GPL licensed so including this code will cause license conflicts.

Upon that, I'd rather work to fix the issue you experienced than work around it. I'm going to close this pull request because of the above but feel free to open a new issue with the problem you found.

ecappa:
Hi @ssddanbrown,
Ok, I understand, I didn't know it was a code from Wordpress... Thanks for the information.
My bug was a jammed slug link when the book title contains accents.
For exemple try to name a book '_Mon livre en Français avec des accents é à oû où et des caractères que les chanceux d'anglophones n'ont pas!!_'
Let me know if it is not clear.

ssddanbrown:
@ecappa Created book with that exact name and didn't have a problem:

image

Maybe some required server dependencies? Not full sure what would effect something like this. Maybe MySQL charset or something PHP needs?

Can you provide any details on your setup? OS, PHP Version, MySQL version etc...

ecappa:
Hi, I could reproduce the bug (see screenshots)
My stack is perhaps a bit particular :
- IIS 7
- PHP 5.6.30
- MySQL : 'version', '10.1.22-MariaDB'
- MySQL charset was set to utf_8_general_ci

ssddanbrown:
@ecappa I'm not too familiar with IIS but have you got one of the following hotfixes installed?:

* https://support.microsoft.com/en-gb/help/2277918/fix-a-php-application-that-depends-on-the-request-uri-server-variable-may-fail-when-a-request-whose-url-contains-utf-8-characters-is-sent-to-iis-7.5
* https://support.microsoft.com/en-us/help/980363

Looks like IIS has had some trouble dealing with UTF8 characters in the past.

Issue #373, opened 2017-04-25
Create a simple tool for finding missing translations


ssddanbrown:
#### The current issue

Now that we have a great set of translations we need to make it easy for any translators to find strings that have not yet been updated. Currently, It's not much of an issue but as we get more translations and as new features are added with more text then translations will become further out of line. 

#### Proposed solution

It would be great to have a simple command-line php script that can be ran which outputs which translations are missing or have been removed when compared to the english text.

This script would be located in the `resources/lang` folder but would be standalone and not rely on Laravel.
I'd imagine the call signature would be something like this:

```
php resources/lang/check-translations fr
```

And the output would be something like this:

```
Checking translations for 'fr'....

Missing 2 translations strings:
entities.permissions_intro
errors.permissionJson

The following 1 translation can be removed:
common.no_activity

```

PR #384, opened 2017-05-01
Tackling Maintenance task #373 Translation Checker

diegoseso:
I have just finished this first approach to get information about missing translations, missing keys on translations array and making a base in general for better managing translations


ssddanbrown:
Hey @diegoseso, Thank you for attempting to fix this, Sorry I havn't got back to you sooner.

I've just given it a quick go and it looks great. The 'Pending to define' list looks perfect. Here as some things that would be great to amend before merging in:

* The 'items defined but pending for a translation' and 'items not longer needed' lists appears to contain the same contents? At least for FR.
* The script requires the user to run it from the `resources/lang` folder. It would be great if all directory operations were done relative to the script location (Using `__dir__`) such as on the first `if` statement.
* Since it's a standalone script, having `#!/usr/bin/env php` on the very first line would make it able to be ran directly via the command line without invoking php manually (`./check-translations.php` vs `php check-translations.php`).
* It would be ideal if the input language casing did not matter, So I could enter `fr` or `FR` to check french translations.

ssddanbrown:
No longer needed as of b167ae795e2cc4a5d7a38dde7960430abe1e50ee therefore closing.

Issue #381, opened 2017-04-30
Feature:  Switch the HtmlDiff package


mlantz:
### Switch the HtmlDiff package:

Switch to https://github.com/caxy/php-htmldiff rather than `gathercontent/htmldiff` which requires the Tidy extension. Currently homestead doesn't work out of the box nor does anything built with forge, without having to install further extensions.

ssddanbrown:
@mlantz Thanks for opening this issue and creating a pull request but unfortunately the alternative library used is under a GPL-v2 license which is incompatible with this project. I am open to removing the tidy extension requirement if a good alternative option is provided but for now, due to the license issue, I will close this issue and the pull request.

PR #390, opened 2017-05-09
Switch to Caxy\HtmlDiff to drop the tidy requirement

mlantz:
Using this package provides the HTML diff while avoiding the need for the tidy requirement which is not present in homestead or other default server setups. Concerns issue #381


ssddanbrown:
As mentioned on #381, Closing due to incompatible licensing.

PR #438, opened 2017-07-14
Support new lines for book/chapter descriptions + fix tests

Claymm:
Avoid ignoring new lines when renderring the book/chapter descriptions on their respective detailed views.

And fixing the tests too (LDAP needs to be installed before running the LDAP tests).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ addons:
 
 before_install:
   - echo "extension=ldap.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
-  - php -i "(command-line 'phpinfo()')" | grep ldap
 
 cache:
   directories:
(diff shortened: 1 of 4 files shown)


Claymm:
Victory! 🎉

ssddanbrown:
@Claymm Thank you very much for this pull request, Never even thought about multi-line descriptions.

I did make a couple of tweaks to the code. Thanks for trying to fix my LDAP testing, I altered this recently in a way that does not require to have LDAP loaded so I have removed the travis changes to keep builds speedy. Also, The Change to the titles allowed user HTML Injection so I have added some additional escaping. Both of these changes can be seen here: 21247e10d02b22c34883ab060796801867cda732

Thanks again! Will be part of the next bugfix release, v0.17.4, due in the next couple of days.

PR #446, opened 2017-07-22
French spelling

Joorem:
- fix bad french translations
- add missing translation


ssddanbrown:
@Joorem Thank you very much for your work here. Will be part of the next bugfix release, v0.17.4, due in the next couple of days.

PR #468, opened 2017-08-13
Adding environment variable references for more providers

jordankueh:
Just filling in the rest of the environment variables - Useful for getting this running in containerised deployments.


ssddanbrown:
Thank you for this pull request but, as mentioned in the docs pull request, the services in this pull request are not supported as part of BookStack and would require extra dependencies. Therefore I will close this pull request.

There services references, apart from stripe, can all alternatively be used via SMTP.

PR #474, opened 2017-08-16
Update and fix German translation

timoschwarzer:
This PR fixes many mistakes, adds missing strings and converts all HTML notations to UTF-8 (e.g. `&auml;` to `ä` and so on).


ssddanbrown:
Thank you very much @timoschwarzer for these improvements.
I'll merge them in now so they'll be in the next BookStack release, v0.18.

PR #479, opened 2017-08-19
Converting the comment system to Vue.JS

Abijeet:
This might take a week or so to do. Expected date of completion - 27th Aug.


Abijeet:
@ssddanbrown - I'm mostly done. I've run some of the test cases, and I'll run some more tomorrow, but this is ready to be merged to the master.

Please give it a look and let me know if you'd like to change anything.

AbijeetP:
@ssddanbrown - Did you get a chance to take a look at this?

ssddanbrown:
@Abijeet Looks good. Will go through this in further detail probably tomorrow morning. All seemed to function well though so I've merged it in.

I've been thinking and might change the system to remove nested comments. Will mean a single, chronological list. Replies would indicate the comment they're replying to with a link. This way there would be no need to have a reply limit. I'll toy with this change tomorrow and see if it works better for a system like BookStack.

Abijeet:
@ssddanbrown - That'll make the code and the system simpler but a little more difficult to read.

To implement this we will have to add a unique hash to each comment (I think that's already being done), so that the user can jump between messages on the page.

It could be done something like this - 
---
> In reply to [this is the page name (comment)](https://github.com/BookStackApp/BookStack/pull/479#issuecomment-326755738)

This is another comment that is a reply to the previous comment.

---
What do you think?

ssddanbrown:
@AbijeetP Yeah, That's along the lines of what I was thinking. I'll have a play about now.

PR #483, opened 2017-08-24
Japanese lang update

msaus:
Added some missing Japanese language for code editor.
Only one file  is modified.


ssddanbrown:
@msaus Great, Thank you very much for your contribution. Will merge in now to be in the next release (v0.18)

PR #490, opened 2017-08-28
Add language changing func to tinymce

msaus:
I just added language changing function for tinymce since this is not changed when I select language except for English.


ssddanbrown:
Hi @msaus, Thank you for your pull request. There are a few issues here though which mean this cannot be merged.

Firstly, You have done this work on the `release` branch which is only used for final releases. That's why a mass of commits is shown above.
All development work should be done off the `master` branch. 

Secondly, This overwrites all existing tinymce config and breaks a lot of existing functionality. The language control needs to be set within the current config rather than being replaced.

Additionally, We'd need to include the actual lang files as no translations will actually be used at the moment.

msaus:
Thank you for your suggestions.
I will check it out.

Issue #417, opened 2017-06-14
Wrong pluralization for non-english languages


jonas-koeritz:
### For Bug Reports

* BookStack Version: v0.16.3
* PHP Version: irrelevant
* MySQL Version: irrelevant

##### Expected Behavior
Correct pluralization of terms for non-english languages.

##### Actual Behavior
The profile view uses str_plural() to pluralize terms. This only works for english nouns. German nouns get pluralized the wrong way.

##### Proposed fix
Switch to [`trans_choice()`](https://laravel.com/docs/master/localization#pluralization) to provide localized pluralizations.

timoschwarzer:
+1

ssddanbrown:
@jonas-koeritz Thanks for creating this issue. View now updated, for v0.18 release, to use trans_choice. Lang files will need to be updated.

Relevant strings can be found under the keys: `entities.x_books`, `entities.x_chapters`, `entities.x_pages`.

PR #500, opened 2017-09-10
Update german translation

timoschwarzer:
This adds missing strings and covers https://github.com/BookStackApp/BookStack/issues/417#issuecomment-328340443


ssddanbrown:
@timoschwarzer Wow, That was quick. Thank you very much. Will merge now for release later today.

PR #506, opened 2017-09-12
Russian lang added

turbotankist:
Added Russian language


ssddanbrown:
@turbotankist Wow, This is fantastic, Thanks so much for your work here. Will merge in now for the next patch release.

PR #580, opened 2017-11-01
Fixes #466. Adds support for header highlighting using intersection o…

### LINKED ISSUE #466: Page Progress
Desired Feature

This is a nice to have, really not something that is a deal breaker.

As we add headers and whatnot to a page, BookStack adds page links on the right. I'd think it would be cool to have the page "follow" the user through the page. So, if I'm head section 3, the link on the right is highlighted in some way (arrow, color change, underline, whatever).

Would just be nice so I can keep track of my progress while I read :).

Note: I'm not worried about the page remembering where I left off.

Abijeet:
…bserver.


Abijeet:
There's an issue that I'm not sure how to solve - 

screenshot-127 0 0 1-8000-2017-11-02-01-16-55-784

The current header is decided based on the proximity of a heading to the top of the viewport. To be precise, if a heading is 100px within the top corner of the viewport it is marked as the current heading. Problem is with this heading at the bottom, it never gets close to 100px of the top of the viewport.

Abijeet:
@ssddanbrown - Do you think we should add a scroll event check, and if the user has reached the bottom of the page, we highlight the last header in the page nav?

ssddanbrown:
@Abijeet Sorry for the delay in my response. 

I know its not conventional by maybe, instead of guessing what the user is viewing (By what's near the top of the page), we highlight all sections currently in view? 
So, In your example screenshot, Both the bottom two items in the page nav would be highlighted?

resources/assets/sass/_grid.scss
```diff
@@ -135,7 +135,7 @@ body.flexbox {
     width: 30%;
     left: 0;
     height: 100%;
-    overflow-y: scroll;
+    overflow-y: auto;
```
Abijeet: Firefox 57 adds a scrollbar. 

bc1bc2e9-c489-425d-b210-cd3606a7c06e

Abijeet:
Like you said, it's unconventional but I've gone ahead and done that since that seemed like the "correct" thing to do. Here is how it looks now - 

download

In addition, a property in _grid.scss was causing a scrollbar to appear in **Firefox 57**. I've fixed that as well - 

bc1bc2e9-c489-425d-b210-cd3606a7c06e

ssddanbrown:
Thank again @Abijeet, Now merged.

PR #582, opened 2017-11-03
Russian slug  and Multibyte String

wowkaster:
Generate russian slug not correct if using strtolower.
Better using mb_strtolower if that function exists.


ssddanbrown:
@wowkaster Thanks for creating this pull request. I made a small change just to remove the check for `mb_strtolower` since `mbstring` is a required PHP extension for BookStack. Otherwise all good.

Issue #541, opened 2017-10-03
Disabling Comments


Ethanb00:
Desired Feature: Ability to disable/hide comments either globally and/or by page/chapter/book. 

For my purposes (Technical Documentation for a small team), the comments section, for the most part, is an unneeded resource. I totally under the utility for others, but I do not intend on using it. 

As a result, it would be nice to be able to disable it. I don't have a preference if that switch is global, by book, by chapter, or by page.  Turning it off would remove an unused a page component and clean up the UI a bit.

Thank you!

Shackelford-Arden:
I like this idea. Definitely like the idea of having book/chapter level control of comments.

I realize you want to keep this simple @ssddanbrown but would this be something that we could potentially look at adding to a **Chapter/Book Settings** menu?

salimnanji:
+1

criskolkman:
+1 for me!

faizmali:
We would greatly appreciate this feautre as well

galaxyfeeder:
+1

Abijeet:
Ok, there a few ways this can be done,

| Option                | Role                                         |
|-----------------------|----------------------------------------------|
| Global option         | Only administrator can do this.              |
| Option on the book    | People who have modify access to the book    |
| Option on the chapter | People who have modify access to the chapter |
| Option on the page    | People who have modify access to the page    |

I think row 1 and 2 should suffice. Let me know what you guys think. Once we have the requirements nailed down and @ssddanbrown has taken a look, I can submit a PR.

Ethanb00:
@abijeet: I concur. Chapter and page are likely too granular. Per-book or global rights fit much better in terms of generating forms for documentation. 

Thank you for your work!

ssddanbrown:
Yeah, I think we should keep this at a global level. Will mark up for next feature release (v0.19). @Abijeet Feel free to go ahead with this one if you'd like.

Abijeet:
Have started working on this, few things to note - 

1. Added an option under app settings - **Disable comments**
2. If this is turned **on** then, 
   1. Adding comments is disabled.
   2. Existing comments will not be shown.
   3. The entire comments section is removed and no message is shown stating that comments are disabled.

Two things are debatable here,

1. _Do we show a message stating that comments are disabled?_ Not sure if it's needed, but can be easily added.
2. _Do we show existing comments?_ I removed them to save an extra query to fetch the comments.

Shackelford-Arden:
[...]

PR #593, opened 2017-11-15
#541 - Disable comments

Abijeet:
Towards #541 

Following has been done,

1. Added an option under app settings - **Disable comments**
2. If this is turned **on** then, 
   1. Adding comments is disabled.
   2. Existing comments will not be shown.
   3. The entire comments section is removed and no message is shown stating that comments are disabled.

Two things are debatable here,

1. _Do we show a message stating that comments are disabled?_ Not sure if it's needed, but can be easily added.
2. _Do we show existing comments?_ I removed them to save an extra query to fetch the comments.

### TODO
- [x] Add test cases for this option.


Abijeet:
@ssddanbrown - I'm done with this change, please take a look and let me know if there is anything you'd like me to change/tweak.

ssddanbrown:
@Abijeet All looked great, I just renamed a variable but that's it. Thank you very much once again!

PR #598, opened 2017-11-26
Add socialite authentication for okta

lommes:
Luckily somebody created a provider for okta some days ago, so I was able to create this pull request without much effort. I only tested with php7.1 and oktapreview.com so far, since I'm not using okta in production it would be nice if someone else could give it a try. 

Documentation still missing.


s0n-:
We will test this, need to spin up another Okta app and test instance for PHP 7.1

Thanks for the work!

ssddanbrown:
Thank you @chaseconey for adding this provider to socialite.

And thanks @lommes for this pull request. I will merge in very soon for the next release. Only thing I'll probably change is to convert the SVG to be a full SVG rather than a png embedded in an svg.

ssddanbrown:
Thanks again @lommes, Tested and all works perfectly. Now merged in ready for the next release.

Any changes I made afterwards can be seen here: 90bf13c1ab0642d70e4c101c4a7d30f9403ca6ae

Basically I just re-named the the env keys to align with the other login services and made the SVG fully vector based.

PR #614, opened 2017-12-11
Updated BookController:store method to sanitize input

LookLittleJohn:
Fixed a bug where a person could mess up the router by having books with slashes in the name as well as create books with the name of the `create`, which makes a book that is totally accessible since the route `/books/create` is already used for the book creation view.

### DIFF
--- a/app/Http/Controllers/BookController.php
+++ b/app/Http/Controllers/BookController.php
@@ -72,7 +72,7 @@ class BookController extends Controller
     {
         $this->checkPermission('book-create-all');
         $this->validate($request, [
-            'name' => 'required|string|max:255',
+            'name' => 'required|alpha_dash|notIn:create|max:255',
             'description' => 'string|max:1000'
         ]);
         $book = $this->entityRepo->createFromInput('book', $request->all());


ssddanbrown:
Thank you for your pull request.

In regards to slashes in the names, These should be stripped [when URL slugs are generated](https://github.com/BookStackApp/BookStack/blob/master/app/Repos/EntityRepo.php#L556).

This pull request would prevent spaces or underscores from being able to be used within Book names which I'd prefer to allow. 

The usage of keywords is a good point. I do think we should add something to prevent clashes but I don't think it should be part of the name validation logic but instead part of the slug generation. Keywords should still be able to be used for book/page/chapter names if specified, We would just generate non-conflicting slugs.

LookLittleJohn:
@ssddanbrown Sounds good! I've got some spare time this weekend - I'll propose an update for y'all then :smile:

ssddanbrown:
Now closing this as the core issues has been mitigated through a different solution within 1a9f6764164cb730399a47969d1c33099f1682b8

PR #621, opened 2017-12-18
Adds code to allow deletion of users via cmd line.

### LINKED ISSUE #579: Need to delete all users from BookStack
We are migrating and need to remove all the users from bookstack but maintain the data. When we attempt to drop the users table we get errors, which doesnt come as a surprise.

is there a routine or something we can run to delete all the users except admin

Abijeet:
Fixes #579

I've also submitted a patch to update the documentation - https://github.com/BookStackApp/website/pull/16

### DIFF
--- a/app/Console/Commands/DeleteUsers.php
+++ b/app/Console/Commands/DeleteUsers.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace BookStack\Console\Commands;
+
+use BookStack\User;
+use BookStack\Repos\UserRepo;
+use Illuminate\Console\Command;
+
+class DeleteUsers extends Command{
+
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'bookstack:delete-users';
+
+    protected $user;
+
+    protected $userRepo;
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks @Abijeet for working on this. Marked for next release.

Instead of having raw SQL executed via this command could we:
1. Select non-admin users via an Eloquent query.
2. Cycle through those users and [call the `UserRepo->destroy($user);`](https://github.com/BookStackApp/BookStack/blob/master/app/Repos/UserRepo.php#L144) for each of the selected users.

This approach would keep things less syntax-specific if we're ever to support other DB's in the future and it also keeps all user-delete logic in a single place which makes updates and maintenance easier. 

I'm happy to make changes after merging when it comes to the next release otherwise I'll keep it open so you can update if you'd like to make the above changes.

Abijeet:
@ssddanbrown - Will take care of it.

Abijeet:
@ssddanbrown - Done, please let me know if you'd like me to tweak anything else.

ssddanbrown:
Thanks again @Abijeet, Now merged. Thanks also for fixing the `hasSystemRole`  function in your pull request. 

I added a small commit (0c1b1cd435b1b757fecbdd4122803b7fff58c1be) after merging this to standardise admin checks using this function instead of using the name since I was being inconsistent before.

Abijeet:
Sounds good!

PR #623, opened 2017-12-20
Update styles.css

yoginth:
New Generation Styles!


ssddanbrown:
Hi @yoginth,
Thank you for your pull request. Unfortunately I cannot merge this for a few reasons:

1. All dev work should be done on master, the `release` branch is strictly for publishing releases.
2. Any styling work to be included should be done in the source `scss` files rather than directly to the minified css. 
3. It's not clear what this pull request actually achieves or what issues it relates to.
4. The changes I could see when testing the pull request were quite strongly themed and did not match the rest of the application styling.

For the reasons above I will close this pull request.

yoginth:
yeah fine! the customized theme is good! take a look at here https://docspen.ga/

PR #629, opened 2017-12-24
Added Design Updates!

yoginth:
Removed Navbar in Settings Page only for users! which is looking bad!

### DIFF
--- a/resources/views/books/list-item.blade.php
+++ b/resources/views/books/list-item.blade.php
@@ -1,5 +1,5 @@
 <div class="book entity-list-item"  data-entity-type="book" data-entity-id="{{$book->id}}">
-    <h4 class="text-book"><a class="text-book entity-list-item-link" href="{{$book->getUrl()}}"><i class="zmdi zmdi-book"></i><span class="entity-list-item-name">{{$book->name}}</span></a></h4>
+    <h4 class="text-book"><a class="text-book entity-list-item-link" href="{{$book->getUrl()}}"><i class="zmdi zmdi-book"></i><span class="entity-list-item-name">{{$book->name}}</span><small class="small-title">{{ trans('entities.book') }}</small></a></h4>
     <div class="entity-item-snippet">
         @if(isset($book->searchSnippet))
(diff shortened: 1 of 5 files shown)


yoginth:
Added Small Title on the side of Titles!

resources/views/settings/navbar.blade.php
```diff
@@ -1,4 +1,4 @@
-
+@if($currentUser->can('user-manage') || $currentUser->can('settings-manage') || $currentUser->can('user-roles-manage'))
```
Abijeet: Not sure what this achieves because currently if a user doesn't have access to any of these roles, this is what appears,

screenshot-2017-12-26 user profile bookstack

Isn't what you are doing the same?

resources/views/books/list-item.blade.php
```diff
@@ -1,10 +1,10 @@
 <div class="book entity-list-item"  data-entity-type="book" data-entity-id="{{$book->id}}">
-    <h4 class="text-book"><a class="text-book entity-list-item-link" href="{{$book->getUrl()}}"><i class="zmdi zmdi-book"></i><span class="entity-list-item-name">{{$book->name}}</span></a></h4>
+    <h4 class="text-book"><a class="text-book entity-list-item-link" href="{{$book->getUrl()}}"><i class="zmdi zmdi-book"></i><span class="entity-list-item-name">{{$book->name}}</span><small class="small-title">{{ trans('entities.book') }}</small></a></h4>
```
Abijeet: Not sure if this will be merged since in BookStack generally icons and colors are used to distinguish between entities.

yoginth: But it looks ambiguous!

ssddanbrown:
Hi @yoginth, Thanks for creating this pull request.

Unfortunately I will not be merging this, Pretty much fully due to the points raised by @Abijeet above. 

Adding the entity names within lists adds a lot of extra UI text for something that's already defined via color and iconography (Although I do know some of the icons could be better. Got ideas planned for the future on that).
The settings header should already collapse down if a user does not have permission to access the other areas.

Issue #619, opened 2017-12-16
draw.io integration


Kdecherf:
Desired Feature:

We deployed bookstack at current job for writing technical documentation about our platforms. Because it's better with a diagram, especially when you deploy µservices-oriented platforms, we try to add diagrams using draw.io to our pages to have a better view of interactions. Our current workflow for achieving that is:
- Create a diagram using draw.io on the standalone app
- Export a png
- Add the png to bookstack

Writers must manually share their draw.io sources (xml) in order to contribute to them.

It could be awesome to integrate draw.io to bookstack editor and save xml sources as editable attachments.

Abijeet:
I would be interested in this too.

ssddanbrown:
@Kdecherf Thanks for the suggestion. I agree this would have great benefit and saves us having to build-in and maintain a custom diagram editor. 

@Abijeet Thanks for the docs link. Integration is much simpler than I would have originally thought.

I've had a quick go and almost have it fully integrated into the WYSWIYG editor although TinyMCE is being a pain. I'll continue with this soon hence why I've assigned myself. 

I've had a quick look and can't see any license issues or payment required to add draw.io integration. My only slight concern is that their current revenue source is solely dependant on confluence integration. Hopefully we're too small-scale to be a concern to that but we need to keep it in mind.

lynchpatrickj:
I use draw.io this way as well, and it would save me a lot of time.

aljawaid:
I have also used draw.io for some diagrams for the business.  I usually use draw.io from within my nextcloud setup on the same domain.  Integration would be wicked.

The only reason I don't like to use the main draw.io (I prefer the nextcloud integrated version) is because the social media links/code seem to bloat the confluence integration that I have no use for.  Also, as it is within nextcloud, I can save my diagram direct into my business cloud.

Integrating draw.io into Bookstack would be excellent. So we could draw the diagram, save the xml (maybe even export it direct to owncloud/nextcloud) and then insert it into one of the books.

Very much look forward to this.

ssddanbrown:
Base integration now complete as part of #632, Ready for next release therefore will close this request.

AbijeetP:
@ssddanbrown - Just deployed this at my workplace today for my company, and they are all very happy about this. Thank you.

davidjgraph:
"My only slight concern is that their current revenue source is solely dependent on confluence integration. Hopefully we're too small-scale to be a concern to that but we need to keep it in mind."

[...]

PR #632, opened 2017-12-24
draw.io integration

ssddanbrown:
Relates to #619.

Very much a WIP at time of creating pull request. 

### TODO

- [x] Add updating of existing images.
- [x] Implement within markdown editor.
- [x] Add some basic testing to new endpoints.


Abijeet:
This looks good to me, I'll test this on my end once you are done.

resources/assets/js/pages/page-form.js
```diff
@@ -220,6 +220,124 @@ function codePlugin() {
 }
 codePlugin();
 
+function drawIoPlugin() {
+
+    const drawIoUrl = 'https://www.draw.io/?embed=1&ui=atlas&spin=1&proto=json';
```
tuaris: This should be user configurable in case a self-hosted version is preferred.

ssddanbrown: At first it will not be as I don't really know how/if self-hosted instances provide the same level & ease of integration via a public url such as this. I doubt they would otherwise it could be used by anybody.

Kdecherf: The default self-hosted instance is a static one (check https://github.com/jgraph/drawio#running) which does not provide any integration (Google Drive, Dropbox, …). Also you must specify `https=1` in query string if your own instance does not support TLS.

This lack of integration is not really important if files are saved as attachments inside Bookstack.

ssddanbrown:
Okay, This is now complete with some basic tests to cover so will merge into master for next release.
Feel free to raise any issues for problems and bugs found, I'm sure there'll be a few I've not found. 

There's a couple of limitations with the current implementation:
- For the markdown editor HTML is used to standardise the format used to recognise drawings. Think this is okay, Better than ignoring the markdown editor, but some may not like HTML in their markdown.
- Updating an drawing will update the drawing file/upload straight away, Not upon saving of the page. In the future it will probably be best to flesh this out with an image-manager-like implementation which will version drawing changes and retain old copies but what we have now is a good start.

PR #635, opened 2017-12-26
Adds button to allow users to toggle the book view via the books list…

### LINKED ISSUE #613: Add toggle to change view type on Books view
Instead of having a user setting in the profile view, Show buttons at the top of the page to switch between list and grid view.

If these are pressed by a logged in user the selected view would then be saved in the same way as the current profile setting.

Abijeet:
… page.

Closes #613

### DIFF
--- a/tests/UserProfileTest.php
+++ b/tests/UserProfileTest.php
@@ -103,7 +103,7 @@ class UserProfileTest extends BrowserKitTest
         $this->actingAs($editor)
             ->visit('/books')
             ->pageNotHasElement('.featured-image-container')
-            ->pageHasElement('.entity-list-item');
+            ->pageHasElement('.content .entity-list-item');
     }
 
     public function test_books_view_is_grid()
--- a/app/Http/Controllers/BookController.php
+++ b/app/Http/Controllers/BookController.php
@@ -46,7 +46,7 @@ class BookController extends Controller
             'books' => $books,
             'recents' => $recents,
             'popular' => $popular,
-            'new' => $new, 
+            'new' => $new,
             'booksViewType' => $booksViewType
         ]);
     }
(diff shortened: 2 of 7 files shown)


tests/UserProfileTest.php
```diff
@@ -103,7 +103,7 @@ public function test_books_view_is_list()
         $this->actingAs($editor)
             ->visit('/books')
             ->pageNotHasElement('.featured-image-container')
-            ->pageHasElement('.entity-list-item');
+            ->pageHasElement('.content .entity-list-item');
```
Abijeet: Just made the test a little more specific.

Abijeet:
@ssddanbrown - I think you might have a few things to add about the way I've handled this. Please let me know if you'd like me to make any changes.

Also I've written the test cases but I'm not able to get them to pass. Not really sure what's wrong. Any pointers will be appreciated.

yoginth:
it can be done only by admin, not the normal user!

yoginth:
Working Awesome!

yoginth:
surely need to be merged!

ssddanbrown:
Thanks @Abijeet for implementing this. Sorry I didn't answer your questions here and on the issue.

Your implementation looked good and I agree that UserController was probably the best place for this. If found to be awkward we can always move it in the future. 

I done a few tweaks after pulling down and playing with it, All under commit: 141bf227251287174d0a4cff833b88762031e6a1

I moved the toggle to the left to separate UI options and actions which aligns it with the homepage. I changed the HTTP method to PATCH as that seemed to be the most suited to me (Since we're `PATCH`ing the user settings with a single change). I also updated the button to change text and icon depending on view so the action behind the button is clear to the user.

Thanks again!

Abijeet:
Thanks, the changes you made make sense. I will keep it in mind for future.

Issue #630, opened 2017-12-24
After Deleting the User, user's profile picture is not deleted in the server!


yoginth:
### For Feature Requests

Desired Feature: none

### For Bug Reports

* BookStack Version *(Found in settings, Please don't put 'latest')*: 19
* PHP Version: 7.2
* MySQL Version: 5.7

##### Expected Behavior

Profile Picture needs to be deleted if the user is deleted.

##### Current Behavior

After Deleting the User, user's profile picture is not deleted on the server!

sampathg9:
Hi @ssddanbrown 
I'll work on this bug. 
I'll do the following to fix this bug

After successful deletion of a User account
- Delete profile pics from uploads folder
- Delete user profile pic record from images table

Please let me know If you want to perform any other thing in this action.

yoginth:
**_Thanks a Lot!_**

sampathg9:
@ssddanbrown - Submitted a merge request - https://github.com/BookStackApp/BookStack/pull/646/files

Please review it.

ssddanbrown:
Now closed as part of #646, 35e00ddb95733221a8868d81dec65e4d329eb1c3 and ead4b14d94d6daecdc7869c9c0f4ce597ef8f58c

PR #646, opened 2017-12-29
#630: Deleting user's profile pics on deleting of user account

sampathg9:


yoginth:
Fixed, working fine!

app/Repos/UserRepo.php
```diff
@@ -145,6 +148,12 @@ public function destroy(User $user)
     {
         $user->socialAccounts()->delete();
         $user->delete();
+        
+        // Deleting User profile pics
+        $profilePic = $user->image_id ? $user->avatar->findOrFail($user->image_id) : FALSE;
```
ssddanbrown: Here you could simply do `$profilePic = $user->avatar;` instead to achieve the same functionality and to be a bit clearer.

app/Repos/UserRepo.php
```diff
@@ -145,6 +148,12 @@ public function destroy(User $user)
...
+        if ($profilePic) {
+            $this->imageService->destroyImage($profilePic);
```
ssddanbrown: With the services, Instead of injecting it in via the constructor I've currently been using them via Facades when only sparingly used. Would be good to keep it this way to align with rest of image use in this Repo:
```
\Images::destroyImage($profilePic);
```

ssddanbrown:
Also, When it comes to merging this, do you mind if I 'squash and merge' since there's a lot of commits in this branch which are simply merges of master?

sampathg9:
I never mind. Please "squash and merge". I tried, but failed.

PR #651, opened 2017-12-31
Checks the target and the source book before performing the sort.

### LINKED ISSUE #638: Permission of Sorting Books
### For Bug Reports

* BookStack Version *(Found in settings, Please don't put 'latest')*: 19
* PHP Version: 7.1
* MySQL Version: 5.7

##### Expected Behavior

Disallow permission to sort others book.

##### Current Behavior

There is no Permission to sort others book but we can able to do it in BookStack!

Abijeet:
Fixes #638 

## TODO

- [x] Update the 'Books' sidebar list on the sort page to only show books the user has permission to update.
- [x] Add test cases. 

### DIFF
--- a/app/Repos/EntityRepo.php
+++ b/app/Repos/EntityRepo.php
@@ -113,9 +113,9 @@ class EntityRepo
      * @param bool $allowDrafts
      * @return \Illuminate\Database\Query\Builder
      */
-    protected function entityQuery($type, $allowDrafts = false)
+    protected function entityQuery($type, $allowDrafts = false, $permission = 'view')
     {
-        $q = $this->permissionService->enforceEntityRestrictions($type, $this->getEntity($type), 'view');
+        $q = $this->permissionService->enforceEntityRestrictions($type, $this->getEntity($type), $permission);
         if (strtolower($type) === 'page' && !$allowDrafts) {
             $q = $q->where('draft', '=', false);
         }
@@ -196,14 +196,15 @@ class EntityRepo
     }
 
     /**
-     * Get all entities of a type limited by count unless count if false.
+     * Get all entities of a type with the given permission, limited by count unless count is false.
      * @param string $type
      * @param integer|bool $count
+     * @param string $permission
      * @return Collection
      */
-    public function getAll($type, $count = 20)
+    public function getAll($type, $count = 20, $permission = 'view')
     {
-        $q = $this->entityQuery($type)->orderBy('name', 'asc');
+        $q = $this->entityQuery($type, false, $permission)->orderBy('name', 'asc');
         if ($count !== false) $q = $q->take($count);
         return $q->get();
     }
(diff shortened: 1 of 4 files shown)


app/Repos/EntityRepo.php
```diff
@@ -196,14 +196,15 @@ public function getPageByOldSlug($pageSlug, $bookSlug)
...
     /**
-     * Get all entities of a type limited by count unless count if false.
+     * Get all entities of a type with the given permission, limited by count unless count is false.
      * @param string $type
      * @param integer|bool $count
+     * @param string $permission
      * @return Collection
      */
-    public function getAll($type, $count = 20)
+    public function getAll($type, $count = 20, $permission = 'view')
```
Abijeet: I've modified the existing method. There should not be any side-effects. Let me know if you would like me to create a separate `getAllByPermission` method?

yoginth:
Not Working!

Abijeet:
Hey @ssddanbrown - Ready for your review.

Abijeet:
@yoginth - Can you please tell me what you are testing? The page you are accessing is not found?

ssddanbrown:
Thanks for tackling this @Abijeet. Happy (early) new year to you too!

Tested and all works fine in normal use.

I do have a concern though about an unlikely scenario. Basically, As it currently stands, The permission checks are done during the sort process.
If a permission check was to fail mid-way, Which would stop execution of the sort by throwing an exception, then it would leave the database in a dirty state since the book relations would be updated but the joint permissions would not be rebuilt and no activity would be added for the sorted books. 

This should never happen as the user shouldn't be able to select books they don't have permissions for but if permissions were to change while a user is sorting a book then this scenario could play out. 

Instead the following process would be safer:
1. Get the books being updated.
2. Check permissions on those books.
3. Only if all permission checks pass, Then proceed with any sort operations.

Let me know if you need me to explain this better at all. I'm happy to continue from this pull request if the above does not make sense.

Abijeet:
@ssddanbrown  - Makes sense, I'll make the changes tomorrow. My only concern is that I'll have to loop over `sortMap` twice, but I should be able to cache some of the details during this time.

Abijeet:
Couldn't take it up today, will try to handle it sometime this week.

ssddanbrown:
@Abijeet No worries, No rush at all on this.

Abijeet:
@ssddanbrown - I'm now checking for permissions on all the books being updated in a loop before the loop that does the actual update. A small part of the code had to be duplicated. In addition a query is being run twice. Got any ideas to avoid that?

yoginth:
Undefined Constant Update Error!

screenshot

ssddanbrown:
@yoginth Again, [That appears to be a 'DocsPen' only issue](https://github.com/DocsPen/Platform/blob/a8b9204d42af92be2258efd9082a44bbd39a2b89/app/Http/Controllers/BookController.php#L180). Issue never been apparent within BookStack.

ssddanbrown:
Thanks again @Abijeet for this.

I done some further refactoring in 281da59bae04068965213c74ff91d7f92749fe34 after looking how to answer your previous question. I decided the code was getting hard to read (Mostly due to how my original code was written) so I refactored this using laravel collections and the logic you introduced.

PR #665, opened 2018-01-13
Adds ability to secure images behind auth

ssddanbrown:
This feature puts images behind the authentication barrier so they are only viewable by logged-in users. Permission levels not taken into account, It's simply based on 'Is user authed?'.

Cannot be used alongside public viewing.
Set to be opt-in for now.

Public image access should still be served by the webserver as before, Secure image access goes through the app so will have a performance penalty.

Still in testing. Would be great for folks to test on their setups for any issues. (Do not test on production instances though). I experienced issues doing something similar when initially creating BookStack but so far everything has been working without issue on my dev machine.

## Use Instructions

To use simply set `STORAGE_TYPE=local_secure` in your `.env` file.
Files will be stored in a `storage/uploads/images` folder similar to attachments.

If you are migrating to this option with existing images you will need to move all content in the folder `public/uploads/images` to `storage/uploads/images`.

### DIFF
--- a/app/Http/Controllers/ImageController.php
+++ b/app/Http/Controllers/ImageController.php
@@ -1,6 +1,7 @@
 <?php namespace BookStack\Http\Controllers;
 
 use BookStack\Exceptions\ImageUploadException;
+use BookStack\Exceptions\NotFoundException;
 use BookStack\Repos\EntityRepo;
 use BookStack\Repos\ImageRepo;
 use Illuminate\Filesystem\Filesystem as File;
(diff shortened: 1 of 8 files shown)


yoginth:
still, it is visible to all user account who is not admin! it must be visible to admin alone! it will be safe.

screenshot

ssddanbrown:
@yoginth I don't think profile pictures have a reasonable expectation of being private. Their use is to identify a user to others. 

Adding extra layers of auth to image requests will slow them down considerably and I don't think it will be worth it. 

If you are worried about this then [following the advice here](https://www.bookstackapp.com/docs/admin/security/#images) will make image urls unreasonably difficult to find for other users.

yoginth:
Cover images and other attachments are also visible to logged in public users!

screenshot

yoginth:
even though if it is used for the internal purpose the security of other employee is important and not all people are going to use it for internal documentation someone can use it is as central documentation where others can host it!

ssddanbrown:
@yoginth I think I understand the user profile image problem you're describing.

Looking at your fork it looks like [you've disabled the relevant routes](https://github.com/DocsPen/Platform/blob/f56dfda8a2387c624825386450aae68907517c34/routes/web.php#L87) that control user images.

If you fix that things should work as expected in regards to user profile images.

yoginth:
ok thank you so much, fixed!

ssddanbrown:
Going to merge in as would be ideal to have the ImageService cleanups for #632 and should not effect existing instances.

PR #684, opened 2018-01-29
Add twitch socialite auth provider

moutonnoireu:
Followed some of the recent socialite added providers (okta, etc). It should work without any problems. If somebody could test ? 👍


lommes:
Currently the entry in composer.json is missing which leads to failing tests.

moutonnoireu:
Forget to run a ``composer update`` to update ``composer.lock``.

lommes:
Thanks for your work!

I just tested this PR and it worked for me! Only thing that would be nice is to change the keys in .env and config/services.php to match those of other services. 

TWITCH_KEY should become TWITCH_APP_ID
TWITCH_SECRET should become TWITCH_APP_SECRET

The TWITCH_REDIRECT is not needed in .env since it will be the same for each install of bookstack. It can bet set in config/services.php:

```
'twitch' => [
        'client_id' => env('TWITCH_APP_ID'),
        'client_secret' => env('TWITCH_APP_SECRET'),
        'redirect' => env('APP_URL') . '/login/service/twitch/callback',
        'name'          => 'Twitch',
    ],
```

moutonnoireu:
Made modifications following your inputs @lommes ;)

lommes:
There seem to be an issue with the svg file, which is displayed way too large.

moutonnoireu:
I reduced the icon size. Was thinking the svg was used dynamically :)

lommes:
In my opinion this can be merged if the icon thing is changed.

ssddanbrown:
Thanks both for your efforts on this! I've marked this to be in the next release.

Just one thing I noticed looking over the code, The SVG icon appears to simply be a png embedded within an SVG. It would be great if this was an actual SVG drawing.

I'd be happy to make that change before merging but I'll leave this open until then in case you want to make this change yourself.

moutonnoireu:
I'de be happy if you could do it, i'm not very familiar with svg. :)

ssddanbrown:
SVG updated and now merged in, Thanks, Works great!

PR #692, opened 2018-01-31
Corrected the keys for okta auth

lommes:
.env.example had wrong keys for okta from the original PR which where changed in ```config/services.php?```
This PR changes these keys in .env.example to the ones uses in config/services.php to match other auth services.


ssddanbrown:
@lommes Good spot, Thanks, Will merge in now for next release.

PR #695, opened 2018-02-03
Added Search Permission for not logged in Users

Yoginth:
**Added Search Permission for not logged in Users**

When before user logged in, if user try to search something, in options tag there is a personalized checkboxes.

I Removed it by adding simple condition.


ssddanbrown:
@Yoginth Good addition, Thanks, Will merge for next release.

PR #699, opened 2018-02-07
Dont require email confirmation for new users using google auth with …

thomseddon:
…domain restriction

We use G Suite for all our users, we have restricted logins to our own domain only via the "registration restrict" setting and so we don't want to use verification emails.

This is the most basic implementation (which we're now using), please let me know if there is any appetite for this change and if so, if you'd like any extra settings. For example perhaps we could just have an extra setting like "Don't use verification emails with Google login when using domain restriction" ?

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -158,7 +158,10 @@ class RegisterController extends Controller
             $newUser->socialAccounts()->save($socialAccount);
         }
 
-        if (setting('registration-confirmation') || setting('registration-restrict')) {
+        if (setting('registration-restrict') && $socialAccount && $socialAccount['driver'] === 'google') {
+            $newUser->email_confirmed = true;
+            $newUser->save();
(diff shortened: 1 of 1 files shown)


ChrisMcKee:
Similar change made in #966 disabling email verification for social logins

ssddanbrown:
Thanks for offering this pull request @thomseddon and sorry I never replied to this earlier, I could swear I had posted my thoughts but it appears not. 

I didn't accept this implementation since:
1. It's very google specific and I'm trying to avoid provider specific options where possible.
2. It infers that all admins will always trust google auth fully and does not provide the option to still have emails confirmed as an extra layer.

Since the time this was opened an alternative approach has been implemented which allows auto-confirmation of each social service. You could enable this functionality by adding:
```
GOOGLE_AUTO_CONFIRM_EMAIL=true
```
to your `.env` file. 

[Details of this can be found here](https://www.bookstackapp.com/docs/admin/third-party-auth/).

Sorry again for my lack of response.

thomseddon:
No problem, that's perfect! Thanks

PR #711, opened 2018-02-21
Added ability to configure email sender name

duncanbarnes:
Added env variable MAIL_FROM_NAME to allow the email sender name to be
customised. Also added MAIL_FROM to .env.example


ssddanbrown:
Thanks @duncanbarnes for creating this pull request, I'll merge in now for the next release.

Just as a FYI, I'm thinking about cleaning up the default `.env.example` soon so this variable may disappear from the `.env.example` file but it will be moved & documented within the official docs and/or within a `.env.example.full` type of file.

PR #714, opened 2018-02-26
Add CACHE_PREFIX to the .env.example file

pataar:
We had some problems with multiple BookStack instances using the same caching server. Perhaps it's a good idea to have this available in the `.env.example` file.


ssddanbrown:
Thanks @pataar for this pull request. I agree and will merge in now for future releases.

Just as a FYI, I'm thinking about cleaning up the default `.env.example` soon so this variable may disappear from the `.env.example` file but it will be moved & documented within the official docs and/or within a `.env.example.full` type of file.

PR #734, opened 2018-03-06
Reject crawlers by default

benrubson:
Hi,

This PR modifies `robots.txt` file so that crawlers are rejected by default.
I think it's a more secure default setting.
One who is in an indexing process generally knows what he does,  and then modifies `robots.txt` accordingly.

Thank you 👍 

Ben

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow:
+Disallow: /


ssddanbrown:
Hi @benrubson,

Thanks for bringing this to my attention and creating this pull request, Never really thought about the included `robots.txt` file before.

I completely agree with rejecting by default but thinking we should achieve this with a configurable solution. As it stands, If the `robots.txt` is altered it will throw a conflict upon update. 

Thinking we could have this toggle-able via the `.env` and either add a `robots.txt` endpoint or possibly set a [robots meta tag instead](https://developers.google.com/search/reference/robots_meta_tag).
I'm siding on a `robots.txt` endpoint as we could configure the file as a template which then could be overridden by the user, to extend or set additional rules, via theming if they wanted.

benrubson:
Hi @ssddanbrown,

Thank you very much for your answer 👍 

I agree, would be nice it if could be user-configurable.
Regarding the method, I don't really know what would be best.
Perhaps a simple on/off switch (reject/accept crawlers) would be sufficient ?

Thank you !

Abijeet:
I'm OK with allowing users to configure, but would also like to have it turned off by default.

ssddanbrown:
Thanks again @benrubson for raising this and providing a potential solution.

I have now implemented a configurable solution as discussed above. Details of implementation and use can be seen in #779. This will be in the next release. 

I will therefore go ahead and close this pull request.

benrubson:
Perfect thank you Dan 👍

Issue #732, opened 2018-03-04
Dynamic Links for Internal Pages


timconner:
**Enhancement Request**

BookStack currently creates all links as hard coded static links which causes links to easily become broken. 

As it stands these actions will cause broken links to a page:
- Renaming the Page
- Moving the Page
- Changing the name of the Chapter the page is in.
- Changing the name of the Book the page is in.
- Changing the server name of BookStacks. (e.g. books.company.com to docs.company.com)

**Proposed Change:**
Leverage database to generate dynamic links to entities internal to the BookStack instance when pages are rendered. This would require some type of markdown for identifying that a link is to an internal entity (Book/Chapter/Page) and the id of that page.

timconner:
**Update:** Most of the below stated changes do not actually cause broken links as the history lookup routine does a pretty good job of finding the page and redirecting the browser to it's new location. However, if you ever link to a deleted book it seems to break that functionality. I would still be more comfortable with an internal linking system that doesn't rely on revision lookup.

Also, I have created a PR that has the beginnings of an internal linking system. I piggy backed off the current transclusion system, but the editing UI doesn't have any of the needed changes yet. I want to come up with a way to also determine if the link is to a Book or Chapter as well. 

We have this precedent for pages thanks to the transclusion system: {{@page-id}}

Need something similar for Book and Chapters.

timconner:
Current thought would be something like this:
Chapters: {{{chapter-id}}}
Books: {{{{book-id}}}}

This could go well with the other enhancement request asking for "Book Shelves" for grouping books together which would just be one step further: {{{{{shelf-id}}}}}

ssddanbrown:
I'm going to go ahead and close this off since the current state of things is quite different compared to when this was opened. Many of the listed link breaking scenarios should now be handled by our references system, and base URL changes are now assisted with a built-in command. If there are common scenarios which are not properly handled we can address them individually. 

If you're coming to this in the future, Please note that there's extra context within #737.
I still feel the same, in that I don't want to introduce additional complexity to content, and therefore its usages. I want to keep proprietary magic out of content where possible.

PR #737, opened 2018-03-07
POC: Dynamically Render Entity Links

timconner:
Example render code for #732.

This will transform any link that has the HTML code like the following code block to the pages actual URL using page->getUrl(): `<a <a title="Test Link" href="{{@page-id}}">Link Text</a>`

Added support for Chapters and Books as well.
Chapters: `<a <a title="Test Link" href="{{{@chapter-id}}}">Link Text</a>`
Books: `<a <a title="Test Link" href="{{{{@book-id}}}}">Link Text</a>`

Incomplete Solution
This only handles the page rendering. The editors and link selection screens will not support this (yet). I modified the page's html in the database for testing the render functionality.

This also does not currently have a way to handle links to Chapters/Books, but I just wanted to put out a proof of concept.

### DIFF
--- a/app/Repos/EntityRepo.php
+++ b/app/Repos/EntityRepo.php
@@ -728,19 +728,35 @@ class EntityRepo
     public function renderInternalLinks($content, $ignorePermissions = false)
     {
         $matches = [];
-        preg_match_all('/<a .*href\="{{@([0-9]*)}}"/', $content, $matches);
+        preg_match_all('/<a .*href\="{{{*@([0-9]*)}}}*"/', $content, $matches);
         if (count($matches[0]) === 0) {
             return $content;
         }
         
-        foreach($matches[1] as $index => $pageId){
-            $matchedPage = $this->getById('page', $pageId, false, $ignorePermissions);
-            if ($matchedPage === null) {
-                $emptyLink = preg_replace('/href=".*"/', 'href="#"', $matches[0][$index]);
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @timconner,
Thanks for putting this together.

I think when building BookStack I thought about dynamic links but chose not to since it made the content fully standard and easier to render/export. I think now, after the project has evolved, dynamic links would be more of a benefit and we are now already doing some level of dynamic rendering anyway.

The only thing I'm not fond of is the syntax. I would prefer something that's a little easier to read and maybe more extensible. Maybe something like:

* `{{page@5}}`
* `{{chapter@50}}`
* `{{book@2}}`

The downside is that it's more fixed to English in addition to the default entity names but I'm thinking this format would be easier to read both in visual readability and code pattern searching.

timconner:
I like that format and it would be easy enough to code as well. If it being English only isn't a limiting factor I'd be happy to update the POC. I took a quick shot at the GUI side of things, but it seems that the markdown interpreter escapes the curly braces inside of a href attribute. 

I will have some time next week that I can spend on this. Would be happy to help out.

ssddanbrown:
Thanks for your original work here @timconner but I'm going to now close this off.

I would like to standardise on id-based URLs in the future, but I'd like them to still be full valid URLs to avoid any editor or raw content usability complexity. At some point in the future I'll start centralising feedback into a proposal to define a new format.

PR #752, opened 2018-03-18
Update entities.php for french language

Alwaysin:
Follow-up of #753, sorry for separate commit!

It is very cumbersome to add translations because you have to compare and see exactly where you have to insert the strings to match the default english file. I suggest that when adding string in english you add the same to all languages with a //TODO comment so that it is easier for translators!


ssddanbrown:
Thanks @Alwaysin, Will merge in now for the next release. 

I'll consider your comments on making translations easier. Not sure about leaving placeholder EN content everywhere but I'll think about a way to automate the placement process.

PR #759, opened 2018-03-21
Pull new Gravatar on email address update

Falklian:
Thought it might be useful to update a user's Gravatar if their profile email address has changed.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -161,6 +161,10 @@ class UserController extends Controller
             $user->external_auth_id = $request->get('external_auth_id');
         }
 
+        if ($request->filled('email') !== $request->old('email')) {
+            $this->userRepo->downloadGravatarToUserAvatar($user);
+        }
+
         // Save an user-specific settings
         if ($request->filled('setting')) {
             foreach ($request->get('setting') as $key => $value) {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -161,6 +161,10 @@ public function update(Request $request, $id)
             $user->external_auth_id = $request->get('external_auth_id');
         }
 
+        if ($request->filled('email') !== $request->old('email')) {
```
ssddanbrown: `$request->filled('email')` does not get the new email but is a boolean if it's in the request data.
`$request->old('email')` is not garunted to be the users old email address, Only old input.

This line doesn't reliably check if the email is being changed. This can easily be amended though to work as expected.

app/Http/Controllers/UserController.php
```diff
@@ -161,6 +161,10 @@ public function update(Request $request, $id)
...
+            $this->userRepo->downloadGravatarToUserAvatar($user);
```
ssddanbrown: If enabled, This method call will always replace the user's existing profile image with the gravatar image. I'd assume the user wouldn't want their profile picture replaced with a gravatar image if they've previously customised it. I think we'd need to account for this.

Falklian:
I'm not hugely familiar with Laravel, although write PHP for a living, so should be able to make this "more better".  When installing BookStack, I thought it was a bit weird for the Gravatar image to not come through when modifying the default admin account, which is why I was prompted to look into this issue. 

Anyway, would it be advisable to add something like an "update Gravatar" check box, or something similar?

ssddanbrown:
Yeah, That sounds like a good approach, Keeps control with the user.

ssddanbrown:
Since this is almost a couple of years old I'll close this off.
Thank you still for your efforts in offering this though!

PR #767, opened 2018-03-27
update japanese translation

msaus:
I updated some translation for Japanese although it was a little bit late.


ssddanbrown:
@msaus Awesome, Thanks very much, Will merge in now for the next release.

PR #783, opened 2018-04-03
Completely overhaul of the Spanish translation,  added missing strings

moucho:
- Checked all translated strings, and added missing from the english translation
- Fixed two minor concordance issues on the English translation


ssddanbrown:
Thank you @moucho for updating these translations. Will be in the next release.

PR #805, opened 2018-04-17
16 of April fix css

msaus:
There is always new line on the right top drop-down menu.
So, I added fix for that.


ssddanbrown:
Hi @msaus,
Thank you for the pull request.

Unfortunately we do not accept merge requests into the `release` branch directly, Only master, Therefore I'm going to close this pull request.

Feel free to open a new one with `master` as the target with your edits to `.scss` files only.

Also, If possible, it would be great if you could include a screenshot of the original issue just to confirm what this fixes.

Issue #770, opened 2018-03-27
Implement uploaded image versioning & deletion 


ssddanbrown:
Just want to finish off the base drawing system implementation.

Currently two flaws in the drawing system to address:

- [ ] Updating drawings will overwrite the old file, Even if the page is not saved.
- [ ] No current way to delete drawings

---

#### Implementation details

* Versions must be saved as separate versions on the file-system.
* Upon delete all versions should be deleted.
* Maybe abstract versioning to all images? Yep

#### Plan

- New Model/Table - `Image Revision`
    - Has fields: id, image_id, path, url, revision#, uploaded_by, timestamps
- On migration, migrate old paths & urls to new table before removal from `images` table.
- Build in updating/revisioning into image manager.
- Build drawings view into image manager.

ssddanbrown:
Updated title to reflect increase scope of this feature.

ssddanbrown:
This is now implemented in an alternative, simpler way to what's described in the above. See #837 for details.

The image manager can be used for drawings via a dropdown on the WYSIWYG drawing button or by Ctrl+Click'ing the 'Insert Drawing' option in the Markdown editor.

PR #837, opened 2018-05-13
Image Revisions

ssddanbrown:
Related to #770

### Remaining tasks

- [ ] Add configurable system limit of revisions to keep.
- [ ] A revision view to image manager when in drawing mode. (List, view, restore, delete?)
- [ ] Update html drawing id reference to be version-specific. Can currently show old drawing (Old path) but when going to edit it will fetch data for latest version (Based on ID).


app/Services/ImageService.php
```diff
@@ -262,43 +297,58 @@ public function getThumbnail(Image $image, $width = 220, $height = 220, $keepRat
...
+        $image->delete();
+    }
+
+    /**
+     * Destroys an image at the given path.
+     * Searches for image thumbnails in addition to main provided path..
+     * @param string $path
+     * @return bool
+     */
+    protected function destroyImagesFromPath(string $path)
```
Abijeet: Maybe we can also use return types? So `bool` here for example?

mendiromania:
> Add configurable system limit of revisions to keep.
very useful, just yesterday I was worried about how big my database could get

ssddanbrown:
Coming across a lot of awkward edge cases that will mean a lot of custom handling for drawings/revisions. 

Am now thinking it will be better overall to create images on drawing save instead of update which would cover the original issue and then build better tools for admins to clean-up & delete images in the system.

PR #846, opened 2018-05-18
Updated Spanish translation

moucho:


ssddanbrown:
Thanks for opening this pull request @moucho.

Could you confirm this change is definitely okay? I can't speak Spanish at all but the original text looks like it's closer to the EN original text of 'Set Date'.

moucho:
I did this change because in Spanish the literal translation doesn't fit on one line, so it doesn't look right on screen.
My reasoning for changing it it's because I think it's clear enough, on each line there is a header for the filter option so a link with the 'date' string should be used to set the date.
Just my two cents :)

msaus:
?

Message From Masao Soseki

________________________________

msaus:
Message From Masao Soseki

________________________________

PR #851, opened 2018-05-20
Update german translation

vriic:


ssddanbrown:
Thank you @vriic for these translations, Will merge in now for the next release.

PR #908, opened 2018-07-01
[WIP] Fixes the video export in PDF, HTML and Text

Abijeet:
TinyMCE only fiddles around with certain URLs, see - https://github.com/tinymce/tinymce/blob/0f7a0f12667bde6eae9377b50b797f4479aa1ac7/src/plugins/media/main/ts/core/UrlPatterns.ts#L22

For other URLs, TinyMCE just puts a `video` tag wrapped around a `source` tag. I handled that in the PDF and Text exports.

See attached files.

- Text - [tinymce-testing.txt](https://github.com/BookStackApp/BookStack/files/2153118/tinymce-testing.txt)
- HTML - [tinymce-testing-html.txt](https://github.com/BookStackApp/BookStack/files/2153121/tinymce-testing-html.txt) - Github doesn't allow me to attach HTML file.
- PDF - [tinymce-testing.pdf](https://github.com/BookStackApp/BookStack/files/2153116/tinymce-testing.pdf)

Not so sure if the code is very organized. Let's discuss and see if we can make it better.

I think we can also fix issues that might come if user's add an iframe.


ssddanbrown:
Hi @Abijeet, I've added a test and simplified the service code a little.

While looking at this I was thinking it may be better to make this less specific to the TinyMCE issue? Would allow us to make the code simpler. Instead of targeting specific services it may be better if we, on export:

1. Convert all `src` attributes starting with a protocol relative url to be https.
2. Convert all iframe/video embeds to your new formats for PDF and text exports. Accompanying text would be 'Embedded content:' unless it contains youtube/vimeo/dailymotion in URL. 

This way the src replacement logic would be separate from the embed conversion logic and we won't be regex matching the content against each service, Just once to fix `src`'s and once to convert embeds. This would also allow this to work for other media types that aren't defined in the pre-set service regexes.

Abijeet:
Hey @ssddanbrown - Would you mind if I refactored the code to not use regex but rather use the DomDocument class (https://secure.php.net/manual/en/class.domdocument.php) to perform the replacements?

ssddanbrown:
@Abijeet I don't mind if it would produce cleaner code or if there's a reason to do so (DOM Targeting accuracy, For example)

Abijeet:
Hey @ssddanbrown 

I think overall the code looks much better now. I'm also handling more scenarios with respect to text file exports.

app/Entities/ExportService.php
```diff
@@ -238,4 +298,37 @@ public function bookToPlainText(Book $book)
         }
         return $text;
     }
+
+    protected function getDom(string $htmlContent) : \DOMDocument
+    {
+        // See - https://stackoverflow.com/a/17559716/903324
+        $dom = new \DOMDocument();
+        libxml_use_internal_errors(true);
```
Abijeet: I'd like to have not done this, but there doesn't seem to be an alternative.

Abijeet:
Changing to WIP for now. See - https://stackoverflow.com/questions/8218230/php-domdocument-loadhtml-not-encoding-utf-8-correctly

ssddanbrown:
Since this pull request is fairly old, remains as WIP, has a few conflicts now and is patching an issue that is rarely raised I'm going to close this off. 

Still though, Thanks for your efforts in this @Abijeet. Hope working at Wikimedia is working out well, is a real cool company to work for!

Issue #75, opened 2016-03-08
Integrate LDAP with Roles System


ssddanbrown:

ssddanbrown:
Sorry this issue has been open for so long. I was hoping to add this in the next release (v0.14) but my limited knowledge of LDAP is making this difficult. If someone can answer the below it will help my understanding of LDAP so I can implement this feature:

* How are user groups normally managed?
    * Are they dictated by being nested within group objects or by users being directly assigned to groups (via `memberUID`?)?
     * Looking around there's many ways so it it up to the LDAP service and owner how they set up groups?
* Is there a standardised way to find users belonging to a group? If found references to using `memberOf` but this does not seem to be standard on OpenLDAP?

----

After looking around, [KanBoard](https://kanboard.net/documentation/ldap-configuration-examples) seem to have good support for LDAP including groups so maybe inspiration should be taken from there. It looks like it's going to be easier to get the groups of a user rather than getting the users of a group.

AceBlade258:
First, I am highly interested in this feature, and would love to assist/see it make progress. I have an installation in place at work, and this functionality is the one thing holding us back from a company-wide rollout.

* For group management in LDAP there isn't really a "standard", as it is up to whomever implemented the LDAP schema. The only time there are standards to work with are when LDAP is part of a larger directory service such as Active Directory, or FreeIPA/RedHat IDM. Most LDAP auth implementations focus on supporting them, as they are by far the most widely-deployed LDAP implementations.

* as far as I am aware of, `memberOf` is the most commonly used method of determining if a user is a member of a group, however some POSIX-only implementations of LDAP use `memberUid` instead. Also, as far as I am aware, it is more normal to get the groups of a user than the users of a group.

Please let me know if I can help test this: I have two installations in my lab I can test against with AD and IPA authentication set up. I would also be more than happy to spin up additional testing instances of any of the needed services.

brennanmurphy:
Hi @ssddanbrown, I was wondering if you still are working on this, as I've been activity looking at deploying Bookstack at my workplace, and this is also holding us back. If you don't have a lot of done on this, I would be happy to take a run at making this work in the next few weeks. 

I would agree with @AceBlade258 that ``memberOf`` is more common, as most of the installations of an LDAP server I've seen do try to replicate an Active Directory structure.

[...]

PR #911, opened 2018-07-02
#75 - LDAP user groups sync to Bookstack on Login

brennanmurphy:
Closes issue #75 

I set it up so that when users login using an LDAP connection, Bookstack will get all names of all groups that user is a part of on the LDAP server, and will try and match them to the names of roles created in Bookstack. For any matches it finds, it will add the user to that role on Bookstack. The groups to roles sync function only runs on login, so if a user's groups change on the LDAP server, they need to log out of Bookstack, and log back in for any changes to roles to take affect.

You do need to create roles on Bookstack with names that match groups on LDAP exactly for this to work. I did contemplate having Bookstack create roles on a LDAP login for all groups it found for a user, but talking to a few network admins, most applications have the app admins create the roles for matching.

I've added a few settings to the .env file to make this work. I've tried to explain them inline, and I can update the documentation on the website to reflect the changes if you would like. 

Happy for any feedback, and willing to make changes as needed. At the place I work, we have been using this patch in production for ~2 months, with 40ish regular users, and have not ran into any issues. I did test it with a few OpenLDAP servers, and one Windows Active Directory server, and it worked in both instances. 

This is my first PR here, so apologies if I have made any mistakes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -67,6 +67,15 @@ LDAP_DN=false
 LDAP_PASS=false
 LDAP_USER_FILTER=false
 LDAP_VERSION=false
+#do you want to sync LDAP groups to BookStack roles for a user
+LDAP_USER_TO_GROUPS=false
+#what is the LDAP attribute for group memberships
+LDAP_GROUP_ATTRIBUTE="memberOf"
+#what LDAP group should the user be a part of to be an admin on BookStack
+LDAP_ADMIN_GROUP="Domain Admins"
+#would you like to remove users from roles on bookstack if they do not match on LDAP
+#if false, the ldap groups-roles sync will only add users to roles
+LDAP_REMOVE_FROM_GROUPS=false
 
 # Mail settings
 MAIL_DRIVER=smtp
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -5,6 +5,7 @@ namespace BookStack\Http\Controllers\Auth;
 use BookStack\Exceptions\AuthException;
 use BookStack\Http\Controllers\Controller;
 use BookStack\Repos\UserRepo;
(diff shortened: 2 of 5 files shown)


ssddanbrown:
Thank you very much for your efforts here @brennanmurphy, This looks great. Will review in-depth later this week.

ssddanbrown:
Thanks again for this @brennanmurphy, Took me longer than expected to review since I needed to learn how to properly set up my own Ldap groups in a way where `memberOf` works properly :sweat_smile: .

All worked great and worked as you described. It did make a few changes/updates afterwards but nothing major. These can be seen in the following commits:

- 17bca662a7757a66538a8074527b43b2ba20a592
- be2ca9d4bbc8850ab70fce9549cf1087d4e4d721
- f421d83627494d1aae9dab29cb5c5bf7d1ea86ac

As an overview, Here are the changes I made:

1. Added tests to cover this new functionality.
2. Refactored so the LdapRepo was not needed and moved the methods into the LdapService.
3. By default, Mapped based on role `display_name` instead of `name` (But formatted lower-case and hypenated) since the `name` is not exposed within the interface and can be different to the display name. 
4. Added `External Authentication IDs` field in role edit view to allow a comma separated list of names to match against which will be used instead of the role name. This allows group/role mapping without needing to match the names and also opens up many-to-many ldap-group-to-role mappings.
5. Removed `LDAP_ADMIN_GROUP` option since this would be somewhat redundant after the custom mapping in (4).

Hopefully my changes won't effect your patched, in-production BookStack instance upon upgrade of the next release but be cautious just incase my mapping changes affect your setup.

I'll tag this so I remember to document all this functionality before release.

yoyokko:
jietu20180810-190426

Login error if set LDAP_USER_TO_GROUPS=true with Active Directory LDAP

Issue #835, opened 2018-05-12
Seemingly incomplete German notifications


alex2702:
* BookStack Version: 0.21.0 (has happened for a while though)
* PHP Version: 7.2

##### Expected Behavior
After certain actions, e.g. creating a page, notifications are shown. In English, this may read something like `Page Successfully Created`.

##### Current Behavior
In the German language files (`resources/lang/de/*`), these notifications are the equivalent to `has successfully created page`. Not only should there be a username in front of the text, it also seems like there should be the page name at the end of it. However, no other language (that I speak) shows this format. Instead, other translations simply state that the page has been created, no username, no page name. While I would appreciate the more informative approach that the German translator has suggested, it just doesn't look like it is supported at this point?

##### Steps to Reproduce
Any action like creating a page, saving an existing one, making comments etc will show the behavior.
It seems like it was introduced in commit [0d6a6b5b](https://github.com/BookStackApp/BookStack/commit/0d6a6b5b63c73ea631cb305c00b7b39627eb1eb1#diff-3dd70a88a403fb708cb5c84f003fb7c1).

alex2702:
I'd be happy to provide more appropriate translations if you agree with the need for them.

ssddanbrown:
Hi @alex2702, Thanks for opening this issue.

Yes, I agree this needs to be updated. Most of the `activities` translations do show an entity name + username alongside the text but any `*_notification` translation strings will be shown alone.

If you could amend these that would be great, Otherwise I can comment on the commit you referenced above and ask the original author.

alex2702:
Sorry for the silence. I'm willing to fix the problems. It would take me a few weeks though since I'm busy right now. If that works for you, just leave this open and I'll get back to you.

ssddanbrown:
@alex2702 That's cool, No rush at all, thanks for the update!

alex2702:
I created a pull request in #925. Let me know if there is anything to fix.

ssddanbrown:
Closed in #925

PR #925, opened 2018-07-15
Fixed German translations for notifications

alex2702:
As discussed in #835, I attempted to fix the translations.
Let me know if I made a mistake with the pull request or anything like that.


ssddanbrown:
Thanks @alex2702 for making this change, Will merge in now for the next release.

PR #933, opened 2018-07-22
French translation update

nicobubulle:


ssddanbrown:
Thank you @nicobubulle for providing these translation updates, Will merge in now for the next release.

PR #942, opened 2018-07-30
Updated Swedish translation

marcusforsberg:
Is there currently any way to be notified of upcoming new releases so I can update the translations in time? 😊


ssddanbrown:
@marcusforsberg Currently not. Releases arn't really on a strict schedule which makes implementing such a notification tricky. Will have a think about it.

Thanks again for these translations. Will merge in now for the minor release being released today.

Issue #939, opened 2018-07-24
RTL support


kmoj86:
Desired Feature: RTL support for both the text editor and the application as a whole.

--

hello,
I'm currently working on translating the application into Arabic, but the issue I'm facing is RTL support. **specifically in the text editor itself**. when writing in arabic (or any rtl language), a number of problems occur:

- bullet points like these are placed at the end of the sentence instead of at the beginning (they're always on the left side).
- when writing a word in a LTR language like english in the middle of the sentence, the whole sentence gets split into two segments in backwards order. making the sentence unreadable or very hard to read at best. especially when using more than one word.
- punctuation marks are placed in the wrong side when used at the end of a sentence.

this is the biggest issue in RTL support. not only in BookStack, but in any other application out there.
I'd like for the entire app to support RTL, but for now, the text editor is the highest priority. if not fixed then the app is unusable for any RTL language.

I'd like to contribute in making the RTL support but I lack the ability to do so. if there are any resources offering guides for adding RTL support then please mention them. also if anyone has the experience to do this then it would be greatly appreciated.

ssddanbrown:
Thanks for opening this @kmoj86.

I've now made some updates that will be in the next release to assist with RTL support and to accompany your Arabic pull request. The updates I've made can be seen here: https://github.com/BookStackApp/BookStack/commit/1cb6ae39c8a6eedad457a37ce406c8c0a700bc80

This brings some basic RTL support groundwork and features.
The main WYSIWYG editor will now have its direction set depending on the direction of your language.
If your language is `rtl` then you'll also get some extra toolbar buttons to control text direction.

The page content text direction should display correctly depending on the contents. I've made some updates to the bullet point styles to ensure they're rtl-compatible. 

This should now mean basic RTL support is now included so I will now close this request. There's definitely going to be further issues but for those it's probably best if they are raised as their own, focused, issue.

Aimranekds:
@ssddanbrown  Thank you for supporting RTL

ali-dashti:
Hi
Some other languages are `rtl`.
Please add all of them:
[rtl languages](https://www.andiamo.co.uk/resources/right-to-left-languages/)

ssddanbrown:
@ali-dashti Our language list contains languages that BookStack has been translated for. We don't simply list every language possible.

ali-dashti:
> 
Now persian language (`fa`) is in language list but it's not `rtl`.
Maybe `rtlLocales` should read from `.env` variable.

[...]

PR #945, opened 2018-08-03
Added Arabic language

mrkhalidj:
- translated the application into Arabic.
- added Arabic language to the drop-down list so users can choose it.
- added Arabic language to locales.

hope this will help more people to use this application 😃


ssddanbrown:
Thank you @kmoj86 for the work you put into this. A couple of questions:

1. Does everything generally work okay? From my understanding Arabic is a right-to-left language so I would have expected more work to be required to support such a language.
2. Is it okay if I squash this before merging? Just to reduce the amount of commits that will be merged for a cleaner history. Will still show as yourself being the author.

kmoj86:
@ssddanbrown 

hello dan,

- yes, everything works ok. but it's not the ideal way to view the application in Arabic. for best results, the CSS and text editor need to be configured for RTL. CSS only needs to be mirrored for the most part, the text editor on the other hand has to be reconfigured to handle RTL languages (Froala's WYSIWYG editor as an example).
But nothing that renders the app unusable. I've already tested it and the found no issues with it.
That being said, this is a step forward in supporting Arabic language by translating the user interface, so anyone who doesn't speak a second language can use the application. and if one day a more experienced contributor comes to add RTL support for a better user experience, then the translation will already be here for them to use.
**In short: yes, everything is _mostly_ ok** 😄 

- sure thing! you can squash this for a cleaner history.
this is my first pull request and I've been working on the translation on 2 different PCs, so sorry for the mess.

ssddanbrown:
Thanks again for this, Will be in the next release. My notes on RTL support here: https://github.com/BookStackApp/BookStack/issues/939#issuecomment-423740828

Didn't squash the commits in the end (Didn't know how to git-squash properly with an active merge conflict :laughing: )

kmoj86:
@ssddanbrown 
Great! thanks for all your work!
I tried to add rtl support myself by adding directionality plugin for TinyMCE, but for some reason it just didn't work for me. so this is really great news!

PR #948, opened 2018-08-05
Fix/Chinese translation

houbaron:


ssddanbrown:
@houbaron Thank you for providing these updates, I will merge in now for the next patch release.

PR #959, opened 2018-08-13
Update LdapService.php

yoyokko:
fix can't sync group roles in WIndows AD

### DIFF
--- a/app/Services/LdapService.php
+++ b/app/Services/LdapService.php
@@ -116,6 +116,8 @@ class LdapService
         } catch (\ErrorException $e) {
             $ldapBind = false;
         }
+        
+        $user->_username = $username;
 
         return $ldapBind;
     }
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @yoyokko,
Thanks very much for raising this. I knew there was some strange behaviour occurring with the LDAP group sync but I didn't understand what is was until you opened this since, in my test case, the username matched the external_auth_id.

Unfortunately the changes you made to fix this did break other areas of the application. The direct temporary usage of the `_username ` property on the User model would throw errors when that model needed to be saved to the DB.

I instead put together d2a9b312e98128d08bfc5838c8ac72ac13d0f47a which carried the username through from the request.

I'll put this is in a patch release now. Thanks again and sorry I could not merge your code.

yoyokko:
Hi,
I'm not a php coder and tried to find a way passing the username parameter to the method...
Sorry for the broken code.

Issue #574, opened 2017-10-25
Allow social login to create new accounts if registration is disabled


tuaris:
This is related to #572

In a business setting you may want to have registrations disabled but still allow new users to sign in/create accounts using SSO that is limited to a specific domain.

##### Expected Behavior

SSO is enabled.
Registration is disabled -or- Registration is enabled but signup form is disabled
Domain restriction is enabled.

A new user attempts to sign on using SSO.  The user account is automatically created.

##### Current Behavior

New users are unable to create accounts using company SSO when registration is disabled.
Or if Registration is enabled and SSO is enabled, a signup form is still present.

##### Steps to Reproduce

Enable SSO (ie Google)
Disable registration.

s0n-:
This is great! We have recently submitted an Okta Authentication and are running into this exact issue. Signing up and verifying isnt needed with SSO options. Would love this to be looked at so SSO options can automatically create the account and login.

ssddanbrown:
This is also somewhat related to #477.

My concern with this is that I'd like to support environments where registration is fully disabled but where SSO is used. 

Maybe, instead of allowing registration when registration is enabled, This issue should be re-focused to something along the lines of `Provide setting to disable registration with via standard username/password auth`?

ibrahimennafaa:
@ssddanbrown I think this could be an important feature to have.

As part of a company, I want all my coworkers to be able to authenticate themselves using Google (with the domain of the company). If they login with Google and the user does not exist => a new user is created. If the user exists => we log him in.

For external users, I want to be able to create users with the right Role and set an email/password for them to access.

So if we simply add the option to automatically sign up users authnticated with Google for the first time, it could cover all use cases:
- Allow signup of new users with email/password or no (depending on your need)
- Allow Google auth on a specific domain
- Do not require Google user of this specific domain to signup (automatically done on their first visit when they log in with Google)

ibrahimennafaa:
I opened a PR for that if you guys have time to check it

ssddanbrown:
As of commit e60d11ee04161d16875d706110876c31f9064565 this behaviour should now be possible.

Using Google, for example, setting:

```env
GOOGLE_AUTO_REGISTER=true
GOOGLE_AUTO_CONFIRM_EMAIL=true
```

in your `.env` file will allow registration via Google from the login page and will also skip email confirmation for google-signups while still checking the domain.

Will be in the next release due within the next few days. Therefore I'll close this issue.

PR #965, opened 2018-08-16
Feature/autoregistration social login

ibrahimennafaa:
For #574 

I added a setting to allow auto silent signup of user when social auth is used.

Few things to be aware of:
- enabling this setting would disable email verification for social auth but it's described under the setting + I think email verification for social auth is useless.
- if domain restriction is enabled => the auto social registration process respect it. 

Missing: translation & probably better wording (my english is not perfect)


ibrahimennafaa:
Sorry for pushing files under public/* :/ I know it's useless.

PR #990, opened 2018-08-28
OWASP Recommendation for input password

sebastienhouzet:
I propose to add autocomplete="off" on input password.
It is recommended by OWASP to increase security.

### DIFF
--- a/resources/views/form/password.blade.php
+++ b/resources/views/form/password.blade.php
@@ -1,4 +1,4 @@
-<input type="password" id="{{ $name }}" name="{{ $name }}"
+<input type="password" autocomplete="off" id="{{ $name }}" name="{{ $name }}"
        @if($errors->has($name)) class="neg" @endif
        @if(isset($placeholder)) placeholder="{{$placeholder}}" @endif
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for creating this pull request. 

I've done a little looking into this. I'm not sure this is relevant for the current browser ecosystem.

From https://www.owasp.org/index.php/Testing_for_Vulnerable_Remember_Password_(OTG-AUTHN-005):

> Since early 2014 most major browsers will override any use of autocomplete="off" with regards to password forms and as a result previous checks for this are not required and recommendations should not commonly be given for disabling this feature.

From that is sounds like this could be pointless and potentially misleading if those reading the code see the attribute but a experience a different behaviour in the browser.  

Do you know of any more recent, reliable sources that indicate this attribute would provide additional security?

ssddanbrown:
Closing due to lack of response

BigBadSheep:
Hi, now when LDAP login in enabled. Give possibility to storage in browser a domain password is very bad Idea :(

PR #998, opened 2018-09-03
Update AttachmentService.php

SkyAndClouds:
Couldn't upload some large videos (60 - 100 MB). In the logs, I saw the errors:

[Mon Sep 03 12:21:08.799485 2018] [php7:notice] [pid 101] [client IP:PORT] [2018-09-03 12:21:08] production.ERROR: Allowed memory size of 134217728 bytes exhausted (tried to allocate 121655240 bytes) {"userId":1,"email":"admin@admin.com","exception":"[object] (Symfony\\\\Component\\\\Debug\\\\Exception\\\\FatalErrorException(code: 1): Allowed memory size of 134217728 bytes exhausted (tried to allocate 121655240 bytes) at /var/www/bookstack/app/Services/AttachmentService.php:185)"} [], referer: http://NAS_IP:PORT/books/learning-materials/page/pate-de-ficat-de-pui/edit

solved the issue by modifying  /var/www/bookstack/app/Services/AttachmentService.php, by adding ini_set(‘memory_limit’,’256MB’); after use Symfony\Component\HttpFoundation\File\UploadedFile;

### DIFF
--- a/app/Services/AttachmentService.php
+++ b/app/Services/AttachmentService.php
@@ -4,6 +4,7 @@ use BookStack\Exceptions\FileUploadException;
 use BookStack\Attachment;
 use Exception;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
+ini_set(‘memory_limit’,’256M’);
 
 class AttachmentService extends UploadService
 {


TBK:
This should not be a default setting. If you want to allow larger uploads then add the setting to your php.ini

ssddanbrown:
Thank you for offering this pull request @SkyAndClouds but as mentioned above, this is not a setting that should be forced upon everyone but should be set in your PHP configurable file instead.

Therefore I will close this request.

PR #1002, opened 2018-09-10
Ru locale fix

mullinsmikey:
Fixed russian translations, added several missing strings


ssddanbrown:
Thank you for these translations @mullinsmikey. Much appreciated. Will be included in the next release.

Issue #784, opened 2018-04-04
Desired Feature: Ability to remove particular revisions


jsreynolds:
### For Feature Requests

Desired Feature: Ability to remove revisions

##### Expected Behavior

Ability to remove a particular revision - for security reasons for example where someone has pasted in code or notes which contain private information and saved. Make a change, it still lives on in a revision.

##### Current Behavior

Revisions are unable to be removed.

##### Steps to Reproduce

Edit a page, see the revision

Abijeet:
Hi @ssddanbrown,

I'd like to take a crack at this. I guess adding a **Delete** option here next to **Restore** would work?

![screenshot-2018-5-21 page revisions for examples of different bookstack](https://user-images.githubusercontent.com/1685517/40319947-a7a4578c-5d47-11e8-92c0-b106b9315ff2.png)

I'll look further into the technical complexities involved with this.

jsreynolds:
That looks like a great idea to me; thank you for considering it.

Abijeet:
Hey @ssddanbrown, Let me know if my comments [here](https://github.com/BookStackApp/BookStack/issues/784#issuecomment-390718746) sound okay, and I'll get started on this.

ssddanbrown:
@Abijeet Ah, Sorry I missed this, Yeah sounds good!

Abijeet:
This has been added in PR #1008 and will be in the next release.

jsreynolds:
Just wanted to say thank you - really looking forward to this!

PR #1008, opened 2018-09-15
#784 - Adds ability to remove particular revision.

Abijeet:
Hey @ssddanbrown ,

Been a little busy with my professional life and haven't had a chance to work on BookStack. Finally got some time this weekend.

Here's what's been done,

1. Added a drop-down for deletion confirmation on the revision page.
2. Wrote code to delete the revision
3. Added test cases.

Note that I'm not reducing the `revision_count` column in the pages table. I think the current revision number should not be changed.

# Screenshot
bookstack-revision-ddl-2


resources/assets/sass/_tables.scss
```diff
@@ -41,6 +41,9 @@ table.table {
   .text-center {
     text-align: center;
   }
+  td.actions {
```
Abijeet: Added this to allow the dropdown to remain visible when it overflows the `td`

Abijeet:
@ssddanbrown - Not entirely sure why the test cases are failing.

app/Http/Controllers/PageController.php
```diff
@@ -454,6 +454,40 @@ public function restoreRevision($bookSlug, $pageSlug, $revisionId)
...
+    /**
+     * Deletes a revision using the id of the specified revision.
+     * @param string $bookSlug
+     * @param string $pageSlug
+     * @param int $revisionId
+     * @throws NotFoundException
+     * @throws BadRequestException
+     * @return \Illuminate\Http\RedirectResponse|\Illuminate\Routing\Redirector
+     */
+    public function destroyRevision($bookSlug, $pageSlug, $revId)
```
ssddanbrown: `revId` does not match `@param`  comment on line 462

app/Http/Controllers/PageController.php
```diff
@@ -454,6 +454,40 @@ public function restoreRevision($bookSlug, $pageSlug, $revisionId)
...
+    {
+        $page = $this->entityRepo->getBySlug('page', $pageSlug, $bookSlug);
+        $this->checkOwnablePermission('page-update', $page);
```
ssddanbrown: Think it may be better to check `page-delete` permission here instead since it's effectively a related delete operation?

app/Http/Controllers/PageController.php
```diff
@@ -454,6 +454,40 @@ public function restoreRevision($bookSlug, $pageSlug, $revisionId)
...
+            throw new NotFoundException("Revision #{$revId} not found");
+        }
+
+        // Get the current revision for the page
+        $currentRevision = $page->getCurrentRevision();
+
+        // Check if its the latest revision, cannot delete latest revision.
+        if (intval($currentRevision->id) === intval($revId)) {
+            session()->flash('error', trans('entities.revision_cannot_delete_latest'));
+            return view('pages/revisions', ['page' => $page, 'book' => $page->book, 'current' => $page]);
```
ssddanbrown: Since this is not a successful request it's probably best to set a non-successful status code here. Maybe a 403?

Can just add a status code as an extra parameter to the `view` function. 

Making this change gives a clue as to why the test case fails.

app/Page.php
```diff
@@ -112,4 +112,16 @@ public function entityRawQuery($withContent = false)
...
+     * Get the current revision for the page if existing
+     * @return \BookStack\PageRevision|null
+     */
+    public function getCurrentRevision()
+    {
+        if ($id = PageRevision::where('page_id', '=', $this->id)->max('id')) {
+            return PageRevision::find($id);
+        }
+        return null;
+    }
```
ssddanbrown: The revision table can contain also contain 'update drafts' (Autosaves) so querying the revisions directly like this may not return what you expect. 

The `$page->revisions` relation (line 62) has the correct filtering applied and is ordered by created_at. Therefore you should be able to simplify this function the following:
```php
return $this->revisions()->first()
```

tests/Entity/PageRevisionTest.php
```diff
@@ -29,4 +28,30 @@ public function test_revision_count_shown_in_page_meta()
...
+
+        // Delete the first revision
+        $revision = $page->revisions->get(0);
+        $resp = $this->asEditor()->delete($revision->getUrl('/delete/'));
+        $resp->assertStatus(200);
+
+        $page = Page::find($page->id);
+        $afterRevisionCount = $page->revisions->count();
+
+        $this->assertTrue($beforeRevisionCount === ($afterRevisionCount + 1));
```
ssddanbrown: Tests are failing here. Took me a while but found why, Actually is a mistake in my own test above this one so my bad :man_facepalming: 

There is only one revision being created. The call to delete does not delete the revision since you've got a check to not delete the latest revision (Which also returns a `200` code as mentioned above).

There is only one revision before on the first update call (Line 33) the page slug updates due to the page name change. Therefore the second update call (Line 34) does not actually update the page since the page slug used (via `$page->getUrl()`) is the old one (Since the test `$page` object is not updated by the change in the application).

To fix this you can simply add a `$page = Page::find($page->id);` between the two update calls.
If you could do the same for my test (`test_revision_count_shown_in_page_meta`) that would be ideal.

Abijeet: Argh!! I should have seen that!

app/Http/Controllers/PageController.php
Abijeet: Fixed.

Abijeet:
Thanks for the review @ssddanbrown. Made the changes. 

Returning status code 400 instead of 403 in case user tries to delete the current revision.

PR #1021, opened 2018-09-22
Spanish translation

moucho:


ssddanbrown:
@moucho Awesome, Thanks for providing these. Will merge in now for the next release.

PR #1026, opened 2018-09-24
Update german translation

vriic:


ssddanbrown:
Thank you for these @vriic. Will merge in now for the next release.

PR #1047, opened 2018-10-06
Chinese language pack update 20181006

qianmengnet:
Chinese language pack update 20181006


qianmengnet:
In addition to the Chinese language pack, I added "Registration" in the navigation section of "Not Logged In" (link hidden after login), and the profile page profile.blade.php adds an anchor link to the number of posts, if it affects Please cancel the program development plan, thank you.
I was using github for the first time to update the program. I didn't expect to modify the contents of my fork and submit it.

PR #1057, opened 2018-10-11
Add markdown-it-deflist for <dl> in markdown

nilium:
This adds the markdown-it-deflist extension for markdown-it to support definition lists in the Markdown editor. In addition, there's some basic styling added to definition lists to make terms bold and indent their definitions.

Ideally this should have no negative impact on existing markdown held by a BookStack instance, but I can only speak for my team's use of BookStack. So, if this seems like something that could be merged, it might be good to hear from anyone else using BookStack if this would affect their pages.

This is intended to help with keeping glossaries and other things that fit the mold of a definition list (my team does both). That sort of formatting can always be worked around with some careful HTML, but figured I'd put this together and see if it's worth considering. Happy to make any changes if there's anything off about this, or I can close it if it doesn't seem like a good fit for BookStack.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6501,6 +6501,11 @@
         "uc.micro": "^1.0.3"
       }
     },
+    "markdown-it-deflist": {
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/markdown-it-deflist/-/markdown-it-deflist-2.0.3.tgz",
+      "integrity": "sha512-/BNZ8ksW42bflm1qQLnRI09oqU2847Z7MVavrR0MORyKLtiUYOMpwtlAfMSZAQU9UCvaUZMpgVAqoS3vpToJxw=="
+    },
     "markdown-it-task-lists": {
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/markdown-it-task-lists/-/markdown-it-task-lists-2.1.0.tgz",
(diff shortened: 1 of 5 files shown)


ssddanbrown:
Thanks @nilium for creating this pull request.

It all looks good and I can understand why you'd want this in BookStack.

My current concerns is that adding this will further mis-align the WYSIWYG and Markdown editors.
The long-term future plan is to make them interchangeable and this adds another thing to consider. 

I'll have a think about this though. I can see definition lists being a good feature to support for a documentation platform such as this.

nilium:
That makes sense. If this would make those changes considerably more difficult (since I don't know the level of effort needed on the WYSIWYG side), it'd probably be better to put this on hold for now. Anyway, it's up to you — this change isn't urgently needed by me, and I doubt it's a deal-breaker for anyone. Just want it out there so it's not sitting in my head.

ssddanbrown:
I'm going to close this off as I've standardised on comonmark, apart from a few older previously added markdown features, just to provide the guide/bar in regards to what markdown features and syntax we support.

I'm not totally closed off to supporting the `<dl>/<dd>` elements in the future though.

The Markdown editor still supports HTML input, so if needed you could always use that with some added styles in the "Custom HTML Head Content" setting.

Thanks again @nilium for still offering this to the project.

PR #1063, opened 2018-10-12
Add select account parameter for google authorization

justein230:
Useful for choosing an account if a default account is outside the scope of a G Suite organization.


app/Auth/Access/SocialAuthService.php
```diff
@@ -52,7 +52,12 @@ public function startLogIn($socialDriver)
     public function startRegister($socialDriver)
     {
         $driver = $this->validateDriver($socialDriver);
-        return $this->socialite->driver($driver)->redirect();
+        if ($socialDriver == 'google') {
+            return $this->socialite->driver($driver)->with(['prompt' => 'select_account'])->redirect();
+        }
+        else {
+            return $this->socialite->driver($driver)->redirect();
```
ssddanbrown: The else here isn't really required. Removing the else would make it clear what the default action is.

justein230:
@ssddanbrown I added an env switch for this called GOOGLE_SELECT_ACCOUNT, and I check if it is true before adding that parameter. Let me know if this is sufficient for this feature to be added.

app/Auth/Access/SocialAuthService.php
```diff
@@ -40,6 +40,9 @@ public function __construct(UserRepo $userRepo, Socialite $socialite, SocialAcco
     public function startLogIn($socialDriver)
     {
         $driver = $this->validateDriver($socialDriver);
+        if ($socialDriver == 'google' && env('GOOGLE_SELECT_ACCOUNT')) {
```
ssddanbrown: The `env` function should not be directly used. Ideally a new `config/services.php` entry should be added and then that should be used instead (`config('services.google.select_account')`).

Also, Since the logic here is the same for both login and register it would be ideal to have this code extracted to it's own function within this service class.

justein230:
@ssddanbrown I made the requested changes, and can't figure out why the checks have failed. I tested in my development environment and the changes work as intended. Any ideas?

ssddanbrown:
@justein230 Looks good. The tests being broken was totally my fault. I'd left some broken bits in `master` branch for a while so nothing to sorry about with your code.  Will merge in now for the next release.

Issue #1062, opened 2018-10-12
Images inserted in the markdown editor should be wrapped in an anchor tag leading to the original file


thomasjsn:
When inserting an image in the WYSIWYG editor a link to the original image file is added, it would be useful if the markdown editor did the same.

I'm just starting to migrate my content to BookStack, this feature would save me a lot of manual editing and copying and pasting URLs.

I've looked at the source; mind if I have a crack at implementing it? Anything special? Should a link be added even if no larger version of the image exists?

ssddanbrown:
Now merged, To be in next minor release.

PR #1064, opened 2018-10-13
Wrap images inserted with markdown editor with anchor tag to original file

thomasjsn:
As described in #1062.


ssddanbrown:
Thanks for this @thomasjsn. Have merged in now for next minor release which should be some point this weekend. Also added d0c166c207671f8b558173627160184809a1581a so the same is applied when an image is pasted or dragged into the editor.

PR #1088, opened 2018-10-25
Update zh_TW translation

kejjang:
Update zh_TW translation files to latest changes, also fix/improve some wordings to be more readable.


ssddanbrown:
Thanks @kejjang for providing these translation updates. Will merge in now for release in the next patch release due today or tomorrow.

PR #1101, opened 2018-11-01
Latest bookstack

ParserYa:


ssddanbrown:
Thank you for offering this pull request but I'm going to close this as it includes a whole mis-mash of changes with no clear goal, featureset or even description of what benefit this brings.

PR #1109, opened 2018-11-08
Chinese translation update for v0.24.1

qianmengnet:
0.24.1


ssddanbrown:
Thanks for these updates @qianmengnet.

Have now merged for release in the next patch release due today or tomorrow. 
For future reference, Please do not copy the `language_select` options from the EN files as these are to be in en only. I've removed them in the following commit: cfc1a2f045c6f7f1d4df1768193ed6b6debb9f98

Thanks again!

PR #1111, opened 2018-11-08
Add support for custom avatar provider like Libravatar or self-hosted service

Vinrobot:
This PR fixes JavaScript build process with the new NPM dependencies and adds the possibility to use a different user avatar service like Libravatar.


ssddanbrown:
Thanks for the pull request @Vinrobot.

After merging, I think I'll go through the existing wording and functions in the code and essentially 'de-gravatar' all areas to make them generic 'avatar' wording instead but with gravatar still as a default. Just warning now as I'll probably reduce and re-name some of the `env` variable names. 

Just one thing, Could you remove and changes that relate to the build process? Will make this request neater and I've just gone through and applied a more longer-term fix by updating all modules to standard ES6 import/export syntax which has caused conflicts with this. I know it's 100% my bad though for leaving the build process broken. Feel free to open on a new pull request if needed. 

Since this a a new feature and since I might break an old `.env` var this will be targeted for the next feature release rather than being in the next minor release.

ssddanbrown:
Now merged, ready for the next release.

Just so you're aware after updating, I have changed some of the config for this:
* `GRAVATAR_URL` will now be `AVATAR_URL`
* Placeholders will now use a `$` instead of a `%` in the format, So: `${hash} ${size} ${email}`. This is just to align it with some other BookStack options that have placeholders.

Details of changes are in b56fc21aafa08f8fd7a9909e20d7b9032b3080cb.
Thanks again @Vinrobot!

Issue #1110, opened 2018-11-08
A user with a role with 'manage users' system permission does not see the settings menu


mark-james:
**Describe the bug**
A user with a role with 'manage users' system permission does not see the 'settings' menu on the top bar. In order to see the bar the 'Manage app settings' system permission must also be included in the role.

**Steps To Reproduce**
1. Create a new role that includes the Manage Users system permission.
2. Assign the new role to a user
3. Login with this user and check the top bar. The Settings menu is missing.

**Expected behavior**
A user with the role should be able to access the Users sub menu of settings without having access to change the core app settings. 

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): BookStack v0.24.1 
 - Hosting Method (Nginx/Apache/Docker): Ubuntu 18.04 - Using included installation script

mark-james:
I've also discovered a related issue. A user with 'manage users' system permissions can update their user to any role including the admin role. Essentially giving themselves global access. Should I create a new issue or just update the above?

ssddanbrown:
@mark-james Thanks for raising this and sorry for my late reply, Looks like it'll be covered in #1119.

For the other point, It's probably best to open a new isuse and it may need a bit of discussion to confirm the exact logic to be implemented.

PR #1119, opened 2018-11-13
#1110 Show users link in top nav if user is signed in and only manages users

cw1998:
image


cw1998:
This PR is a potential workaround for users that don't have access to settings but can manage users. #1110

Issue #149, opened 2016-07-14
XML Import


davidPilkington:
Any thoughts on including an xml import tool? allowing users to import from other wiki's, evernote, phpdoc2 checkstyle, etc?

This would allow users to export from existing data sources into a new book. For example "my evernote book that was created because evernote started being d**k's" or "my book that was imported from a phpdoc xml because their ux is kinda crappy"

nydeggerm:
It would be nice if it is possible to import from Confluence.

leveled:
I posted a $25 bounty here. https://www.bountysource.com/issues/36011761-xml-import I'm sure others would love to be able to switch off Evernote, Confluence, etc.

ssddanbrown:
Still not sure about bounties applied to bugs, having features dictated by money.

Just as an FYI, For anyone looking to get the bounty, It is at my complete discretion to accept or deny potential merge requests for any reasons. 

Overall, A general "XML import" feature may not be the best way to achieve this. I don't really know the various formats of other systems but it could be better to work on building an API which then people could script "importers" to.  In which case, the API is probably something I'd like to build myself (At least the base of it) at some point.

ssddanbrown:
I'm going to close this issue off since I don't see it going anywhere due to being too abstract/general since there's no general standard XML format. There are some existing issues, such as #1049, for platform specific imports, but in general I'd leave this to be done by others outside of the core project using the API.

Thanks though @davidPilkington for the initial suggestion.

PR #1151, opened 2018-11-30
Added Import Book page

Jujhar:
Setup very basic import system that imports a book and creates a new page.
Does not work with phpdoc2 as they do not export to a single xml.

Addresses #149

### DIFF
--- a/app/Entities/Repos/EntityRepo.php
+++ b/app/Entities/Repos/EntityRepo.php
@@ -491,6 +491,33 @@ class EntityRepo
     }
 
     /**
+     * Create a new entity from request input.
+     * Used for books and chapters.
+     * @param string $type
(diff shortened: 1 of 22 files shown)


ssddanbrown:
Hi @Jujhar, Apologies for my lack of response on this over the last 15 or so months. 

Thank you for offering this pull request, The code within looks well done. Unfortunately this is not a feature I'd look to incorporate into the core project. While people may find this useful to help migrate their content into BookStack, I think the maintenance and support surrounding such as system will just be too much as the supported migration platforms evolve, change or show their side-effects & edge-cases that make such a system troublesome.

Instead I'd prefer to focus on extending the API so that people can more easily script imports for their desired system, and share such scripts if they wish.

Issue #890, opened 2018-06-17
Second German translation | How can i set an standard language for all new users?


iBooster:
### For Feature Requests

Okay, because I don't know Git(hub) ... I put this in here:

I have added an SECOND german translation.

In German speaking countrys, there are an formal and non formal way to talk to an person, while in english there is only one way, the "you". But in german, there is "Du" (non-formal) and "Sie" (formal).

Because the actual german translation is in formal german, i have added an non-formal translation.

You can download it from my Google Drive link below, would be nice if someone could add this official.

By the way, how can I set an other standard language?

https://drive.google.com/file/d/1J7gve1z35NZCBjXL-EcX7hIAAssiFKqQ/view?usp=sharing

ssddanbrown:
Hi @iBooster,

You can set the default language by adding `APP_LANG=your_lang` in the `.env` file where `your_lang` is the folder name used in the `lang` folder. Keep your custom files backed-up safely though as they could be overwritten upon update.

Are formal and non-formal German official language variations? If there are only small tweaks between them it may not be worth officially supporting both. Would the variant used change depending on how BookStack is used then? Sorry, Not very familiar with German myself.

Details of setting a default language should really by in the docs so I'll mark this as a required doc-update and close it once added.

iBooster:
Thanks for the fast answer. I already thought about it that it would be this easy to set the default language ;)

And for the Non-formal german part: in german it is a real thing with non formal an formal. http://angelikasgerman.co.uk/du-or-sie-when-to-be-formal-or-informal-in-germany/

In my opinion it can change the user expirience completely. While the non-formal german is most of the time choosen for private users (those who visit the page for their personal reasons), you choose the formal german for business users (like on a company website in the b2b area, for example).

If it is worth to support both of them? Well i dont know. For my project i use the non-formal german, because i speak to students and it is way more personal than the formal way (it is like you would call your users „Sir“ or „Miss/es“).

ezzra:
Yes its definitely worth supporting both of them! Well, it depends on how you use the language, one can get around the problematic when using passive forms ("The book has been created" instead of "You created the book"), but this often appears as artificially. And the difference between using formal or informal style in german has a big influence.

I am just planning to use bookstack for a german speaking project, as I noticed for now, the german translation isn't even completed yet, it looks like the translator left out all translation where the word `shelf` is used, thats indeed hard to translate to german because "Schrank" sounds very gruff :)

[...]

PR #1159, opened 2018-12-03
add German informal language

ezzra:
This commit adds German informal language ("Du" instead of "Sie") based on the current formal language files. Fixes this part of #890.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -84,7 +84,7 @@ return [
     */
 
     'locale' => env('APP_LANG', 'en'),
-    'locales' => ['en', 'ar', 'de', 'es', 'es_AR', 'fr', 'nl', 'pt_BR', 'sk', 'sv', 'kr', 'ja', 'pl', 'it', 'ru', 'zh_CN', 'zh_TW'],
+    'locales' => ['en', 'ar', 'de', 'de_informal', 'es', 'es_AR', 'fr', 'nl', 'pt_BR', 'sk', 'sv', 'kr', 'ja', 'pl', 'it', 'ru', 'zh_CN', 'zh_TW'],
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


ezzra:
> Not even they appear to directly support this.

but both of them use the informal "Du" and not the formal "Sie" like here ;) But even wordpress introduced formal "Sie" language variant finally some month before. Its actually an issue, in coding communities for example its very much common to use informal "Du", thats why I think Gitlab will not introduce the formal language. But in Wordpress the users are not part of any community, so its quite right to offer also the formal language variant.

However, you could rather consider not offering the formal language, but not offering the informal language would be wrong. But you are completely right, it would be much better if one would be based on the other, but it looks like there is only one fallback in laravel translation system?!

ezzra:
I just force-pushed the new variant with dependence on the formal german language files. I still dont know how to care for the comments within the files where no more language settings exists, in my opinion they could be removed. And maybe there should be a comment in each file, that this is depending on `lang/de`

How do you think about this new approach  @ssddanbrown ?

ssddanbrown:
@ezzra Looks good. Yeah, I'd probably remove any comments they don't have any translations under them.

Looking at the core laravel `Illuminate\Translation\Translator` class I think it'd be fairly easy to extend to add 'baseLanguage' support. Would also need to setup a custom provider to implement it but simple extend the existing one. This would be a nice step to remove any logic from these translation files and align with how the existing fallback works.

I'd be happy to set this up after merge though.

ezzra:
I would definitely prefer that, even if this here an ok solution, it's not really nice that it doesn't get any simpler. I was also surprised that the language system doesn't offer that option.

I removed the comments, but it's also not really nice that there is apparently no uniform way in the files how the comments should be formatted. Some are just `// pages` others are     
```
/**
* Password Reset
*/`
```

etc.

ssddanbrown:
Now merged. As said above, I tweaked the implementation and added tests to cover in 323bff7d6d3974e5752bfec9581d733b91e933f5.

Yeah, Would be good to standardise the comments at some point. Am looking to update the translation helper script at some point to match `en` line numbers and comments so it will be standardised when I get around to that.

Thanks again for opening this pull request @ezzra

Issue #1199, opened 2018-12-28
Access Copy Page function for a role that only has view only for 'All' but can create their own.


mark-james:
**Describe the bug**
If a user only has view access on a book page, the user should still be able to copy that page to a book that they do have create/update access for.

**Steps To Reproduce**
1. Create a role with Create/Update for own books/pages etc
2. Add View access to other books to the same role
3. Login as user with that role
4. Go to view only book. The 'More' menu on the right hand side is not available.

**Expected behavior**
For a view only page a user should be able to 'copy' that page to a book provided they have create/update access on at least one book.

**Screenshots**
Copy Functionality
![image](https://user-images.githubusercontent.com/26987752/50502116-e8564300-0ab0-11e9-8138-4737935c379f.png)

**Additional context**
Use Case: I have a bunch of pages that are templates. I don't want users to be able to edit them. But I do want them to be able to take a copy and then edit that copy.

ssddanbrown:
Now implemented as of merge in #1202. To be in the next patch release.

PR #1202, opened 2018-12-31
Copy function for view only pages

mark-james:
Hi @ssddanbrown ,
I'd like to open a pull request to handle the ability to copy a read only page as discussed in #1199 

This request is not yet complete.

Currently if a user has at least one of 'create page own' or 'create page all' then the option to copy the page is available. I deemed this as a prerequisite of any copy functionality.

There are (at least) two outstanding issues.
A user doesn't select a destination and pushes the copy button. And a variation when a user attempts to copy a page but does not have an existing destination available to copy it too.

How do you want to handle this?

What else do I need to consider?


ssddanbrown:
Thanks for offering this pull request @mark-james. To properly cover the issue this is going to need some deeper changes.

We can't simply check the `page-create-*` or `page-view` role permissions, Since these changes have the possibility to show the actions for those that cannot actually use them. These permissions also don't take into account the entity-level permissions.

BookStack keeps a record of role-to-entity permissions in the `joint_permissions` table. This will need to be looked-up for each of the current user's roles.

mark-james:
Thanks for the feedback @ssddanbrown , I'll use the joint_permissions table as you suggest.

I'm a little fuzzy on why. Is it because just using the role permissions doesn't account for pages that have been overridden? Is this what you mean by entity level?

I'm assuming taking this approach also handles at least part of the issue I raised initially - "when a user attempts to copy a page but does not have an existing destination available to copy it too"? Eg. They can create pages but don't currently have a available book to do so.

Finally how do you want me to handle the scenario when a user doesn't select a destination to copy the page too. We could either -
1. Disable the copy button until something is selected.
2. Default to the first destination on the list
3. Show a popup error message indicating the user doesn't have permission to copy the page to the existing book so they need to select a destination.
4. Or is there another alternative?

mark-james:
Hi @ssddanbrown , I've updated to use the joint_permissions table. It's all done in one query so it should be performant. I've confirmed that it works with overridden permissions.

Have I understood the requirements correctly?

ssddanbrown:
Thanks for this @mark-james, Sorry for the delay in my response.

Your latest changes did exactly what I desired in my last comment so all way good-to-go.
I did abstract the implementation since we'll be doing similar checks for other permissions and I also added a test. My changes can be seen in 5c9b528517f1cecd1f9277e87a6036c330e86bfe.

The will be in the next patch release.

PR #1272, opened 2019-02-12
Add german translations for shelves

Xiphoseer:
This PR should add the necessary german translations for shelves. If there are other translation keys missing, please let me know.


ssddanbrown:
Thanks for these translations @Xiphoseer, I'll merge in now ready for the next patch release.

PR #1274, opened 2019-02-13
add testing env variables for database host and db name

sonrac:
I'm runnig tests inside docker containers with gitlab-ci before deploy bookstack to swarm cluster, thats why using mysql test hosts as 127.0.0.1 is impossible.


ssddanbrown:
Thanks for the pull request @sonrac.

It's probably a good idea to make these configurable but I might standardise the variable naming before merge and remove the old env vars used here so watch out for that.

Just as an addition, I don't commit any level of stability to the testing components of BookStack so if you are using them in your production flow in any way just be aware that things may break. The testing components are really intended for the development process of BookStack rather than deployment/integration of user environments.

PR #1278, opened 2019-02-14
Bugfix: ldap_connect() ignores argument when given an URI

Fahrplan:
ldap_connect() ignores the port argument, if $hostname is in the Form of an ldap-URI (like ldaps://....), therefore the $port must be added to the first argument, if an URI is present, otherwise it allways tries to use Port 636.

This is a quick fix, which probably needs some further adjustment.

### DIFF
--- a/app/Auth/Access/Ldap.php
+++ b/app/Auth/Access/Ldap.php
@@ -17,7 +17,11 @@ class Ldap
      */
     public function connect($hostName, $port)
     {
-        return ldap_connect($hostName, $port);
+        if(strpos($hostName,":")===NULL) {
+            return ldap_connect($hostName,$port);
+        } else {
+            return ldap_connect($hostName.":".$port);
+        }
     }
 
     /**


ssddanbrown:
Hi @Fahrplan,
Thanks for offering a pull request.

The file you have edited is really only supposed to be a thin wrapper around the core PHP functions so I'd prefer to not add any logic in that file.

There should already be logic to handle this:

https://github.com/BookStackApp/BookStack/blob/25bc28a1be12c27802bfda0e89c46c8864f1fce3/app/Auth/Access/LdapService.php#L166-L183

Does the existing code not handle your use-case? If not, please can you confirm the format of your `LDAP_SERVER` variable?

Fahrplan:
My `LDAP_SERVER` variable looks like this:   

`LDAP_SERVER=ldaps://10.1.1.2:10636`

(Our LDAP Server runs on a different port then default AND uses LDAPS-Encryption). 

The logic quoted seems incorrect, because `ldap_connect($host, $port)` ignores the `$port` argument, when a protocol is given in the `$host` part, see [http://php.net/manual/en/function.ldap-connect.php](http://php.net/manual/en/function.ldap-connect.php)


When I feed my `LDAP_SERVER` string through the logic above the `$ldapServer` variable ends up with:

`Array
(
    [0] => ldaps
    [1] => //10.1.1.2
    [2] => 10636
)`

and the `$hostname` variable is filled with `ldaps://10.1.1.2`

After that `ldap_connect()` is called with the first argument, it tries to connect to `10.1.1.2:389`, because the port argument gets ignored when a protocol is present.

ssddanbrown:
Thanks @Fahrplan for explaining the above, Your description made the issue clear. I was not aware of this PHP functionality. 

I've just commited c24764018a9684fcf4418b359ddd793bb380063b to update the above & simplify the logic, while fixing this issue. Therefore will close this pull request. 
Thank you for your help!

PR #1312, opened 2019-03-05
Replace entity.link with a root relative link in actionShowLinkSelector

maantje:
Making the links from insert Entity link in the markdown editor root relative will make it that the links won't break when changing domains.


ssddanbrown:
Thanks for offering this pull request @maantje.

Unfortunately I'm going to reject this pull request due to the following:

* While this change may work on simple BookStack installs the URL logic is more complex and this would break on some other setups, Such as installs that are on the sub-path of a domain (https://example.com/docs/) for example.
* I would prefer to keep the URL strategy consistent throughout BookStack, and there are a lot more places where URL's are used. You can find my thoughts on moving to relative URL's in pull request #1275. 

Apologies if you find my response disappointing.

PR #1314, opened 2019-03-06
Update Dutch password_hint translation to correspond with validation

maantje:
At the moment the Dutch translation says ```Minimaal 5 tekens``` which means your password should be at least 5 characters long. But a 5 character long password is not allowed by the validator. 

I think this was a translation error from the English one where it says ```Must be over 5 characters```. To make the Dutch translation correct the Dutch translation should be changed to ```Minimaal 6 tekens```.

```
/**
 * Get a validator for an incoming registration request.
 *
 * @param  array $data
 * @return \Illuminate\Contracts\Validation\Validator
 */
protected function validator(array $data)
{
    return Validator::make($data, [
        'name' => 'required|max:255',
        'email' => 'required|email|max:255|unique:users',
        'password' => 'required|min:6',
    ]);
}
```

### DIFF
--- a/resources/lang/nl/auth.php
+++ b/resources/lang/nl/auth.php
@@ -27,7 +27,7 @@ return [
     'email' => 'Email',
     'password' => 'Wachtwoord',
     'password_confirm' => 'Wachtwoord Bevestigen',
-    'password_hint' => 'Minimaal 5 tekens',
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for spotting this @maantje.
I will merge in now ready for the next patch release.

Issue #1306, opened 2019-03-02
Allow custom name attribute in LDAP


craftbyte:
**Describe the feature you'd like**
I would like to be able to set the name attribute in LDAP just like I can set the mail attribute. We use a Windows AD server and the actual names are mostly stored in the `displayName` attribute instead of `cn`.

**Describe the benefits this feature would bring to BookStack users**
This would allow BookStack to be used in environments where the name attribute is not `cn`.

ssddanbrown:
Thanks for the suggestion @craftbyte, I can see how this would be useful.
I have assigned this to be done for the next feature release. 

#### Note for reference

LdapService@getUserDetails, Copy setup for email attribute.

dfanara:
I've worked on the changes for this issue, but I'm having issues with the unit tests. 

I haven't been able to get the tests on the master branch to pass, so I haven't started writing tests for this feature. I'm running phpunit version 6.5.14, setup the bookstack-test databases and ran the migrations. 

The majority of the tests are failing with errors similar to this:
```
Expected status code 200 but received 419.
Failed asserting that false is true.
```

Is this expected behavior on the master branch?

ssddanbrown:
Functionality now implemented as per #1317. Will be in the next patch release.

PR #1317, opened 2019-03-09
Issue #1306 - Allow custom name attribute in LDAP

dfanara:
Adds the ability to configure which attribute is mapped to the display name of the user.


ssddanbrown:
Thanks for this @dfanara. All looked and worked great! Merged ready for the next patch release.

craftbyte:
Hey, @dfanara I had some issues while using this.

When I set the config option to `displayName`, nothing was being populated while if I set the config to `displayname` it worked. Our LDAP server is a Microsoft AD and if I do an `ldapsearch` from the CLI, the attribute is name with a capital N, while the code in BookStack changes that attribute to lowercase somewhere. This should either be noted in documentation or looked at (cc @ssddanbrown).

dfanara:
@craftbyte I've looked into your issue and I found that php-ldap converts all of the ldap attributes to lower case (http://php.net/manual/en/function.ldap-get-entries.php).  

I'll submit a PR for a fix for this, but until it is released I think @ssddanbrown could add a note to the documentation.

Sorry for the confusion and thanks for pointing it out!

PR #1333, opened 2019-03-15
Fix LDAP attribute case sensitivity issue

dfanara:
Fixes an issue described in #1317 where attributes with upper-cased characters were not mapped correctly.

### DIFF
--- a/app/Auth/Access/LdapService.php
+++ b/app/Auth/Access/LdapService.php
@@ -80,7 +80,7 @@ class LdapService
     public function getUserDetails($userName)
     {
         $emailAttr = $this->config['email_attribute'];
-        $displayNameAttr = $this->config['display_name_attribute'];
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for this @dfanara.

Just one question, I understand that the casing will need to be lowercase for accessing the response from the LDAP query but I see this change will also lower-case the attribute in the query itself. Will that be an issue?

Just wondering if it would safer to leave the casing for the query and convert to lower-case only when reading the result (In the `getUserResponseProperty` function maybe).

ssddanbrown:
As of [this change](https://github.com/BookStackApp/BookStack/commit/f9fa6904b95409908e1ad8cbb344c7d1773bfc66#diff-202d4ca74adb4f11f304a3b15f428d63R97) I believe this pull request is now redundant. 

Thanks again @dfanara for offering this pull request.

PR #1344, opened 2019-03-22
Enable syntax highlight for OCaml, Haskell, Rust

XVilka:


ssddanbrown:
Thanks @XVilka,
Will be in the next release.

Issue #738, opened 2018-03-09
Easier tag organization and usage


jtjosvold:
**Desired Feature**
As our BookStack is growing I'm seeing multiple variants of tags being created because people do not know that a similar tag has been created (capitalization, abbreviations, etc.).  It would be helpful to have a page showing an index of tags so that editors can see what tags exist and, in addition, the tags could be linked to searches showing pages that are using that tag. This would also help us with individuals that are not as comfortable with using proper search syntax.

ssddanbrown:
@jtjosvold Thanks for the suggestion, So like a tag-overview page?

I can see the benefit. I think this would be worthwhile once Books and Chapters support tags.
Thinking about this I should really increase the priority on Book/Chapter tags just to finish the base tag system off. 

Somewhat related to #121.

jtjosvold:
Precisely. It would help to keep them organized and make searching easier for our users.

Thanks for looking into it.

Cave-Johnson:
Would just like to add my plus one to this. I love the way bookstack uses tags and it makes the search feature really powerful. 
This would be really useful for managing large instances of bookstack. I like the way that its been done in the pull request. It would also be neat if there was the ability for an admin to combine tags so that multiple similar tags can be cleaned up. As an example, 3 tags for Jan2020 currently appear as jan-20, January2020 and jan2020. It would be really neat functionality to merge the 3, selecting a master tag and then updating the other tag throughout bookstack. 

One other thing that would make using tags much easier is to show all tags that have currently been used within that book or that shelf to avoid users creating a new version of that tag. For example show tags currently used in this book under the add tag box.

Klemet:
I'd really love this to be implemented too. Right now, I kinda feel in the dark about the tags that exist in the Bookstack. It seems like the only way I could find them would be to look at each element of the Bookstack one by one, and make a manual list. To help users search for particular resources, a "all tags" page would be wonderful !

ssddanbrown:
Potential alternative implementation offered here: https://github.com/BookStackApp/BookStack/issues/2943#issuecomment-927850224

ssddanbrown:
In regards to the original issue, there seems to be a desire for something like a tag exploration page but the use-cases and requirements would need to be clear. This is something I've hard trouble defining. Don't want just a plain list of tag names.

I'm thinking:

- **Purpose:** To explore usage of tag names and values in the system.
- **Implementation:**
  - View would primarily be a paging list of tag names
[...]

PR #1153, opened 2018-12-01
WIP: 2019 design

ssddanbrown:
This is a WIP design update targeted for early 2019. It includes some functional aspects of the design as well as just styles.

### Design Goals

* Improved design consistency and feature usage throughout application.
* Provide a more modern, less "stock" feel.
* Cleanup of colour-scheme for easier customizability. 
* Lessen usage of glaring entity colours.
* Improved mobile usability.
* Increased functionality in core areas for improved app usage efficiency. (Book list sorting, for example)

### Early Preview (Comparison)

**Old Books List View**
bookstack_books_redesign_old

**New Books List View**
bookstack_books_redesign

PR #1354, opened 2019-03-27
Add Tag page.

jacac:
Regarding #738.

### DIFF
--- a/app/Actions/TagRepo.php
+++ b/app/Actions/TagRepo.php
@@ -59,6 +59,23 @@ class TagRepo
     }
 
     /**
+     * Get all tags for a particular entity.
+     * @param string $entityType
+     * @param int $entityId
+     * @return mixed
+     */
+    public function getForIndex($searchTerm = false)
+    {
+        $query = $this->tag->select('*', \DB::raw('count(*) as count'))->groupBy('name');
+
+        if ($searchTerm) {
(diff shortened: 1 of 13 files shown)


jacac:
@ssddanbrown Any comments if this tag page could be added?  I've attached a screenshot of the tag page.
image

ssddanbrown:
Thanks @jacac,
Sorry I have not responded sooner, This comes at a bit of an awkward time with the redesign (#1153) in progress.

It looks good! I think we'll need to evaluate the best place for this to sit in BookStack along with how we ensure it adds value. Will also need some edits to the code along with merging into the redesign. 

Probably won't be until after the redesign release that I can properly review this.

fklappan:
Sorry for thread necroing, but is this still relevant/doable? Some kind of "tag UI" would be great.

ssddanbrown:
Since this is relatively old I'm going to close this off, Although I appreciate that's due to my lack of review.

I'm still open to such a page being added, but it'd need to have a little more purpose than simply a list of tag names. I think the number of usages of that tag and the number of unique values used, for that tag, would be a good start since then it can provide an idea of utilisation. 

In terms of UI, A paginated tabular list may work better here than being alphabetical breakdown, especially when considering potential emoji and non-ascii usage.

Issue #1260, opened 2019-02-05
Add a "Create Book" button to the shelves.


Nebucatnetzer:
**Describe the feature you'd like**
A button which would allow to create a new book in the current bookshelf.

**Describe the benefits this feature would bring to BookStack users**
It would make it easier to create a book which would belong to a shelf right from the start.
At the moment I have to:
1. Create a book.
2. Go to the bookshelf.
3. Edit the bookshelf.
4. Move the book into the shelf.
5. Open the book.
6. Create content.

With the button I could:
1. Go to the shelf.
2. Create the book.
3. Create content.

**Additional context**
Books and chapters already work this way. Therefore it would be quite logical if the highest element in the structure would work the same way.

ssddanbrown:
Thanks for the suggestion.

This was mentioned in #1171 but is good to have this issue open as it is a separate requirement to that issue. Might possibly look at implementing this as part of #1153.

ssddanbrown:
Thanks to @cw1998, This feature is now in master and will be in the next release. Therefore I'll close this request.

Nebucatnetzer:
Nice, thank you @cw1998!

PR #1366, opened 2019-04-03
Add a 'Create New Book' button on shelf list

cw1998:
A solution for #1260 .
image

Some notes:
- When a book is added, it is added to the top of the shelf. 
- When a book is added, you get redirected back to the shelf rather than the new book. This works for me, but may not work for the majority.

### DIFF
--- a/app/Auth/Permissions/PermissionService.php
+++ b/app/Auth/Permissions/PermissionService.php
@@ -577,7 +577,7 @@ class PermissionService
                         $query2->where('has_permission_own', '=', 1)
                             ->where('created_by', '=', $userId);
                     });
-        }) ;
+            });
 
         if (!is_null($entityClass)) {
             $entityInstance = app()->make($entityClass);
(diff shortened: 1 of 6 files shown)


ssddanbrown:
Great stuff, Thanks @cw1998. 

I'm currently in the middle of a design update (#1153) so I'll probably wait until that's merged into master (Couple of weeks) before I review and merge this. Looks good from a quick scan of the changes though!

cw1998:
Hi @ssddanbrown , I've updated this PR after the merge of the 2019 design. I went for a blue "Create  New Book" button on an empty shelf rather than red (based on seeing orange text for adding a new chapter to an empty book) - hopefully this is correct.

ssddanbrown:
Thanks again for this @cw1998.

All looked pretty good. Made a few logical changes, in addition to aligning some existing wording, in 7f3f6e65b9705beabbafa5ceba8636a74ac22138:
* I changed the addition of books to be an append instead of a prepend. I think this is what users will generally expect, and aligns with the analogy where you'd generally add a new book to the end of a shelf.
* As you mentioned in your original post, I updated the post-create redirect to navigate to the book since I think most people will go on to create book content and this behaviour aligns with the existing book creation process. 

All merged, ready for the next release!

cw1998:
Awesome, thanks. I agree with the changes mentioned and they all make sense. I'll keep wording in mind next time too.
Looking forward to seeing the new design in a release!

PR #1391, opened 2019-04-15
Enable syntax highlight for Mathematica, R & sTeX

JtheBAB:
Enable syntax highlight for Mathematica, R & sTeX


ssddanbrown:
Thanks for offering this pull request @JtheBAB.
Have you tested these changes?

Just from a quick look at the code changes, I'd have thought these additional highlighting options will never be used since they're not in the mode-map, and won't be shown as an option in the UI.

JtheBAB:
I'm sorry. I was blind...

Issue #1325, opened 2019-03-11
On the user profile form, Roles with a dot in the name do not remember their input on validation failure


tpoindessous:
Hi

I got hit by a small bug but an unpleased one, as it takes me long time to figure it out.

**Describe the bug**
I made a new user : I checked one role for this user and I mistyped the password and the password-confirm didn't match. When I save my user, page is refreshed, error message is displayed but the role is unchecked. I did several tests and I can reproduce it : it seems that this bug occurred because :
  * I checked the last role of the list
  * OR because my role contains a dot "Client.NAME"

**Steps To Reproduce**
1. Go to '/settings/users/create'
2. Fill name and email
3. Check the last role, which contains a dot in his name
4. Mistype password and password-confirm so they don't match
5. Save user 

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): BookStack v0.25.1

Thanks !

ssddanbrown:
Thanks for reporting @tpoindessous.

I've confirmed this to occur when a dot is part of the role name. I'll update the issue name to be more specific based on the results that I've found.

ssddanbrown:
Fixed by @cw1998 in #1392, Fix will be deployed in the next release.

PR #1392, opened 2019-04-15
Replace dots with something else on user create and edit screens

cw1998:
Fix for #1325 
Dot notation handling in laravel means it was looking for the existence of a value in the wrong place.
Eg, if the role name was `a.role.name` it would look in `roles['a']['role']['name']` instead of `roles['a.role.name']`

I couldn't find anywhere where the name of the role is used in setting roles (they are set with IDs), so changing the dots in the form elements name attribute to something else seems to have fixed this without any unintended side effects.

Could be an issue if someone had a role called `a.role` and a second role called `aDOTrole`, but I can't see this being the case for too many users.

Edit: I should note that the display name of the role isn't changed, so it will still appear in the form with dots.

### DIFF
--- a/resources/views/form/role-checkboxes.blade.php
+++ b/resources/views/form/role-checkboxes.blade.php
@@ -3,10 +3,10 @@
     @foreach($roles as $role)
         <div>
             @include('components.custom-checkbox', [
-                'name' => $name . '[' . $role->name . ']',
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks again @cw1998,
Now merged for next release.

PR #1395, opened 2019-04-16
Add sign up link to login screen and fix registration form validation

### LINKED ISSUE #1239: missing proof on name length on registration
When a user registers, she can choose a name with only one letter. But when you change data on the user preferences page (by admin, possibly also by user herself), there will be an error that minimum 2 letters are required.

cw1998:
I think people expect to find a sign up link close to the login form, so added one next to forgot password if registration is enabled. Also fixed #1239 while I was there.

image

image

Hope this helps.

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -70,7 +70,7 @@ class RegisterController extends Controller
     protected function validator(array $data)
     {
         return Validator::make($data, [
-            'name' => 'required|max:255',
+            'name' => 'required|min:2|max:255',
             'email' => 'required|email|max:255|unique:users',
             'password' => 'required|min:6',
         ]);
--- a/resources/views/auth/forms/login/standard.blade.php
+++ b/resources/views/auth/forms/login/standard.blade.php
@@ -6,5 +6,10 @@
 <div class="form-group">
     <label for="password">{{ trans('auth.password') }}</label>
     @include('form.password', ['name' => 'password', 'tabindex' => 1])
-    <span class="block small mt-s"><a href="{{ baseUrl('/password/email') }}">{{ trans('auth.forgot_password') }}</a></span>
+    <span class="block small mt-s">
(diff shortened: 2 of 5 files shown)


JtheBAB:
I would vote that this "sign up" link can be hidden with the admin interface.

cw1998:
@JtheBAB the sign up link the header can be hidden by disabling new registrations in the admin interface. The link here will behave in the same way. I may be misunderstanding - do you mean hiding the link while keeping registration enabled?

JtheBAB:
Ah ok. Then it is ok :)
I was looking in my interface and somehow missed this....

ssddanbrown:
Thanks @cw1998 for this and for also writing the tests, That's great.

The "Already have an account?" link looks great. I might play around with the "Sign up" link placement if that's okay? It seems a little out-of-place next to the "Forgot password?" link, which has it's position very focused on the context of the task (Under the password field).

cw1998:
Hi @ssddanbrown , yeah absolutely. The placement seemed tidier than adding a new line, but I'm not 100% happy with the use of the bullet point to separate it. 

I actually meant to add a point in my first comment about the text.
On the login page says "Already have an account?", so I think something like can be done for the sign up link as well, but needs discussion as there's obviously translation work on the back of it,
Some things I had thought of were:

- Need an account?
- Don't have an account?
- Create a new account
- Register an account

cw1998:
@ssddanbrown Had a play around with the sign up link, and I think it looks good as a button. What do you think? Although it might confuse people if they fill out the form and click sign up, thinking it will create an account...
An option to consider anyhow.

image


Changes are on the following branch, so no changes to this PR.
https://github.com/cw1998/BookStack/tree/feature/sign-up-button

ssddanbrown:
I've had a play about and landed on this:

image

Looks better when you don't have so many social auth options :laughing:.

"Sign up" as a button looked okay but I think your point about creating confusion is very much true. I think it sits well at the bottom, after all options that relate to logging-in, as a link so it's obvious it's not a form control/action but instead a different location. 

Thanks again for your help on this one!

cw1998:
Looks great! 👍🏻👍🏻

PR #1415, opened 2019-05-06
Update swedish language files

Hambern:
Here's a new updated version of the swedish language files


ssddanbrown:
Closing, since re-opened in #1417

PR #1416, opened 2019-05-06
Updated sv language files

Hambern:
I've updated the Swedish language files but I'm not really good at this Git-stuff. Hope this is the correct way.


ssddanbrown:
Thank you very much for these updates @Hambern.
Unfortunately I cannot merge this, or #1415, since both originate from the release branch, and this request it attempting to merge back into release. 

The `release` branch is strictly for releases, no common development. You need to work on and target your pull request for master.

Here's how you can do this:

1. Copy out your updated sv language files so they're backed-up somewhere safe for now.
2. Clone this git repository, using the default master branch. 
3. Copy back over your updated language files into that newly cloned repo.
4. Commit and push to your own GitHub repo and create a pull request from that, with BookStackApp:master as the target.

You might have trouble previewing your changes while on `master` since the CSS/JS is not ready built on that branch but if you know they're okay on the `release` branch they should be good.

ssddanbrown:
Closing, since re-opened in #1417

PR #1420, opened 2019-05-06
Spanish translation

moucho:
Updated Spanish translation for v0.26.0


ssddanbrown:
Thank you very much @moucho for these updates. Will merge now for the next patch release.

PR #1433, opened 2019-05-07
Updated the Swedish language files

Hambern:
An even newer update of the Swedish language files


ssddanbrown:
Thanks again @Hambern, Just in time for the next patch release, your pull requests have great timing :smile: 

Will merge in now.

PR #1437, opened 2019-05-08
Added Dutch translations for some new texts

NootoNooto:
Translations for Bookshelf and a few common ones


ssddanbrown:
Thank you very much for these updates @NootoNooto.
Will merge in now, They will be part of the next BookStack patch release.

PR #1462, opened 2019-05-23
Update maintenance.php to link to Github release page & minor style tweak

DeftNerd:
* Added a link to the Github releases page when someone clicks the current release version (to look for changelog information, or to see if there are new updates)
* Removed unnecessary BR tag by fixing the CSS class for the version display so it is properly aligned with the rest of the menu

### DIFF
--- a/resources/views/settings/maintenance.blade.php
+++ b/resources/views/settings/maintenance.blade.php
@@ -7,9 +7,10 @@
         <div class="py-m">
             @include('settings.navbar', ['selected' => 'maintenance'])
         </div>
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for this pull request @DeftNerd. I'm sure admins will find this change useful.

I think, if we're linking the maintenance page version, it would be good to also link the same version string shown on the settings page for consistency.

ssddanbrown:
Thank again @DeftNerd. I've gone ahead and rolled this out to the normal settings page also.

PR #1485, opened 2019-06-11
Updated the french lang files

lucaguindani:
Hope this helps with the new interface.


ssddanbrown:
Great stuff @lucaguindani, Thank you so much. Will merge in now to be in the next feature release.

PR #1486, opened 2019-06-11
Added locale and text direction to html templates

kostasdizas:
Added top level flags to allow browsers to render locale specific features correctly

### DIFF
--- a/app/Http/Middleware/Localization.php
+++ b/app/Http/Middleware/Localization.php
@@ -59,6 +59,8 @@ class Localization
             $locale = setting()->getUser(user(), 'language', $defaultLang);
         }
 
+        config()->set('app.lang', $this->getLocaleIso($locale));
+
         // Set text direction
         if (in_array($locale, $this->rtlLocales)) {
             config()->set('app.rtl', true);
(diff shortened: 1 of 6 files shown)


ssddanbrown:
Thank you for this @kostasdizas, Is actually quite timely since I was looking to add the lang attribute soon as part of accessibility recommendations.

I think we'll need to use different locale values for the lang attribute. We have a few values in the BookStack list that are not really official lang values (`de_informal` for example). We have a mapping to system locales for some of the date logic so we can look to expose those values for this.

In regards to the text direction, do you have much experience of globally forcing the direction on an application like this? So far I've been only applying the direction selectively as I know it can have other side effects from my own working experience, and the browser should auto-set the direction based on the text itself. Would be good to understand if this is solving an issue otherwise auto could possibly be better? Just thinking about multi-region scenarios where perhaps the bulk of the content is written in English but the user's native language is a RTL language so they use that just for the interface.

kostasdizas:
Indeed, that wouldn't work for some of the non-standard names. I've updated it to use that map instead. 

As for text direction, I cannot claim any expertise in the subject. It is how it's normally done, but you'd probably need some expert's advice about how the layout is affected.

ssddanbrown:
Thanks again @kostasdizas, Now merged for next release.

I've removed the direction items for now to not effect live instances, Can always add in the future if found to be needed.

PR #1497, opened 2019-06-18
Added i18n support for default data

kostasdizas:
Default data loaded as part of the initial setup and/or upgrades can now be translated.

This is meant to be a draft commit, most new strings are not very i18n friendly yet.

I'd appreciate some feedback on how this was done. One of the main issues that came up, is that the translation file introduced contains strings that are no longer valid, which might be an unnecessary burden for translators.

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -23,7 +23,7 @@ class CreateUsersTable extends Migration
 
         // Create the initial admin user
         DB::table('users')->insert([
-            'name' => 'Admin',
+            'name' => __('migrations.users.admin'),
             'email' => 'admin@admin.com',
             'password' => bcrypt('password'),
             'created_at' => \Carbon\Carbon::now()->toDateTimeString(),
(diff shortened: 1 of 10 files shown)


ssddanbrown:
Hi @kostasdizas,
Thank you for offering this and your efforts in making such changes, And sorry for my very late response. To be totally honest, I'd rather not include migration translations for the following reasons:

* The migrations should be kept as independent as possible in regards to external files & logic to ensure they remain absolutely stable at all times.
* These translations will be more for translators to manage, and their use may be confusing or unclear since they're not part of the interface itself and, as you mentioned, some may no longer be used. 
* All this text is generally content that can be altered in the UI fairly simply anyway.
* I'd expect most instances would migrate as part of the setup, before they realise the UI is not in their preferred language and before they discover the global language option, leading to these translations being under-utilised. 
* The globally-set language will not always reflect the preferred language of those managing the system, whom most of the text will be shown to.

I'm therefore going to close this; Sorry. The changes you made did look of good quality though. From a quick scan, The only change I would have requested, If I did wish to have this feature included, would be to use the older `trans()` syntax for the translations to align with the rest of the app.

PR #1502, opened 2019-06-21
fix translate for "actions"

ezzra:
"Actions" is plain word in page show file, translation is not used. This is a fix based on release, but its also unfixed in master yet.


ssddanbrown:
Cherry picked into master under a82d9fdba5175054a9f1707e41d0fd6577447b00

Thanks again @ezzra.

PR #1503, opened 2019-06-21
add missing translations to german language file

ezzra:
this adds missing translations for german language, there might come some more, so just keep it open


ssddanbrown:
Thank you very much for this in addition to the good find on #1502. Would you be able to offer these changes on a branch of master, Targeting merge into master? 

The `release` branch is designed to be untouched apart from when deploying a release.

ssddanbrown:
Closing since these changes have effectively already been made via other means, and manually cherry-picking these changes into master led to no change.

Thanks again though @ezzra for offering these.

PR #1505, opened 2019-06-23
Hide permissions table unless custom permissions are enabled

timoschwarzer:
I just spent too much time figuring out why my custom permissions were not working. Turned out I didn't enable them.

This PR hides the permissions table unless you enable custom permissions. That should save some time and frustration for users who apparently cannot read (like me :smile:).

hide-permissions-table

### DIFF
--- a/resources/assets/js/components/entity-permissions-editor.js
+++ b/resources/assets/js/components/entity-permissions-editor.js
@@ -1,0 +1,20 @@
+
+class EntityPermissionsEditor {
+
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Neatly implemented, thanks @timoschwarzer.

Eventually I want to change the permissions system to be sensible so you selectively add roles to permit/deny but that is on the roadmap for the future, as this is a nice UX improvement until then. 

Will merge for the next feature release.

Issue #76, opened 2016-03-09
Postgres support


thomasfrivold:
I think you should add support for, or change the database layer to Postgresql for future proofing and ease of developing new exciting features.

How about 2226 interesting posts about how great Postgres is: https://hn.algolia.com/?query=postgres&sort=byDate&prefix&page=1&dateRange=all&type=story

When I switched the data layer from Mysql to Postgresql on my own project, I could immediately begin coding new exciting features into the app. These were great because I didn't have to use libraries or hardcore anything, just simply made calls to the database layer. Stuff like awesome searching capability and many other great features.

ssddanbrown:
Hi @thomasfrivold, Thanks for the suggestion. This would be a fairly large change to require people to use PostgreSQL. MySQL is still very popular and is really simple for people to install and use as many are already familiar with it due to it's heavy usage in open source projects such as Wordpress and others. I also like the fact BookStack can currently be used with either MySQL or MariaDB. I would be happy to look into switching it to PostgreSQL but I feel there would really need to be a reason to switch things up.

Is there anything that you can think of that PostgreSQL could specifically bring to BookStack?

tpetrauskas:
In [Laravel documentation page](https://laravel.com/docs/5.1/database#introduction) there is a statement that it supports:
- MySQL
- Postgres
- SQLite
- SQL Server

Would it be possible to choose between those in setup?

ssddanbrown:
@tpetrauskas Yeah, Thanks to Laravel, supporting PostgreSQL would be really simple for most of the database work. We would need add code to provide separate support for setting up full-text searching and it would be an extra system to test on. The current search indexes & Logic is focused on MySQL and I believe PostgreSQL uses a different syntax for full-text queries.

thomasfrivold:
Full text searches are really awesomely implemented in Postgres.


Check out these resources:
https://hn.algolia.com/?query=full%20text%20postgres&sort=byPopularity&prefix&page=0&dateRange=all&type=story

simhnna:
I'm also interested in Postgres support.

From the looks of it, either the fulltext search for mysql is not good enough, or it's not used right.

In postgres fulltext search strips words in a language specific way, so search uses the stem of a word. This ensures that searching for "s" will also get results that contain the german "ß". It also strips common filler words, like "a", "and", ...

I'm not sure if mysql can be configured to do that, but postgres does that more or less by default.

As for ease of deployment:
[...]

PR #1509, opened 2019-06-25
Add support for GCS storage

simonswine:
This seems to work well for us. Not too sure not touched PHP code in years, but more than happy to invest some more time into it.

I think we should use the URL abstraction of laravel-filesystem rather than building it in our code base (getPublicUrl). This would make it easier to support signed temporary URLs as a next step

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -120,6 +120,16 @@ STORAGE_S3_ENDPOINT=https://my-custom-s3-compatible.service.com:8001
 # An s3-format URL will be generated if not set.
 STORAGE_URL=false
 
+# Google Cloud Storage (GCS) configuration
+STORAGE_GCS_PROJECT_ID=your-project-id
+STORAGE_GCS_BUCKET=your-bucket
+
+# GCS service account path
+STORAGE_GCS_KEY_FILE_PATH=/etc/google/service-account.json
+
+# GCS custom URL
+STORAGE_GCS_URL=http://no-https-possible.example.net
+
 # Authentication method to use
 # Can be 'standard' or 'ldap'
 AUTH_METHOD=standard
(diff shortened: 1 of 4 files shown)


ssddanbrown:
Hi @simonswine,
Thank you for working on this and providing the code in this pull request. 

To be totally honest I'm not sure GCS is something I'd like to include and support at this time. It hasn't really been requested before so the overall benefit is likely limited at this stage.

From a quick search GCS appears to have an s3 compatible API, Maybe that could be used instead? From checking support for Minio and ceph I know other S3 API compatible services can work so maybe GCS can too.

nilium:
The S3-compatible API in GCS is generally not recommended -- it can incur extra costs, and authentication with it doesn't benefit from IAM permissions. I'd personally love to see support for this, since it would make our deployment of BookStack that much cleaner.

simonswine:
Hi @ssddanbrown,

thanks for taking a look and replying, it fair enough if you decide not to include support for GCS and potentially other Cloud providers (it is open source after all)...

I was well aware of the compatibility that GCS provides, but security wise it is a no-go for any serious use case (as @nilium points out). I am happy to take another look at this PR if you decide to change your mind and get it in a mergeable shape.

Whoever you gonna decide, thanks for Bookstack it works really well for our use case

ssddanbrown:
Since this issue was created I have had to use GCS myself so I can appreciate why the s3-compatible API isn't optimal. That said, GCS support really has had very little demand so I'd still not look to support it at this time. I kinda see this as horizontal growth of the platform as explained in [my comment here](https://github.com/BookStackApp/BookStack/issues/76#issuecomment-494956958).

If you still want to patch this in yourself, I'd be happy to accept a fresh pull request for some of the public URL generation cleanup in this to make maintaining such a patch easier for yourself. 

Additionally, it may be worth noting there appears to be a more recently supported GCS Laravel library here:
https://github.com/Superbalist/laravel-google-cloud-storage

Due to the above I'm going to close this. Thank you still @simonswine for originally spending some time and offering this pull request, and for your understanding in your previous comment.

PR #1517, opened 2019-07-02
Fix missing word in users_social_accounts_info string

bjubes:
fix en file to read:
> Disconnecting an account here does not **revoke** previously authorized access


ssddanbrown:
Thanks @bjubes, Good spot. Will merge in now so this fix is in the next release.

PR #1534, opened 2019-07-11
Fix pt_BR translations

DeehSlash:
As stated in the Upgrade Notes of version 0.26.0:

> Translations - Since many interfaces and lines of text have been updated, It may take a little while for some translations to catch-up. Expect to see more English text than usual if you're using a non-English language option.

This pull request fixes the pt_BR translation strings.

### DIFF
--- a/resources/lang/pt_BR/activities.php
+++ b/resources/lang/pt_BR/activities.php
@@ -1,12 +1,10 @@
 <?php
-
+/**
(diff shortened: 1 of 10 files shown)


ssddanbrown:
Great work, Thank you very much for offering these translations @DeehSlash. Will merge now so they're in the next feature release.

PR #1554, opened 2019-07-27
Hungarian translation

miles75:


ssddanbrown:
Thank you very much for offering these @miles75.

I will merge in now for the next release.

PR #1561, opened 2019-07-31
German translation extended

danielroehrig-mm:
I went through all (formal) german translation files - using the lang/check.php tool - and tried to fill in all missing translations and deleted the redundant ones.


ssddanbrown:
Thank you very much @danielroehrig-mm for these translation updates. Will merge in now to be in the next release.

PR #1584, opened 2019-08-13
Adds autofocus on the email field of the standard login page.

almandin:
Adds autofocus on the email field of the standard login page.


ssddanbrown:
Thank for this pull request @almandin and sorry for the late response.
Needed to be sure this wouldn't have a bad impact on accessibility. Don't think it should but have made some further tweaks to remove the altered tabindexes for these form components to help ensure a non-confusing accessible experience.

Have merged in for the next feature release, v0.28

Issue #1575, opened 2019-08-05
Inline code format cuts off long text when viewing page


c0shea:
**Describe the bug**
I have a list of text elements and some have an inline code snippet. If the inline code is longer than one visible line, the rest of the inline code text becomes invisible when editing and viewing the page. Unlike existing issue #1507, this happens when editing and viewing the page.

**Steps To Reproduce**
1. Paste in the following HTML into the page editor source code:
```
<li class="md-list-item">
<p class="md-end-block md-p"><span class="md-plain">Use for publicly accessible members in shared libraries like Core.Utilities using </span><span class=" md-link"><a spellcheck="false" href="https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/"><span class="md-plain">XML documentation comments</span></a></span><span class="md-plain">, e.g. </span><code><span spellcheck="false">/// &lt;summary&gt;Converts the &lt;paramref name="obj"/&gt; to the type specified by T. If the input type is not convertible to T, an exception will be thrown.&lt;/summary&gt;</span></code><span class="md-plain">.</span></p>
</li>
```
2. Notice that the text is cut off after `T, a`

**Expected behavior**
The inline code should wrap to another line.

**Screenshots**
![image](https://user-images.githubusercontent.com/17714002/62476149-7051a900-b774-11e9-82e8-f91aad970a46.png)

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): v0.26.3
 - PHP Version: 7.3.8
 - Hosting Method (Nginx/Apache/Docker): IIS

james-geiger:
@c0shea Is this what you were thinking the expected behavior would be?
![Annotation 2019-08-09](https://user-images.githubusercontent.com/15711820/62784428-1149a800-ba84-11e9-8ac3-d995c103c94e.png)

c0shea:
@james-geiger I was expecting it to look like this:
![image](https://user-images.githubusercontent.com/17714002/62815786-9320ec00-baeb-11e9-8fb0-6fa87f70b78c.png)

james-geiger:
@c0shea Is there a reason one should have the line space visible and the other not?

My thought would be that inline code should not default to a multiline block—the functionality exists in the CodeMirror utility to wrap the line instead of scrolling it.  Maybe this would be a better solution for the multi-line block. ([See CodePen example.](https://codepen.io/james-geiger/pen/rNBOGOR))

[...]

PR #1587, opened 2019-08-15
Fix 1575

james-geiger:
This PR addressed the problem in #1575 where inline code will overflow off the page.  Added line wrapping in codemirror block.  Could refactor this into a setting or remove if it doesn't fit into the overall design.


ssddanbrown:
Hi @james-geiger,
Thank you for this pull request. I have now merged it for the next feature release.

I did revert the codemirror line wrapping changes, at least for now, since this change could have quite an effect of how people's content would be displayed and I think further discussion would be needed to determine if wrapped is the better approach.

PR #1601, opened 2019-08-27
feature: support nested inclusions

scrooloose:
Previously we could only support one level of page inclusion. So, if we
have a three pages with this content:

    Page one: 'Test'

    Page two: 'Include page one: {{@1}}'

    Page three: 'Include page two: {{@2}}'

Rendering 'Page three' would not expand the nested includes, and would
result in:

    Include page two: Include page one: {{@1}}

To fix this we just keep expanding the includes until there are none
left.

Extract out the regex that matches the includes into a constant - since
this is now used in three places.

### DIFF
--- a/app/Entities/Repos/EntityRepo.php
+++ b/app/Entities/Repos/EntityRepo.php
@@ -53,6 +53,13 @@ class EntityRepo
     protected $searchService;
 
     /**
+     * PAGE_INCLUDE_MATCHER
+     *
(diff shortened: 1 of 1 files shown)


scrooloose:
Apologies for the delay. I will hopefully get onto this this week.

ssddanbrown:
I'm going to close this off now but would still happy to review PR if it was still desired and implemented the above controls/checks.

PR #1627, opened 2019-09-04
Update common.php

kostefun:
update ru


ssddanbrown:
@kostefun Thanks again for all your translation update pull requests. I'll merge them in now for the next patch release.

PR #1633, opened 2019-09-04
Updated Spanish translation

moucho:


ssddanbrown:
@moucho Thank you again for these translation updates, Will merge in now for the next patch release.

PR #1646, opened 2019-09-09
Update settings.php

kostefun:
fix rus


ssddanbrown:
Thanks again @kostefun. Will merge in for the next feature release.

PR #1674, opened 2019-09-19
Turkish Language Support

fatihbahceci:
See  Translated Example Site: http://kitap.ilimevim.com


ssddanbrown:
Thank you so much for putting together these translations. Unfortunately, Someone had already made a pull request before this one offering Turkish translations which has now been merged for the next release meaning I cannot now merge these due to conflicts and duplication. I'm so sorry if you spent a lot of time on these but the good news is that Turkish will be in the next release.

PR #1688, opened 2019-09-26
Show list of bookshelves that a book belongs to on a book view

### LINKED ISSUE #1598: Display shelves a book belongs to
Would like to be able to display which shelves a book belongs to on the books page. 

Currently, if you go to the books page, you have no way of knowing which shelves the book belongs to. Hence, a user is not able to access other similar books that might be categorized together under a certain book shelf. 

**Additional context**
Could display the name of the shelf (or shelves) that the book belongs to right below the description and above "created x days ago".

cw1998:
Closes #1598
I put these underneath actions because there could potentially be many that push the actions off-screen. But you will likely want to change the placement of this if this is a feature you wish to merge 👍 

Another part of this PR to look out for is the heading above the list. I went with Bookshelves because I think it should be clear enough from the context (a book) that it is referring to the bookshelves that the book can be found on.

Thanks

image
image

### DIFF
--- a/app/Entities/Repos/EntityRepo.php
+++ b/app/Entities/Repos/EntityRepo.php
@@ -413,6 +413,17 @@ class EntityRepo
         return collect($tree);
     }
 
+
+    /**
+     * Get the bookshelves that a book is contained in.
+     * @param Book $book
+     * @return \Illuminate\Database\Eloquent\Collection|static[]
+     */
+    public function getBookParentShelves(Book $book)
+    {
+        return $this->permissionService->enforceEntityRestrictions('shelf', $book->shelves())->get();
+    }
+
     /**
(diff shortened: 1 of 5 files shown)


ssddanbrown:
Apologies for the massive delay in review @cw1998 & thanks for offering this.

Code looked great. Also thanks for writing a test to cover it.
I performed an couple of changes upon your implementation:

- Moved the list to the left sidebar - This is just to align with the other navigation/traversal elements shown in BookStack, which are also shown on the left.
- Instead of showing some empty state text ("Shelves this is added to will show here...") I instead just hid the list. Via permissions, it is possible to currently completely hide shelves within BookStack, so this accommodates that scenario where someone may not use shelves otherwise they will have an always-empty list. 

Again, Massive thanks. Will be part of the next release (v0.29).

PR #1699, opened 2019-10-03
Enable pasting base64 images to the editor

### LINKED ISSUE #1697: Pasting base64 images broken
**Describe the bug**
In HTML it is possible to specify an image inline with a base64 encoded source, like:
``
Images created in Jupyter notebooks, like graphs generated with the `matplotlib` Python package, are for example of that form. In the latest version(s) of BookStack it is not possible to copy/paste those images, at least not in the WYSIWYG editor. Neither by copying or by dragging them. One has to save the image and upload them separately.

**Steps To Reproduce**
1. Create a new page with the WYSIWYG editor
2. Go to https://nbviewer.jupyter.org/github/empet/Math/blob/master/DomainColoring.ipynb
3. Right-click on one of the color wheel images and click 'Copy image'
4. Go back to the editor and try to paste (ctrl+v, right clicking does not give the option to paste)
5. Nothing happens
6. Open for instance the Gimp, press ctrl+v and the image is pasted correctly there

Alternatively:
1. Create a new page with the WYSIWYG editor
2. Go to https://nbviewer.jupyter.org/github/empet/Math/blob/master/DomainColoring.ipynb
3. Drag and drop one of the color wheel images into the bookstack editor; the image appears
4. Save the page; the image is not there
5. Click 'Edit' again, the image is there again, but saving again still doesn't show it
6. Delete the image, and try to drag and drop again; this doesn't work anymore and no image is shown

This second set of steps only works in Firefox, dragging from Chrome has the same result as trying to copy/paste, i.e. the image doesn't show up in the editor at all.

**Expected behavior**
Base64 encoded images can be copied or dragged into the editor, and are there after the page is saved.

 - Exact BookStack Version (Found in settings): v0.27.4
 - PHP Version: 7.1
 - Hosting Method (Nginx/Apache/Docker): Docker (FROM php:7.1-apache-stretch)
 - Browsers: Firefox 69.0.1, Chrome 77.0.3865.90
 - OS: Ubuntu 18.04

**Additional context**
This could be related to #1651, but seemed different enough to report separately.

3mmarg97:
This PR resolve #1697 
For more information about the edited option:
https://www.tiny.cloud/docs/plugins/paste/#paste_data_images

A screenshot for the editor after the PR:
image

### DIFF
--- a/resources/js/components/wysiwyg-editor.js
+++ b/resources/js/components/wysiwyg-editor.js
@@ -451,7 +451,7 @@ class WysiwygEditor {
             end_container_on_empty_block: true,
             statusbar: false,
             menubar: false,
-            paste_data_images: false,
+            paste_data_images: true,
             extended_valid_elements: 'pre[*],svg[*],div[drawio-diagram]',
             automatic_uploads: false,
             valid_children: "-div[p|h1|h2|h3|h4|h5|h6|blockquote],+div[pre],+div[img]",


3mmarg97:
It seems that (Pasting base64 images to the editor) works after this PR. But when I saved the page the image didn't appear.
So the PR doesn't fix #1697 completely. I will close this PR and try to figure out how to resolve the problem.

ssddanbrown:
Hi @3mmarg97, Don't worry about spending time on this. I think that issue may be related to a image paste handling issue that's already targeted for the next patch release.

I don't think I'd support a solution that resulted in Base64 content being stored in page data so please don't spend time going down that path.

Issue #1380, opened 2019-04-09
Ability to Customize Colors


blackshrub:
Especially for the page (now blue) and chapter (now orange).

ssddanbrown:
As of BookStack v0.27 the entity (Chapter, Book, Page, Shelve) colors are now set with CSS variables, which means it's fairly easy to change the colors of these via the Custom HTML head content setting:

```html
<style>
	:root {
		--color-page: #a242d7;
		--color-page-draft: #cfb732;
		--color-chapter: #db5382;
		--color-book: #534ecb;
		--color-bookshelf: #3eeaf0;
	}
</style>
```

I'm still considering making these easily alterable via the UI.

antimatteraccord:
> As of BookStack v0.27 the entity (Chapter, Book, Page, Shelve) colors are now set with CSS variables, which means it's fairly easy to change the colors of these via the Custom HTML head content setting:
> 
> ```
> <style>
> 	:root {
> 		--color-page: #a242d7;
> 		--color-page-draft: #cfb732;
> 		--color-chapter: #db5382;
> 		--color-book: #534ecb;
> 		--color-bookshelf: #3eeaf0;
> 	}
> </style>
> ```
> 
> I'm still considering making these easily alterable via the UI.

Please add these to the UI. 
Is it possible to also add an entry so I can change the main backgrounds of every page in this manner as well? That would appease a lot of the 'MOAR DARK MODE' people like myself.

ssddanbrown:
Dark mode would require a lot more than just a change of background colour, Would be an alternative style-sheet.

james-geiger:
@ssddanbrown Is this something that you'd welcome a pr to place in the settings ui?

> As of BookStack v0.27 the entity (Chapter, Book, Page, Shelve) colors are now set with CSS variables, which means it's fairly easy to change the colors of these via the Custom HTML head content setting:
> 
> ```
> <style>
> 	:root {
> 		--color-page: #a242d7;
> 		--color-page-draft: #cfb732;
> 		--color-chapter: #db5382;
> 		--color-book: #534ecb;
> 		--color-bookshelf: #3eeaf0;
> 	}
> </style>
> ```
> 
> I'm still considering making these easily alterable via the UI.

ssddanbrown:
@james-geiger Yeah, sure!

ssddanbrown:
@james-geiger has now added this feature in #1723 and it has been merged into master for the next release (v0.28).

[...]

PR #1723, opened 2019-10-17
Customize colors in settings

james-geiger:
This is a placeholder to address #1380 creating a UI and storing settings for the css color variables for page, page draft, chapter, book, and shelf.

### DIFF
--- a/app/Config/setting-defaults.php
+++ b/app/Config/setting-defaults.php
@@ -16,6 +16,11 @@ return [
     'app-editor'           => 'wysiwyg',
     'app-color'            => '#206ea7',
     'app-color-light'      => 'rgba(32,110,167,0.15)',
+    'bookshelf-color'      => '#a94747',
+    'book-color'           => '#077b70',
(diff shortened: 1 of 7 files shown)


james-geiger:
@ssddanbrown Do you see this having its own settings card or live within the existing card for customizing the app?

james-geiger:
This didn't take as long as I'd initially expected, largely due to the fact that the future was kept in mind when writing the settings controller.

I decided to, for now, leave the colors where they are in the "customization" section because I think that makes the most sense.

I did have to add a number of new labels/descriptions, so one downstream effect of this PR will be that a number of languages will be missing translations for those added keys.

ssddanbrown:
Awesome, Nice work @james-geiger. I'll review this soon. I would have said to live within the existing card anyway :+1:

ssddanbrown:
Thank you for this @james-geiger, All worked great.

Before merging I tweaked up the UI a little just to take up less space in the settings view:

image

Will be in the next release.

PR #1726, opened 2019-10-17
New setting - Enable inheritance of permission to a book when assigning to a shelf

philjak:
This PR adds a new setting to the settings page. After enabling this setting, a book that will be assigned to a shelf will automatically inherit the permissions of the shelf.

### DIFF
--- a/app/Entities/Bookshelf.php
+++ b/app/Entities/Bookshelf.php
@@ -105,7 +105,7 @@ class Bookshelf extends Entity implements HasCoverImage
     }
 
     /**
-     * Add a book to the end of this shelf. If shelve inheritance is enabled, copy permissions to book.
+     * Add a book to the end of this shelf. If shelf inheritance is enabled, copy permissions to book.
      * @param Book $book
      */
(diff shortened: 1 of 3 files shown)


sunnybeats:
Cant wait for this Setting to become available! :D

philjak:
Hey @ssddanbrown - sure. Understand this!

In our use case, BookStack will be used for a three-digit number of users. We therefore restrict the creation of shelves to a small number of users anyway in order to maintain clarity. 
The users will have their own shelves in their subgroups, which other subgroups will not see. At the same time, one user will be one of several sub-groups. Therefore it was important for us that a "normal" user does not have to worry about permissions.  

I absolutely understand that a more fine-grained setting, e.g. at shelf or book level, makes more sense. Unfortunately, I'm not yet deep enough in the BookStack code, so I can't offer such a pull request at this time. But I hope to be able to help you at a later time. 

So please feel free to close this pull request. 

Best,
philjak

ssddanbrown:
As per the discussion above, will go ahead and close this.

PR #1734, opened 2019-10-20
Update Korean translation

ististudio:
I filled in the missing parts, and I used more concise expressions throughout the translation to improve the consistency of terms.


ssddanbrown:
Thank you for offering this pull request @ististudio.

I used Google translate on a few of the terms that have been changed, translating both the old text and the new text, and to be totally honest, the translation of the old text seemed to often match the original English intent.

On the translations that you have altered, Are the previous translations wrong in any way or is it a matter of preference or style? Sorry to be questioning this but it's just a little tricky to understand the changes without any knowledge of the language itself.

ististudio:
Some are by preference, some are about orthography. Unlike English, Korean tends to delete subjects. In particular, repeating a subject can sound awkward for Korean readers. I removed the subject if I could see it from the context. e.g. 'email_confirm_success' => 'Authentication succeeded!'

Failure to distinguish "문서 만들기" from "문서 만듦" is a mistake in Google translation. Because "-ᄆ" adds meaning to the present completion, the latter means "The document has now been created." In addition, passive expressions that don't suit the Korean language have been changed to active expressions.

Anyway, I tried not to lose any information in given the context compared to the previous translation.

hhk7734:
I think that some part of this commit is weird.

I'm not sure that using noun is proper for notification.
'페이지를 만들었습니다.' is more familiar(?) to me than '문서 만듦'.
and how many people can understand '장철'?

ististudio:
Thank you for your opinion, @hhk7734.

I used sentences ending in a noun only in `activity.php` for notifications. These expressions will be used in the box that is displayed in the following screenshot.

Screen Shot

It disappears in a few seconds. Because of that, I wanted people to read this at once and I used as few characters as possible. These uses themselves are not so unfamiliar in my book. I have also seen them on the Android OS from time to time.

I've seen "장, 절, 장절" being used in [TeX](https://en.wikipedia.org/wiki/TeX), and I thought it was fine. But I agree that these terms look old. I'll change the term "장절" to "챕터" and "서적" to "책자".

If there are any other weird parts you think, I'd appreciate it if you let me know more.

hhk7734:
@ististudio 
I didn't know when the variable is used. After reading your explanation, I became to understand why you write that.

Others look fine.

ssddanbrown:
Thanks @hhk7734 for giving these a look and thanks again @ististudio for offering these updates.

Will merge now to be part of the next release.

Issue #1377, opened 2019-04-08
URL returned after login may be wrong (rev. proxy and alias used)


MatthieuParis:
**Describe the bug**
When both using a reverse proxy and hosting BookStack in a subdirectory, the URL that is computed after login may be wrong (and then lead to a non existent page).

The path used after login comes from session()->pull('url.intended', '/'); (in app/Http/Controllers/Auth/LoginController.php). This function returns the URL in the hosting server form, with is usually hidden and not explicitly set in the .env file.

Then the URL to redirect the user to is composed by this absolute path value and the base URL, in the function baseUrl (in app/helper.php). As it is not possible to guess the relative part in the path variable because the hidden base URL is not known, there are cases where this leads to errors.

**Steps To Reproduce**
1. Host BookStack in a subdirectory (for example 'test/'), on a special port, for example 10001;
2. Put in front of that server a reverse proxy that uses the beginning of the path (for example 'wiki/') to redirect to the BookStack hosting server;
3. Log in Bookstack and it will try to redirect to the page /wiki/test/test
4. This page does not exist.

**Expected behavior**
I understand that it is usefull to redirect, after login, to the initially requested page. But the method session()->pull('url.intended', '/') reflects the context of the hosting server, which can be quite different from the APP_URL that is declared in the .env file.

I don't know the Laravel framework but it may be possible to extract from the session() object the relative part of the path instead of the absolute path. With just the relative part, the helper function baseUrl should give the expected result.

**Screenshots**
-

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): BookStack v0.25.5
 - PHP Version: PHP 7.0.33-0+deb9u3
 - Hosting Method (Nginx/Apache/Docker): Nginx (rev. proxy) + Nginx (hosting) with configuration files included below.

[nginx-hosting.conf.txt](https://github.com/BookStackApp/BookStack/files/3056490/nginx-hosting.conf.txt)
[nginx-rev-proxy.conf.txt](https://github.com/BookStackApp/BookStack/files/3056491/nginx-rev-proxy.conf.txt)

**Additional context**
I would be glad to help in testing a patch. Thanks for your great work.

ssddanbrown:
Hi @MatthieuParis,

I've just deployed a new update, v0.26.2, that fixes some issues found in the URL redirect logic. It's worth updating to that and seeing if this remains an issue.

Let me know either way.

MatthieuParis:
[...]

PR #1743, opened 2019-10-25
Bugfix - takes APP_URL on redirect after login into account. Fixes issue #1377

philjak:
This PR strips the path of the APP_URL from the wrong url.intended string. This causes a correct redirect after login no matter if the APP_URL variable is set or not.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -110,7 +110,7 @@ class LoginController extends Controller
             $app_url = parse_url(env('APP_URL'));
             return redirect(preg_replace("/".addcslashes($app_url['path'], '/')."/", '', redirect()->intended('/')->getTargetUrl(), 1));
         } else {
-            redirect()->intended('/');
+            return redirect()->intended('/');
         }
     }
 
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @philjak,
Thank you for offering this pull request but I'm going to close it for the following reasons:

1. This pull request is targeted at one specific action (Login redirect) but I'd prefer there to be a fix to the base URL system if there is an issue since this kind of issue could appear elsewhere.
2. There is no confirmation this will fix the issue #1377, I think it's more nuanced than this.
3. Your current hosting scenario, as described in issue #1377, where the `public` folder is part of the request path, is not a setup that's supported and I'm assuming this pull request was created and tested based upon that scenario.
4. There's a lot of logic on a single line here that will run for the majority of BookStack user which I am not too comfortable with.
5. Minor point, but `env()` is not really to be used outside of the config files due to possible caching.

philjak:
Quick comment: 
1. You're right. I also faced this issue when moving a page to another book.
2. I think it would. The issue seems to be that the session variable "url.intended" (which you get redirected if you do redirect()->intended) has the APP_URL path included and in addition to that the actual intended page url.
3. This happens no matter if it's /public or another subdirectory.
4. Understand this.
5. You're right.

So I understand your decision. But maybe this still helps fixing this issue in future.

Best,
philjak

Issue #1161, opened 2018-12-04
Ldap import user avatar


lucassmacedo:
Import avatar users when login is from LDAP

lucassmacedo:
any idea?

ssddanbrown:
Hi @lucassmacedo,
Thanks for your request. 

> any idea?

This is not currently supported. I've only used LDAP in the scenario of testing BookStack if I'm honest.
What would really help such an issue is if you could provide details on how avatars are provided by your LDAP system. 

For Example:

* What LDAP system do you use?
* What attribute are avatars served on?
* What is the format of the URL provided? Absolute or relative?
* What file type are avatars provided in?
* What size(s) are avatars provided in?
* Are the answers above standard for your LDAP system?

lucassmacedo:
What LDAP system do you use?  **Active Directory**
What attribute are avatars served on? **thumbnailphoto**
What is the format of the URL provided? Absolute or relative? **(base64_encoded)**
What file type are avatars provided in? **(base64_encoded)**
What size(s) are avatars provided in? **(200x200)**
Are the answers above standard for your LDAP system? **Yes**

lucassmacedo:
https://github.com/Adldap2/Adldap2/blob/6853e2f277013883e900f943c5c80edb72c87a86/src/Models/User.php#L579

ssddanbrown:
LDAP user avatar import has now been merged as per #2320. 
This will be part of the next feature release and, at the time of that release, the docs will be updated to detail usage of this option. 

Thanks again @lucassmacedo for your request.

PR #1746, opened 2019-10-25
Bugfix - download and assign avatar when creating LDAP user in database

philjak:
Fixes issue #1161

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -98,6 +98,7 @@ class LoginController extends Controller
 
             $user->save();
             $this->userRepo->attachDefaultRole($user);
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @philjak,
Thank you for the pull request but I don't believe this addresses the issue fully? 
I'm fairly sure the intention is for the avatar to be retrieved and saved from the LDAP system, not gravatar as this pull request enables.

philjak:
Indeed, thats not pulling any avatar out of the LDAP system. But it fixes an issue, where even if you specify a particular AVATAR_URL, the avatar wasn't downloaded before.

ssddanbrown:
@philjak Ah, Cool, Will therefore merge for next release.

PR #1752, opened 2019-10-28
Feature: Move a page directly into a chapter of another book

philjak:
Currently, when you move a page, you get a list of Books and chapters. Unfortunately, if you choose a chapter as new destination for a page, you get an error, since currently only books as target are supported.

This PR allows you to move a page directly into a chapter by checking, if the parent is a Chapter. If so, the page will be assigned to the target chapter and moved to the corresponding book.

### DIFF
--- a/app/Entities/Repos/PageRepo.php
+++ b/app/Entities/Repos/PageRepo.php
@@ -307,15 +307,13 @@ class PageRepo
         }
 
         if ($parent instanceof Chapter) {
-            $parentChapter = $parent;
-            $parent = $parent->book;
-            $page->chapter_id = $parentChapter->id;
-            $page->save();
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for spotting this @philjak and for providing a pull request. Moving into a Chapter was, and has been, intended functionality so this was more of a bug caused by my refactor earlier in this release cycle. Nice to get these caught before release :sweat_smile: . Just as a warning, you may come across a few of these kind of bugs due to that refactor since I have not done in-depth testing yet.

I also committed 6a1b6a97f992df29d91434fe04c69c6bac8ba4f7 and d64c358c4f75533da5f66557c52e989a6d4ad1f0 to add tests and cover the scenario of moving a page from a chapter to a book.

PR #1754, opened 2019-10-29
Feature - Show pages and chapters in shelf view

philjak:
This PR adds a setting to the settings page, which allows user to toggle between a detailed and the regular shelf view. In the detailed view, all pages and chapters of a book will be shown (similar to the overview of all shelves, where the books are shown). 

I think this is very useful, if you use BookStack as a wiki, because it reduces the amount of clicks until a user finds the right page or chapter.

image

### DIFF
--- a/resources/lang/en/settings.php
+++ b/resources/lang/en/settings.php
@@ -40,6 +40,9 @@ return [
     'app_disable_comments' => 'Disable Comments',
     'app_disable_comments_toggle' => 'Disable comments',
     'app_disable_comments_desc' => 'Disables comments across all pages in the application. <br> Existing comments are not shown.',
+    'app_show_pages_in_shelf_view' => 'Detailed shelf view',
+    'app_show_pages_in_shelf_view_desc' => 'Shows all pages and chapters of a book inside the shelf view.',
+    'app_show_pages_in_shelf_view_toggle' => 'Enable detailed view',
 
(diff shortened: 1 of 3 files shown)


philjak:
@ssddanbrown I know you won't like that I put this into the settings menu. But I really like the view. If you also think this detailed view is a good idea, one could also think about using this as a third view toggle. "grid", "list", "detailed".

ssddanbrown:
Thanks for offering this change @philjak and sorry for my late reply.

I'm going to close this as I personally don't feel that this solution is right for the project. 

As you predicted, I'm not a fan of introducing a new system setting to control view presentation for a single particular view of the app. Toggle option is better but a three-option toggle is a bit awkward. Either way it adds an extra, quite intensive, additional view that we'd have to consider during testing and development.

The listing of child content like that, where there's various different types, in a potentially confusing order, can make the view seem confusing and intimidating, I have to keep in mind the less technically inclined. There are also performance and size concerns here. I know we have a similar view in the shelves view with books, but book content is a lot more variable than the relatively low-quantities of books and shelf child books can be fetched much more efficiently. 

On the whole I don't think just showing extra levels of depths within various current views is the correct answer here. I hear the need from more advanced/technical users to have deeper traversal/navigation but I think that may be better of achieved via other means, potentially by a purpose built interface but that needs some careful consideration.

PR #1755, opened 2019-10-30
Feature - allow toggling between grid and list view in shelf view (shelves.show)

philjak:
This PR adds the list/grid view toggle button to the shelves.show view. This button is already available in the shelves.index as well as in books.index.

This allows users to use the desired view even when looking directly into a shelf.


ssddanbrown:
Apologies for my late review and thanks for offering this pull request @philjak.

Have now added tests to cover and merged to be part of the next feature release. Only thing I changed was to separate the setting out to be stored separately from that of the books-list view, Just for users that want to keep these views different.

PR #1756, opened 2019-10-30
Feature - add the "Create Shelf" resp. "Create Book" to the home view

philjak:
This PR adds the "Create Shelf" button to the Shelves-Homepage and "Create Book" button to the Books-Homepage definied in the settings menu.

Assume you've set the homepage to "shelves". With this PR, there is no need to click on "Shelves" in the top menu first (which is for the user basically the same view) in order to create a shelf.

In our case, some users were confused why they were not allowed to create a shelf even if they had the permission. This was only caused by the missing button in the home view.


ssddanbrown:
Thanks @philjak for the PR and the usage explanation, Now merged for the next feature release.

PR #1762, opened 2019-10-31
Update Italian translation

dellamina:
Added missing translations and changed some others that were badly written.


ssddanbrown:
Thank you @dellamina for offering these updates.

Will merge in now for the next feature release.

PR #1779, opened 2019-11-14
Fix overflow issue of page content

LeonLiuY:
Currently a long table in a page will overflow the white card background, since `overflow: initial` means `overflow: visible`.

### DIFF
--- a/resources/sass/_blocks.scss
+++ b/resources/sass/_blocks.scss
@@ -180,7 +180,7 @@
   margin-left: auto;
   margin-right: auto;
   margin-bottom: $-xl;
-  overflow: initial;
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this @LeonLiuY and apologies for my really late reply.

Unfortunately I don't think this is the right solution to the problem.
Although this prevents such pages from looking broken I think it's more important that the content is accessible to users rather than being completely cut off. Additionally, this would interfere with floating elements such as the content selection popup.

Ideally, I want to implement to a long-term solution that hides overflowing content while keeping it accessible. 

Therefore I'll close off this pull request.

PR #1790, opened 2019-11-19
Code block display in details tag

### LINKED ISSUE #781: Markdown - problem with <details> section
### For Bug Reports

* BookStack Version 0.20.3
* PHP Version: 7.1
* MySQL Version: I'm' using MariaDB (10.3.5)
* Browsers: Chrome:  65.0.3325.181 (64-bit) / Firefox: 59.0.2 (64-bit)
* Operation System: Windows 10 - for browsers, testing and displaying
* Ubuntu 16.04.4 with Docker 18.03-ce for docker-compose.

Here is my docker-compose file which I'm using: https://bitbucket.org/pogosoftware-vagrant/bookstack/src/886bc7d050e89f49559848911205ea6f8a8bfe28/bookstack/bookstack.yml?at=master&fileviewer=file-view-default

##### Expected Behavior
When I'm using code block in the &lt;details&gt; section it should properly display and format text.

##### Current Behavior
It seems something is wrong with CSS styles when I'm using  &lt;details&gt; block with the code section. When I uncollapse section then is see only empty block code 

https://www.screencast.com/t/5XmajwY662j

and when I click on it then the content appears but it is broken because the number of lines is moved to the right.

https://www.screencast.com/t/pDf1ySG0wiv

##### Steps to Reproduce
First set Page editor to Markdown
https://www.screencast.com/t/I4Fx1sfXa7GD

Next, create book and page and paste a code

&lt;details&gt;
&lt;summary&gt;1. Pull httpd image. &lt;/summary&gt;

&#96;&#96;&#96;
docker pull httpd
&#96;&#96;&#96;
&lt;/details&gt;

In the preview window, everything looks fine. But when you save it then this bug occurs.

This is not a very critical bug but it will be great if it will be work :D.

Best regards
Przemek

james-geiger:
Fixes #781.  Use codemirror display/autoReload addon to watch for state changes and resize the codemirror blocks if they were hidden when the DOM was loaded.

> This is done by polling every 250 milliseconds (you can pass a value like {delay: 500} as the option value to configure this).
>
> [codemirror docs](https://codemirror.net/doc/manual.html#addon_autorefresh)

I've set this to poll every 50ms, which was a bit smoother transition.  A 1ms poll, just for fun, produced the best result but is likely to cause performance issues.

### DIFF
--- a/resources/js/services/code.js
+++ b/resources/js/services/code.js
@@ -27,6 +27,7 @@ import 'codemirror/mode/yaml/yaml';
 
 // Addons
 import 'codemirror/addon/scroll/scrollpastend';
+import 'codemirror/addon/display/autorefresh'
 
 const modeMap = {
     css: 'css',
@@ -104,7 +105,10 @@ function highlightElem(elem) {
         lineNumbers: true,
         lineWrapping: false,
         theme: getTheme(),
-        readOnly: true
+        readOnly: true,
+        autoRefresh: {
+          delay: 50
+        }
     });
 
     addCopyIcon(cm);


ssddanbrown:
Thanks for offering this @james-geiger and sorry for my really late response.

To be honest I'd like to avoid polling unless it's really necessary, just because polling UI elements can build up and slow the UI quite quickly or have unintended effects.

Looks like a details element fires a `toggle` event when opened. Could we instead watch for that within page-content then manually call a `refresh()` on any codemirror instances within? The `toggle` event doesn't appear to bubble but can still do this with a single event listener and the `useCapture` parameter.

ssddanbrown:
This has now been addressed in 5404f22bf967a6d597bffbbc7a683e64f9525a03 so will therefore close this.

Thanks @james-geiger for your original PR here.

PR #1794, opened 2019-11-20
RTL support

abublihi:
Adds RTL full support by transforming the styles to RTL using [rtlcss](https://github.com/MohammadYounes/rtlcss/) and [rtlcss-webpack-plugin](https://github.com/wix/rtlcss-webpack-plugin).

All css files will have the extension .rtl
exmaple : `styles.css` will be `styles.rtl.css`


ssddanbrown:
Thanks for offering this @abublihi.

I'll be honest, I thought the browser would auto-decide the default document direction but it looks like that's probably not the case so I can see why adding that is a good idea.

Would we need to consider mixed page content at all? Maybe in cases where the system has english content but used by a rtl-language user? In these cases would we need to set such content areas as `auto` instead?

In regards to using rtlcss, I can see that it would provide a quick fix but I'd prefer to attempt to support RTL properly and learn where potential issues lie instead of trying to auto-patch with parallel css builds and an extended build chain. 

It looks like [CSS logical properties](https://caniuse.com/#feat=css-logical-props) are now well supported within the browsers BookStack is intended for so switching to them could be a good start.

ssddanbrown:
Closing in favour of #2003

PR #1795, opened 2019-11-20
BookStack-Install-Script-Debian-10.sh

th3r3isnospoon:
This script is based on the script from BookStack (https://www.bookstackapp.com/docs/admin/installation/#ubuntu-1804) It has been modified to install on Debian 10.2.


ssddanbrown:
Thank you very much for offering this script @th3r3isnospoon but this repo is used for the core application code only. We have a separate repo for scripts and config:

https://github.com/BookStackApp/devops/tree/master/scripts

I'd be happy to accept your script there, after testing in a VM, but I'll have to close this pull request since we don't track/store the install scripts here.

PR #1796, opened 2019-11-21
add ar_alt local

abublihi:
Update `ar` language translations, add new local `ar_alt`.

In the ` resources/views/pages/show.blade.php` the Actions was not loaded form translations.


ssddanbrown:
Hi @abublihi, Thank you for providing these updates.

Apologies if this question appears ignorant, since I'm very unfamiliar with Arabic text. Could you explain the need for an alternative Arabic translation? 

I just need to know if this is for a one-off need or if this is a common and functional version of the language that others will use. Would prefer to keep the translation count at a minimum to reduce support overhead where possible.

abublihi:
Your question seems very important, yes the alternative translation may be for only a one-off need, but i added it because i think that some one **may**  use it.

So you can **ignore this pull request** it is not very important, 

but **please**  check this pull request #1794 it is very important.

ssddanbrown:
@abublihi Okay, Thanks for confirming. In that case I will close this pull request since I think it may complicate translation management.

If there are non-alt translations here feel free to provide them as a separate pull request.

Just to let you know, The next release will bring the ability to override individual translations via a custom theme directory, which may be helpful to your use-case. The functionality of this system is described here: #1749.

I have seen #1794 so will review that soon but I'll need some time to understand exactly what it does. Will probably review after the next release has deployed.

PR #1797, opened 2019-11-22
Modify Chinese language pack

qianmengnet:
Modify Chinese language pack


ssddanbrown:
Thank you very much once again @qianmengnet for these updates. I have now merged them to be in the next release. Just to warn you, some of the text may appear different to your translations since there was an earlier pull request with Chinese translations, so these changes had to be merged with that.

PR #1804, opened 2019-11-28
Update of polish language

artskoczylas:
Update of all texts in polish language, change translation of book from 'Podręcznik' to 'Książka'. I think it better reflects the meaning of the word.

If I made a mistake, please correct me.


ssddanbrown:
Thank you @artskoczylas for providing these updates. Will merge them in now for the next release.

PR #1808, opened 2019-11-30
added port for bookstack-test

richardhulbert:
Hi Dan
 
For those of us who might test using a mySQL db on a port other than 3306

Richard

### DIFF
--- a/app/Config/database.php
+++ b/app/Config/database.php
@@ -82,6 +82,7 @@ return [
             'driver'    => 'mysql',
             'url' => env('TEST_DATABASE_URL'),
             'host'      => '127.0.0.1',
+            'port'      => $mysql_port,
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @richardhulbert,
Thank you for offering this pull request.

I'd prefer not to share any variables across testing and non-testing config, especially in this case where there's additional logic to parse out the port for backwards compatibility in production, to prevent confusion when actively testing. Therefore I'm going to close off this pull request without merging.

As of the current master branch, There is a `TEST_DATABASE_URL` env option available. You can set this in the format `mysql://username:password@host:port/database` which allows you use a custom port and even a custom database name. This will probably be the preferred approach to customising these testing options from now on.

Issue #1706, opened 2019-10-10
Allow following deeplink after login


JtheBAB:
**Describe the feature you'd like**
Unsure if bug or feature request. 

When a user send a Bookstack link that is deep inside a book that requires a login, then it doesn’t work currently. You get a page not found and when you login through the upper right login button then you are on the default page. When you click again to the link you come to the right page. 

**Describe the benefits this feature would bring to BookStack users**
When the user can login and then got access directly to the requested page, then the general usability would be improved.

cw1998:
From my understanding, this is intentional. If a page isn't public (ie, you must log in to view it), then the URL should not reveal to the public that it is valid by showing a login form instead of a "Not found" page.

JtheBAB:
I understand that. But the page could say 

„page not found or not enough permission“ 

and offer the login form. Then it is still unclear if the page really exist.

ssddanbrown:
@JtheBAB I'm assuming you have public viewing enabled on your instance?

What you describe is the behaviour that should occur but things get a little complicated when public viewing is enabled as the user's intended desired location gets harder to track. Pretty sure it's been discussed before but having trouble finding the history of that.

JtheBAB:
Yes, public is enabled.

ssddanbrown:
As part of #1817 and 7f6cbead333415c9ead093827942a0e08116847a, This has now been implemented in master to be in the next patch release.

The 404 page has been updated to indicate such a scenario could be due to permissions, and "Login" has been added as an action on that card to encourage that as the user's next step. 

![image](https://user-images.githubusercontent.com/8343178/76688238-f4311500-6622-11ea-93ff-74d423c83e77.png)

PR #1817, opened 2019-12-14
Recall intended route when manually clicking login

Xiphoseer:
Currently, for a non-public BookStack instance, the [Authenticate Middleware](https://github.com/BookStackApp/BookStack/blob/02af69ddf29e0a2d03cc3a8660bfca9a65175afc/app/Http/Middleware/Authenticate.php#L44) sets the `url.intended` session key when accessing any link while not logged in, which tells the [login procedure](https://github.com/BookStackApp/BookStack/blob/02af69ddf29e0a2d03cc3a8660bfca9a65175afc/app/Http/Controllers/Auth/LoginController.php#L110) to redirect there after login.

A public BookStack instance will not exhibit this behavior for pages which require additional permissions, as a common 404 is returned (which is the correct thing to do). With these changes, the view at `/login` recognizes an optional query parameter `intended` and sets the intended redirect URL when present.

Our use-case is sharing links to a partially public BookStack instance with our team via E-Mail which they currently need to click twice if they are not already logged in.

In the future it may also be useful to redirect a not-logged-in user to the login page with `intended` set instead of the root when accessing URLs like `/settings`.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -128,6 +128,10 @@ class LoginController extends Controller
             ]);
         }
 
+        if ($request->has('intended')) {
+            redirect()->setIntendedUrl($request->get('intended'));
+        }
+
         return view('auth.login', [
           'socialDrivers' => $socialDrivers,
           'authMethod' => $authMethod,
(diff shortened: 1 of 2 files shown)


Occyss:
This would resolve #1706

ssddanbrown:
Thank you for providing this pull request @Xiphoseer, and sorry it has taken me a few months to respond. This is now merged to be part of the next patch release.

I made some tweaks in 7f6cbead333415c9ead093827942a0e08116847a, just to add tests & to take the intended URL from the previous page instead of needing to pass as a query param.

PR #1819, opened 2019-12-17
update translation

johnroyer:
Add few word which was not translated.


ssddanbrown:
Thank you @johnroyer for providing these updates. Will merge in now for the next release.

PR #1827, opened 2019-12-29
Hebrew Translation

Binternet:
Support for the Hebrew Language


ssddanbrown:
Thank you very much for these translations @Binternet.
I have merged them in, ready to be part of the next patch release.

PR #1830, opened 2020-01-01
Implement GraphQL Endpoint

cipi1965:


ssddanbrown:
Hi @cipi1965,
Thanks for opening this pull request.

I have already started #1826 which aims to set the foundations for a REST API, to be expanded upon in subsequent releases. For the sake of maintenance efforts, I think I'd want to avoid having both REST and GraphQL endpoints available to access data.

Out of the two, I'd be looking to stick with a REST API unless there are major, clear overall benefits of instead using GraphQL.

cipi1965:
Hey @ssddanbrown i saw your pr.
I think they are two completely different type of apis, i would consider to keep both.

cipi1965:
Authentication instead will be equal to REST one

ssddanbrown:
> I think they are two completely different type of apis, i would consider to keep both.

While I agree they are two different types of APIs, The functional scenario where they would be utilised will be extraordinarily similar; In my mind they effectively serve the same purpose but in two different ways.

There'd need to be a clear major functional benefit achieved by having REST and GraphQL in order to warrant implementing and, more importantly, supporting both implementations going forward.

cipi1965:
@ssddanbrown please check these links:


https://help.shopify.com/en/api/getting-started/shopify-and-graphql/graphql-benefits

https://www.shopify.com/partners/blog/graphql-vs-rest

https://github.blog/2016-09-14-the-github-graphql-api/

https://www.graphql.com/articles/graphql-at-github

ssddanbrown:
@cipi1965 None of those links provide any insight into why supporting both REST and GraphQL would be beneficial. Both of those platforms do support both but that's more likely due to supporting a stable platform and not breaking existing applications.

I think it'd be either one or the other. Just putting out some thoughts:

### REST

\+ Better aligns with the existing app structure & endpoints.
\+ Is well supported by the core framework (Laravel) without the need for additional dependencies.
\+ Would be easily accessible to developers due to more broadly known and by the fact it can be easily browsable via browser get requests.
\+ Is more familiar to the core project maintainer.
\- Less efficient in client consumption and network transfer.

### GraphQL

\+ More efficient in transfer and client consumption.
\+ More pleasant for developers that know GraphQL.
\- Less established overall; Hasn't been around for long and lower developer familiarity.
\- Requires external dependencies, or a lot of extra build & maintaining effort.

james-geiger:
I'd echo @ssddanbrown in saying that while GraphQL might be technically more efficient from a data standpoint, I can't imagine a scenario in which the amount of data that users would be consuming from an API endpoint in BookStack would benefit from those performance efficiencies.  Frankly, GraphQL would be overkill.

Additionally, multiple half-baked APIs would likely cause confusion among users and could open the platform to security issues.

Although local users may find some benefit in integrating a roll-your-own GraphQL endpoint, my opinion as someone who has BookStack deployed in an enterprise environment would be to focus on implementing a REST API.

cipi1965:
I don't know, i would take responsibility to maintain GraphQL apis. I think REST apis are better for automation, GraphQL one for making data driven applications.

ssddanbrown:
Thank you for your input @james-geiger, It's good to hear another voice on the matter.

---

> i would take responsibility to maintain GraphQL apis

@cipi1965 That's very generous that you're offering responsibility for the GraphQL APIs but I'd still fundamentally be responsible for anything that's part of the project. Additionally, I don't think it'd be that simple; Potential modification to such APIs may be needed on the back of other changes or additions to the project, For me to assume that you'll be able to watch and keep on top of any such changes is something I'm not really comfortable with. Nothing personal but I just don't know you well enough nor am I comfortable with that dynamic being introduced into the project.

cipi1965:
@ssddanbrown ok closing this pr

Issue #1525, opened 2019-07-05
Update notifications to make dismissal more intuitive


flonology:
**Describe the bug**
While editing a draft, the dialogue stays over the "Save Page" button and does not go away

**Expected behavior**
The dialogue should go away or stay in a place which lets the "Save Page" be visible.

**Screenshots**
![Bild eingefügt am 2019-7-5 11-16](https://user-images.githubusercontent.com/1073814/60712623-b5857f80-9f17-11e9-93d2-d213e1ef1a38.png)

**Your Configuration:**
 - BookStack Version: 0.26.2
 - PHP Version: ?
 - Hosting Method (Nginx/Apache/Docker): Docker

ssddanbrown:
Hi @flonology,
Does the message not go away if you click on it?

Success messages are made to auto-disappear, Since their purpose is affirmation of an event and does not have any consequences whereas warnings & errors will stay around until dismissed since it's more important that they're read.

flonology:
Hi, sorry for the late reply – yes, it disappears when I click it! I just wasn’t aware that I could. After knowing that it seems so obvious (cursor changes to hand etc.). Maybe I was too focused on finding any X (to close) but I do not see any problem anymore ;-)

ssddanbrown:
@flonology Thanks for the response.

It's probably a good idea if we add a close icon to the notification, To make it obvious it can be dismissed. I'll re-open this and mark it for the next feature release.

james-geiger:
@ssddanbrown Is this still open?  Happy to take it.

ssddanbrown:
@james-geiger Yeah, Still open, Go for it, would be a big help. Thanks!

Have assigned to you but no pressure, Just shout if you want to be un-assigned for any reason.

james-geiger:
I've been playing around with this, and it seems like a dedicated dismiss button would be "just for show" unless we removed the click-to-dismiss feature from the notification and forced users to click the close button.

My thought to keep things looking clean is just to add auto-hide for errors and warnings with a 5-10 second timeout.

Thoughts, @ssddanbrown & @flonology?

ssddanbrown:
@james-geiger I'd totally vote for the dismiss button that's "just for show". Although it would not actually function itself, it would still act as an intuitive target for the user to click.

I'm generally adverse to auto-hiding warning and error messages since they should generally be things the user should be aware of, and I personally find it really frustrating when a message zooms off while I'm in the middle to trying to understand what it says.

[...]

PR #1845, opened 2020-01-14
Add close icon to notifications

SoarinFerret:
In relation to #1525

This adds a simple close icon to all notification panes. Here is an example of how it looks:

image

Per the comments on the bottom of the issue thread, it doesn't actually do anything, besides provide the user a visual cue reminding users to click to dismiss the notification.

### DIFF
--- a/resources/sass/_components.scss
+++ b/resources/sass/_components.scss
@@ -15,7 +15,7 @@
   transition: transform ease-in-out 280ms;
   transform: translateX(580px);
   display: grid;
-  grid-template-columns: 42px 1fr;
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thank you @SoarinFerret for this pull request, I'll merge in now for the next release. 

Just a note for the future, not a big issue here since this was a relatively small update, but if someone is assigned to an issue it's generally a good idea to comment on the issue to see if that person is intending to continue work on it. I probably need to make it clear in the readme that assignment is used to prevent work overlap. 

Thanks again though.

PR #1862, opened 2020-01-28
Shared drafts

Hazzard17h:
Added settings flag to enable shared drafts across all users with edit permission on the page.
Also added _editdraft_ permission, that allow users to edit pages and only save draft of them (if they not have also _update_ permission).

### DIFF
--- a/app/Auth/Permissions/PermissionService.php
+++ b/app/Auth/Permissions/PermissionService.php
@@ -426,6 +426,9 @@ class PermissionService
         if ($entity->isA('book')) {
             $baseActions[] = 'chapter-create';
         }
+        if ($entity->isA('page')) {
+            $baseActions[] = 'editdraft';
+        }
         return $baseActions;
     }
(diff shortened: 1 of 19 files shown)


ssddanbrown:
Thank you for offering this pull request @Hazzard17h.

Unfortunately I would not look to include such functionality within BookStack due to the following reasons:

1. I don't really want to complicate the "Draft" model within BookStack at all. The unique feature of the draft is that it is personal. Quite often the "Draft" stage would be one of active working and, in a non-live-collaboration platform such as BookStack, having such content be personal can be useful. Adding variation makes the draft feature harder to support, document & maintain over time.
2. This features could be some-what replicated using existing in-system features, such as tagging all such pages with an "in-progress" tag or adding them to an "in-progress" chapter.
3. There has been little demand of such a feature from existing BookStack users. 

My apologies that this is likely not the response you wanted to hear.

PR #1881, opened 2020-02-08
Log failed accesses

### LINKED ISSUE #728: Log failed accesses
Hello,

I think it would be really nice for BookStack to log failed authentications.
We would then be able to monitor these messages and ban malicious IPs, at firewall level, with for example Fail2Ban on a Linux server.

An easy and nice solution would simply be to log a message into the webserver log for each failed authentication attempt or misbehaviour, using for example PHP `error_log()` function.

Thank you very much 👍 

Ben

benrubson:
Hi,

This PR logs failed login attempts for further processing by tools such as fail2ban.
It then closes #728.

Thank you 👍

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -267,3 +267,9 @@ API_MAX_ITEM_COUNT=500
 
 # The number of API requests that can be made per minute by a single user.
 API_REQUESTS_PER_MIN=180
+
+# Failed access
+# message to log into webserver logs in case of failed access, for further processing by tools like Fail2Ban
+# Apache users should use : user "%u" authentication failure for "BookStack"
+# Nginx  users should use : user "%u" was not found in "BookStack"
+FAILED_ACCESS_MESSAGE=''
--- a/app/Actions/ActivityService.php
+++ b/app/Actions/ActivityService.php
@@ -183,4 +183,21 @@ class ActivityService
             session()->flash('success', $message);
         }
     }
+
+    /**
+     * Log failed accesses, for further processing by tools like Fail2Ban
+     *
+     * @param username
+     * @return void
+      */
+    public function logFailedAccess($username)
+    {
+        $log_msg = config('logging.failed_access_message');
+
+        if (!is_string($username) || !is_string($log_msg) || strlen($log_msg)<1)
+            return;
+
+        $log_msg = str_replace("%u", $username, $log_msg);
+        error_log($log_msg, 4);
+    }
 }
--- a/app/Config/logging.php
+++ b/app/Config/logging.php
@@ -79,4 +79,9 @@ return [
         ],
     ],
 
+    // Failed Access Message
+    // Defines the message to log into webserver logs in case of failed access,
+    // for further processing by tools like Fail2Ban.
+    'failed_access_message' => env('FAILED_ACCESS_MESSAGE', ''),
+
 ];
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -2,6 +2,7 @@
 
 namespace BookStack\Http\Controllers\Auth;
 
+use Activity;
 use BookStack\Auth\Access\SocialAuthService;
 use BookStack\Exceptions\LoginAttemptEmailNeededException;
 use BookStack\Exceptions\LoginAttemptException;
@@ -102,7 +103,7 @@ class LoginController extends Controller
             $this->fireLockoutEvent($request);
 
             // Also log some error message
-            $this->logFailedAccess($request);
+            Activity::logFailedAccess($request->get($this->username()));
 
             return $this->sendLockoutResponse($request);
         }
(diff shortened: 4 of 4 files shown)


benrubson:
Hi @ssddanbrown, may I ask you to review this when U'll have some time ?
Many thanks 👍

benrubson:
I've made the failed authentication message an option, so that it's user configurable.

Let's have a look at this @ssddanbrown please ?
Many thanks 👍

ssddanbrown:
Hi @benrubson,
Apologies for my lack of response, Didn't mean to ignore this pull request but I've always been in the middle of something else when coming across it.
Thanks for offering this pull request.

It'd need some changes before going into a release:
* The `env` function should only be used from `app/Config/` files so any options will need to be redefined there (Maybe in app/Config/logging.php?) and then accessed via `config('logging.failed_access')` (Or whatever key this results as). 
* I'd probably want to move the `logFailedAccess` into the `ActivityService.php` file just so we don't have logging logic hidden deep within a controller, and since we'll likely have other activity based logging in the future.  
* I'm not sure the error_log will always be the best location since that could vary depending on system, making it harder to document & support on. Additionally that should probably be just reserved for error usage. It might just be easier to have a set file in the `storage/logs` folder since this dir should be writable. Laravel has some great built-in logging support so we can just extend that. Will be a case of adding a extra channel to `app/Config/logging.php`, following the `single` channel as an example. Then we can write directly to that channel. https://laravel.com/docs/6.x/logging#writing-to-specific-channels
* For the purposes of fail2ban I'm assuming we will need an IP address to be part of the log? I'm guessing it's not in the current message since it's expected that error_log will log out to the webserver log which would include the IP. 
* Is there a common format for Fail2Ban logs? Or is it usually a case of providing the pattern and location to Fail2Ban? 
* We'll need to cover this with testing, Using the laravel Logging functionality will make that easier.

Sorry if the above is fairly incoherent, Am writing this fairly late. Happy to explain any parts of this further. Otherwise I can finish this off or do parts of the above if you're not sure but might just be an extra while before I get round to it.

benrubson:
Hi @ssddanbrown,
No worries, thank you very much for your detailed feedback.
And no incoherence there :)
Below are my answers to your 6 points.

1. I just pushed a fix accordingly.
2. I just pushed a fix accordingly.
3. This PR is intended to log failed login attempts, in order to process them with other tools, especially Fail2Ban.
Fail2Ban has webservers jails (for Apache, Nginx etc...) which inspect their error logs for specific patterns, among them auth failure messages (here's for example the [Apache filter](https://github.com/fail2ban/fail2ban/blob/0.10/config/filter.d/apache-auth.conf)).
Idea with this PR is then to take benefit from these jails, so that we don't have to define an additional one, inspecting an additional log file etc... This to save some configuration efforts, but also some system resources.
`error_log` will then log into the webserver error log. With the correct pattern, these messages will be caught by the Fail2Ban webserver jails, which is what we want.
You'll notice that in the `.env.example.complete` file, I gave some patterns for both Apache & Nginx. They match the existing Fail2Ban jails.
This method has been chosen and is used in other webapps such as [Baikal](https://github.com/sabre-io/Baikal/pull/874) and [phpIPAM](https://github.com/phpipam/phpipam/pull/2932).
4. You're absolutely right, and point 3 above confirms this.
5. I think point 3 already answers this too :)
6. I'm not sure what kind of test you would like to implement. Fail2Ban already performs CI tests to be sure all its filters correctly work.

Thank you again :+1:

benrubson:
Hi @ssddanbrown, some news perhaps on this topic ?
Many thanks 👍

benrubson:
Many thanks Dan 👍

ssddanbrown:
Hi @benrubson,
Sorry for the delay again, Just dedicated the morning to getting this done.

Thanks for making the above updates, I've made some further updates in 2ed031712918313b50483d22cf6735aed227dc06, mainly to add testing coverage and to route the logging via Laravel's logging system, which has made it easy to change logging channel if needed (So you could instead choose stdout or the bookstack log for this). 

Just a warning since it may effect your setup if you're actively using your code, I have changed the env option name to `LOG_FAILED_LOGIN_MESSAGE` just to be a bit more focused on the intended use-case.

Now merged, to be part of v0.30.

benrubson:
Perfect, really nice modifications :+1:
Thank you again for having taken time to look at this, and for having merged it.
Good news !

PascalMinder:
Will the error message contain the clients ip?

Currently I am using the following regex in fail2ban:
`failregex = ^.*Username or password is incorrect\. Try again\. IP: <ADDR>\. Username:.*$`

And in the Bookstack config I use
```
# user identifier (Username or email).
LOG_FAILED_LOGIN_MESSAGE='Username or password is incorrect. Try again. IP: %ip. Username: %u'
LOG_FAILED_LOGIN_CHANNEL=errorlog_plain_webserver
```

But of course the `%ip` placeholder does not work. For fail2ban to work it wood need an IP?

benrubson:
Your error messages go into the webserver log, thanks to the `LOG_FAILED_LOGIN_CHANNEL=errorlog_plain_webserver` option, so you should get client IP there, automatically prepended by the webserver itself.

PascalMinder:
> Your error messages go into the webserver log, thanks to the `LOG_FAILED_LOGIN_CHANNEL=errorlog_plain_webserver` option, so you should get client IP there, automatically prepended by the webserver itself.

Okay, thanks for the explanation. Looking forward to the v0.30 release

PR #1889, opened 2020-02-12
Fixed 'interaction_required' response for azure

ch0wm3in:
Azure Conditional Access policy 2FA returns 'interaction_required' 400 response and therefor breaks authentication when forced to use 2FA. 
https://github.com/SocialiteProviders/Providers/issues/208


MikeyMJCO:
In cases where IP based conditional access MFA configurations are used in Azure AD this does indeed fix the issue. Tested on current beta.

ssddanbrown:
Thank you for this @ch0wm3in and thanks for testing it @MikeyMJCO.
Tested on my dev instance, which did not have this issue, and did not cause any side-effects there (Thinking of existing users),  so now merged to be part of the next patch release.

Issue #1897, opened 2020-02-14
Cover images selected in create dialog aren't stored


alexschroth:
**Describe the bug**

Cover images, that are selected while you create a shelf or a book, do not appear after closing the dialog, i.e. if you visit the shelves / books section, cover images are not shown.

Workaround: Open the edit dialog after you've created the book / shelf and upload the image (again).

**Steps To Reproduce**

1. Go to 'Shelves'
2. Click on 'New Shelf'
3. Add a title and upload a cover image
4. Click on 'SAVE SHELF'
5. Go to 'Shelves'

**Expected behavior**

Previously uploaded cover image is shown.

**Configuration**

 - Exact BookStack Version (Found in settings): 0.28.0
 - PHP Version: 7.3.11
 - Hosting Method (Nginx/Apache/Docker): nginx (+ php-fpm)

sunnybeats:
Same thing,
I was able to reproduce the bug like described.

TBK:
Which browser + version + OS do you use?

alexschroth:
> Which browser + version + OS do you use?

Safari 13 on macOS 10.15.3

TBK:
The issue is https://github.com/BookStackApp/BookStack/blob/33ef1cd4fa2e4f8b988658d0d244b2dd67ba09a2/app/Http/Controllers/BookshelfController.php#L93

which should be
```
$this->bookshelfRepo->updateCoverImage($shelf, $request->file('image', null)
```

I will provide a PR.

sunnybeats:
I changed my 

BookStack/app/Http/Controllers/BookshelfController.php

To the changes you have written and it seems to work 👌🏻

ssddanbrown:
Patch now applied in master therefore I'll close this. The fix will be in the next bugfix release.

---

@sunnybeats Beware that editing core BookStack files will likely cause conflicts when you next update. Reverting changes before updating is usually best.

sunnybeats:
Thanks, i copied back the file i have backuped before. So its like before the changes kow.

So i wait for the update now.
Thank you for this great app and this great suppord!

ssddanbrown:
@sunnybeats I deployed the update last night so you should be able to just update to get the fix.

sunnybeats:
@ssddanbrown 
Updated and its working!
Thanks

PR #1899, opened 2020-02-14
Fix for missing cover on create new shelf

TBK:
Fix for #1897


ssddanbrown:
Thanks for finding and providing the fix @TBK.

Now merged in master with tests to cover, to be released in the next bugfix release.

Issue #1570, opened 2019-08-01
Call to a member function getClientOriginalExtension() on null


TBK:
**Describe the bug**
After one of the latest updates (not sure which), I can no longer create new Books or Shelves.

After enabling debugging (debugbar) I got the following:
```
/app/app/Providers/AppServiceProvider.php

    use Illuminate\Database\Eloquent\Relations\Relation;
    use Illuminate\Http\UploadedFile;
    use Illuminate\Support\Facades\View;
    use Illuminate\Support\ServiceProvider;
    use Schema;
    use Validator;
     
    class AppServiceProvider extends ServiceProvider
    {
        /**
         * Bootstrap any application services.
         *
         * @return void
         */
        public function boot()
        {
            // Custom validation methods
            Validator::extend('image_extension', function ($attribute, $value, $parameters, $validator) {
                $validImageExtensions = ['png', 'jpg', 'jpeg', 'bmp', 'gif', 'tiff', 'webp'];
                return in_array(strtolower($value->getClientOriginalExtension()), $validImageExtensions);
            });
     
            Validator::extend('no_double_extension', function ($attribute, $value, $parameters, $validator) {
                $uploadName = $value->getClientOriginalName();
                return substr_count($uploadName, '.') < 2;
            });
     
            // Custom blade view directives
            Blade::directive('icon', function ($expression) {
                return "<?php echo icon($expression); ?>";
            });
     
            // Allow longer string lengths after upgrade to utf8mb4
            Schema::defaultStringLength(191);
     
            // Set morph-map due to namespace changes
            Relation::morphMap([
                'BookStack\\Bookshelf' => Bookshelf::class,
                'BookStack\\Book' => Book::class,
                'BookStack\\Chapter' => Chapter::class,

Arguments

    "Call to a member function getClientOriginalExtension() on null"
```

When feeding the form a Cover Image I get redirected to the frontpage and get the following debugbar exception:

```
The given data was invalid./app/vendor/laravel/framework/src/Illuminate/Validation/Validator.php#306
```

**Steps To Reproduce**
1. Setup docker env + BookStack.
2. Click "Create New Shelf" or  "Create New Book"
3. Fill in info and leave out adding a Cover Image

**Expected behavior**
To be able to create new Books/Shelves.

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): 0.26.3
 - PHP Version: 7.3.7
[...]

PR #1900, opened 2020-02-15
Validation fixes

TBK:
Fix to #1570

### DIFF
--- a/app/Http/Controllers/BookController.php
+++ b/app/Http/Controllers/BookController.php
@@ -86,7 +86,7 @@ class BookController extends Controller
         $this->validate($request, [
             'name' => 'required|string|max:255',
             'description' => 'string|max:1000',
-            'image' => $this->getImageValidationRules(),
+            'image' => 'nullable|' . $this->getImageValidationRules(),
         ]);
 
         $bookshelf = null;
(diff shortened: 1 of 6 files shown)


ssddanbrown:
Thanks for offering these changes @TBK.

I like the standardising on a single image rules validation message, I can only think it's a mistake that I've left it in two places. We should probable remove the method from the ImageRepo completely to stop future duplication.

In regards to the other changes, Have you figured out exactly what these fix? I'd rather not touch the rules unless I can understand what they fix, even if only to be able to write tests for to prevent happening again.

TBK:
The "$request->filled()" checks if the http request contains the value and that is not null/empty.

nullable allows the field to be empty on validation.

With my setup without the changes I can not change the app logo, profile picture or create a shelf/book without an image.

TBK:
I have changed `$request->filled()`  to  `$request->hasFile()` to check if the http request contains a file.

nullable allows the field to be empty on validation.

With my setup without the changes I can not change the app logo, profile picture or create a shelf/book without an image.

I have also removed the getImageValidationRules from ImageRepo.

ssddanbrown:
Thanks again @TBK for these updates.

Have given them a test on my environment and all seems good. I do try to cover any back-end issues with automated tests but, since these issues only occurred in a specific environment, I have been unable to do that here so BookStack on your environment may be a little less stable than others as changes are made but I'm happy to support such updates as this where needed.

ssddanbrown:
Forgot to say, These have been merged into master to be in the next patch release.

Issue #1911, opened 2020-02-19
New Line in List After Sub-List Creates New Paragraph Block


techmattr:
Create List

1. Thing 1
2. Thing 2
    1. Sub-Thing 1
    2. Sub-Thing 2

3. Thing 3

While editing a list or ordered list a new line should remain in the same block unless you Enter twice. There is no way to break out of the sub-list and maintain the paragraph block for the next item in the main list.

techmattr:
Apparently GitHub is even worse at this. Here is screenshot example from BookStack: https://i.imgur.com/R9t39Yy.png

MikeyMJCO:
Just to be clear the input looks like this:

``` md
1. Thing 1
2. Thing 2
  1. Sub-Thing 1
  2. Sub-Thing 2
3. Thing 3
```

MikeyMJCO:
Aha - think I've found the culprit.

The HTML markup for the above is entirely correct.

``` html
<ol>
<li>Thing 1</li>
<li>Thing 2
<ol>
<li>Sub-Thing 1</li>
<li>Sub-Thing 2</li>
</ol>
</li>
<li>Thing 3</li>
</ol>
```

The issue is the CSS which has the following:
``` css
p, ul, ol, pre, table, blockquote {
    margin-top: 0.3em;
    margin-bottom: 1.375em;
}
```
So at the end of the nested `ol` (the sub-list) the CSS is adding enough padding so as to appear to have a newline. The markdown is rendering to the correct HTML.

MikeyMJCO:
What's needed here is a rule to stop nested lists getting the padding. e.g:
``` css
li > ul {
    margin-block-start: 0px;
    margin-block-end: 0px;
}
```
plus
```
li > ol {
    margin-block-start: 0px;
    margin-block-end: 0px;
}
```
this gives us this output:
![image](https://user-images.githubusercontent.com/2288257/74938665-56ad4180-53e6-11ea-8a1f-609a9df58684.png)

MikeyMJCO:
**Side-note:** GitHub is just being stupid at this - they are wrapping the text for each `<li>` element in a `<p>` tag providing the paragraph padding.

techmattr:
That does seem to mostly fix it. There are still a few extra vertical pixels between **sub-thing b** and **thing c** but it doesn't look completely off anymore.

[...]

PR #1913, opened 2020-02-20
Resolve issue #1911 (Additional margin/padding present in nested lists)

MikeyMJCO:
Optionally we could consider removing the rule at line 274. This doesn't handle multiple layers of nested lists and would be better covered by the generic rule which checks for an `ol` or `ul` nested inside a `li` which is how MarkDown renders nested lists as HTML.

**Issue Link:** #1911

### DIFF
--- a/resources/sass/_text.scss
+++ b/resources/sass/_text.scss
@@ -295,6 +295,13 @@ li.checkbox-item, li.task-list-item {
   }
 }
 
+li > ol, li > ul {
+  margin-block-end: 0px;
+  margin-block-start: 0px;
+  padding-block-end: 0px;
+  padding-block-start: 0px;
+}
+
 /*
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for this @MikeyMJCO, Have merged to be part of the next patch release.

> Optionally we could consider removing the rule at line 274. This doesn't handle multiple layers of nested lists

Yeah, I think it's redundant now with your update but would it not have handled multiple layers of nesting? Should build to `ul ul {` which should then match any `ul` in a `ul`. Obviously did not have an equivalent for `ol` and cross-type equivalents which caused the issues here though.

MikeyMJCO:
Yes - you're right - it would have caught nested unordered lists :-)

PR #1929, opened 2020-03-01
[WIP] Email Domain to Role

MikeyMJCO:
We implemented this over this weekend and decided to open a "Work In Progress" PR to see what/if any feedback there is. 

This could do with being a bit more elegant - unfortunately we don't seem to have access to all the roles on the system from the `model/auth/user.php` level so resorted to passing it to the function invocation. It doesn't feel like the best way to do this!

On the plus side - we're using this in production - running BookStack as a documentation system for our clients who use AzureAD auth in `multi-tenant` mode to login.

(**Tip:** If you omit the `AZURE_TENANT` from the Azure AD config in `.env` and configure the app appropriately in Azure you'll get a multi-tenant Azure AD setup which allows login from any Azure Tenant.)


MikeyMJCO:
**To do:**

* Run on login or create a scheduled task/artisan command to re-run the role matching - I'd prefer the console command!
* Abstract the functionality into a single place (new file/class?) and pass in the user object/roles as required/needed.
* Add test coverage.

ssddanbrown:
Hi @MikeyMJCO,

As I think I may have mentioned (On discord maybe?) a while back, I don't really want to extend auth functionality much further with custom requirements since things become hard to maintain in this area and can seemingly grow infinitely due to varied setups. 

For v21.04, I've started building the "logical theme system" ([Detailed here](https://github.com/BookStackApp/BookStack/blob/release/dev/docs/logical-theme-system.md)) which effectively allows some level of customization on certain events such as login/register and run custom logic such as this. It'll be hard to hook and implement something exactly the same as in the PR, but you could hard-code the domain mapping or get inventive and pull from another source (Could get really hacky and store the mapping as tags on a page, or in page content itself). The system is in early stages, so still a bit subject to change, but I'm welcome to feedback on it.

Since this core feature of this PR is not something I'd look to add to core, due to the above thoughts, I'm going to close this off.

PR #1942, opened 2020-03-07
Updated login and registration header actions to be consistent with other header links

Statium:
Removed extra spaces displayed in the header of the login and registration link.


ssddanbrown:
Thank you @Statium for updating these links to be consistent. Have merged for the next patch release.

Issue #854, opened 2018-05-22
Footer Links


ApertureDevelopment:
### For Feature Requests

Desired Feature: Links to articles in Page footer

Why should you add it?

Due to the EU and US privacy law, it will be mandatory for most people to add a Privacy Policy to their BookStack installation. And according to the EU law, the Privacy Policy must be available on every accessible page on your website, so the best way here is to stay with the tradition and allow Footer Links.

And for German users there is a so called Legal Info mandatory, which includes full address details of the website owner, mostly it must include:
- Full Name
- Full Address
- E-Mail
- Phone Number
- UStID

Best way to add it would be to be able to select from existing articles, pages to put in footer links.

iBooster:
This would be awesome.

qianmengnet:
Yes, I also recommend adding a footer.
1, increasing the space at the bottom editor for easy mobile phone add content.
2, increase privacy, copyright
3, increase the statistical information and other scripts, the content increased in the footer, it will not affect the page loading speed.
4, advertising space

Webmasters can add code to APP SETTINGS

qianmengnet:
I tried to add footer, but on the edit page, footer blocked the editor. How can I set css?

![11](https://user-images.githubusercontent.com/12229972/49104520-234e3680-f2b9-11e8-88e7-e5fdf7417bdb.jpg)

![22](https://user-images.githubusercontent.com/12229972/49104536-2cd79e80-f2b9-11e8-916f-7f1c8118badb.jpg)

aidanfindlater:
It looks like this isn't a priority at the moment, but has anyone had luck making their own modifications to add a footer? I'm unfamiliar with Laravel and its templating system, and pretty far from webdev in general these days, and would be grateful for any help.

james-geiger:
@aidanfindlater @ApertureDevelopment 

I'm not as familiar with European regulations but are these mandated to be located on a certain part of the page, or could they be located anywhere so long as they are accessible?

What counts as an accessible page?  Just public-facing pages?  Or also for the admin backend?

ApertureDevelopment:
They have to be easily accessible from every page. That means the link has to be always visible, cannot be hidden and should not require a search from a search functionality to find. Thats why a footer would be perfect for that

james-geiger:
For those interested (@ApertureDevelopment @iBooster @qianmengnet @aidanfindlater), you can see how this was implemented in the referenced PR.

[...]

PR #1973, opened 2020-03-19
Create footer for regulatory links

james-geiger:
Per #854 and feedback from @ApertureDevelopment, provides a mechanism to add links to a privacy policy and terms of service from the settings page.

Currently would not display on the page editor pages, but am open to and welcome feedback.  It just didn't seem to me to fit the flow very well with the full-page editor, but this could also be added as a setting as well.

@ssddanbrown, not sure what your thoughts are on limits to the number of "app settings" but this seemed like it would be a good fit for that section.  Also purposely limited this to two different kinds of links.  Could likely be abstracted into a feature where you could add any number of "custom footer links" each with a label and href, but this seemed like a good place to start.


resources/views/common/footer.blade.php
```diff
@@ -0,0 +1,10 @@
+@if(setting('app-privacy-policy') | setting('app-terms-of-service'))
```
Findus23, line 1:
```suggestion
@if(setting('app-privacy-policy') or setting('app-terms-of-service'))
```
I am not too familiar with blade templates, but without this change, I only get a lot of errors (I only entered a privacy policy link, but not a terms of service)

james-geiger: What version of PHP are you running, out of curiosity?

Findus23: PHP 7.3

Findus23: Thinking about it, I guess || would also be possible, but doesn't | do a bitwise or?

ssddanbrown:
Thank you for this PR @james-geiger and sorry for my late response.

This all looked great, but I knew that people would probably want to change the wording and links so I built upon your work and made further changes in 44c41e9e4dbe31f071b8c87ffae77ad6737c2cbc to make this a configurable, order-able list. 

PR #1984, opened 2020-03-28
Fix for #1377 : wrong URL after login when BookStack in run behind a reverse proxy

MatthieuParis:
Setting the right url.intended session variable when user is not logged yet.

Hope that helps!


ssddanbrown:
Thank you very much for offering this.
As said in https://github.com/BookStackApp/BookStack/issues/1377#issuecomment-619604431, I'm not going to merge this patch in as I'd really like to sort the URL system at a lower-level to account for other scenarios within the system.

PR #1998, opened 2020-04-03
Add support Windows Authentication via SAML

ivir:
With default SAML configuration wasn't possible use Windows Authentication via AD FS and everytime is offered only login via AD FS form. That is due missing 'requestedAuthnContext' set to false  when 
 IDP choose available method. Instead of that is send request with "'urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport" which on AD FS force to use form.
I just use setting from onelogin/php-saml/advanced_settings and provide configuration to .ENV file

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -222,6 +222,12 @@ SAML2_ONELOGIN_OVERRIDES=null
 SAML2_DUMP_USER_DETAILS=false
 SAML2_AUTOLOAD_METADATA=false
 
+# SAML Authentication context.
+# Set to false and no AuthContext will be sent in the AuthNRequest,
+# Set true and you will get an AuthContext 'exact' 'urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport'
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for this @ivir and sorry for my very very late response, Detailed SAML/Auth issues can be a pain to understand and test.

This is now merged, to be in the next patch release, after some tweaks in 9cf4191079a5d08106a6be1344fe9a822742afc4.
I changed the default from `false` to `true` since that is what I found the existing library default to be after digging into things. Also added some testing to cover this and some `.env` option parsing to allow the muli-value array input that the SAML library accepts.

I could not get any error or unexpected scenario to occur on my instance of Azure AD but I think it may depend on the exact AD setup and user authentication mechanisms use.

---

For my own reference:

- [Microsoft reference of AuthNContext values](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/troubleshooting/ad-fs-tshoot-azure#you-are-receiving-an-unknown-auth-method-error).

Issue #1640, opened 2019-09-05
Would like to Request CAS Authentication


jblankenship30:
Would like to request CAS as an SSO Authentication option for Bookstack.
Higher Education primarily uses CAS or Shibboleth for SSO Authentication.

Your product is an awesome tool!

ssddanbrown:
Thanks for your request @jblankenship30, and apologies for my late response.

To be totally open, I would really prefer not to extend to extra auth protocols unless found critical by a large group of users, which CAS really is not since this is the only request so far.

Since you opened this issue we have added SAML2 in a recent release:
* [Blog Post](https://www.bookstackapp.com/blog/beta-release-v0-28-0/#saml2-authentication)
* [Docs Page](https://www.bookstackapp.com/docs/admin/saml2-auth/)

I see a lot of CAS-Support auth systems support SAML2, as does modern versions of Shibboleth. Does that suit your need?

jblankenship30:
Yes, now that bookstack supports SAML2, my needs are met.
Thanks for a great product!


[https://banapp.tulsacc.edu/i/tccemailsig2.jpg]<https://www.tulsacc.edu/>
Jason Blankenship
Lead Oracle Database Administrator, Tulsa Community College
{o} 918-595-7052 | {c} 918-720-2253 | {w} www.tulsacc.edu
{a} 909 South Boston Ave | Tulsa, OK 74119

ssddanbrown:
@jblankenship30 Great & Thanks! I'll therefore close this off since SAML2 covers your requirement.

Note: I Updated your comment and deleted the original revision, to prevent your email address being visible on a public web-page. Looks like the GitHub email-reply system (And included office365 safelinks) can leak a person's email address pretty easily.

jblankenship30:
Thanks!

PR #2001, opened 2020-04-03
[WIP] Add CAS support

zyuiop:
This PR adds basic CAS support to authenticate users in BookStack.

This small hack is already in production for my own needs, but I felt like it would be a nice addition to upstream.

I've never worked with Laravel before, so all suggestions to improve this are welcome!

### DIFF
--- a/app/Auth/Access/CasService.php
+++ b/app/Auth/Access/CasService.php
@@ -1,0 +1,80 @@
+<?php namespace BookStack\Auth\Access;
+
+use BookStack\Auth\User;
+use BookStack\Exceptions\SamlException;
+use BookStack\Exceptions\SocialSignInAccountNotUsed;
+use Illuminate\Support\Str;
+
+class CasService extends ExternalAuthService
+{
+    protected $config;
+    protected $registrationService;
+    protected $user;
+
+    /**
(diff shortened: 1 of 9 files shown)


ssddanbrown:
Hi @zyuiop,
Thank you for offering this pull request.

Did you come across issue #1640 at all? I stand by my comment in that issue, in that I'd look to **_not_** include CAS authentication unless found critical by a large portion of BookStack users, especially as there seems to be some level of overlap with SAML. 

I am genuinely thankful to you for offering this upstream, but it's just the documentation, dev environment, testing, ongoing maintenance and support that I'm mostly concerned with.

zyuiop:
Hi,

Sorry, I did see this issue before I started hacking but I completely
forgot about it after that.

I understand that most users don't need that protocol, but the reason why I
did that is that it's a way simpler protocol to implement.

Anyway, I'd perfectly understand if you close this PR.

Cheers!
Louis

ssddanbrown:
Thank you for your understanding. Good to know it's a simpler protocol though. If there's any tweaks we can do to make it easier to maintain your patches, for your own use, feel free to open an issue here to let me know.

Otherwise, I'll close this off.

PR #2003, opened 2020-04-05
Updated styles to use logical properties/values, for improved RTL support

ssddanbrown:
- Intended to improve RTL support in the interface.
- Also adds hebrew to language dropdown since that was missing.

Related to #1794


ssddanbrown:
Tested across chome, firefox and webkit. No major glaring issues there so will merge this in.

Issue #1234, opened 2019-01-23
Dark theme


gerroon:
Hi

Can you please add a default dark theme to BS. I find the default white theme to be super overwhelming, so I actually avoid using BS myself. If there was a dark theme installed I would definetely go for it.

This brings me a question about Docker install. How can I add a dark theme to the docker install? Like this one https://github.com/domainzero/bookstack-dark-theme

tomaskir:
Please make a dark theme happen! :)

antimatteraccord:
Another vote for dark mode!!
Please make this a simple button click in the UI.  That would help so much!

kaleben0:
Not sure if anyone mentioned this, but this guy made a Dark Theme CSS for Bookstack - https://github.com/domainzero/bookstack-dark-theme

Chrismettal:
+1 for dark mode.

ssddanbrown:
A user-selectable dark mode has been implemented under #2022. This will be accessible via the top-right header dropdown menu (Under logout). This will be part of the next feature release, v0.29. 

A massive thanks to @domainzero for providing & maintaining a dark mode option for users over the last three years :1st_place_medal:

ThorsenRune:
It appears to me that this does not apply to guest visitors. How can that be enabled for the guest account?

ssddanbrown:
@ThorsenRune Being able to set the default state has been addressed already in #2081 and will be part of the next feature release: https://github.com/BookStackApp/BookStack/issues/2081#issuecomment-774786429

PR #2022, opened 2020-04-10
Addition of a user-selectable dark-mode option

ssddanbrown:
- [x] Most elements done, but still need to do editors, tables and final pass.
- [x] Toggled only by quick js check at the moment, checking via css media query. Need to make into user-preference toggle.
- [ ] Say thanks to domainzero

For #1234


ssddanbrown:
bookstack local_ (1)

over-soul:
@ssddanbrown, is there a way to let guest users select dark mode?

ssddanbrown:
Should be an option on the home screen. See above the right-most "Recent Activity" card here:
https://demo.bookstackapp.com/

whentotrade:
Would it be possible to also place this button to switch between dark/light mode on the books, shelf, chapters page sites, e.g. as another item in the "Actions" section below "Export"? This would make is possible to change theme on book/page sites also an not only on the main home screen.

ssddanbrown:
@whentotrade Dark mode does not really relate to the items that the existing actions are centered around so I'd prefer not to add it there.  

Non guest users can toggle dark mode from any page via the top-right profile menu.

whentotrade:
Ok, thanks. Then maybe just on the /books page. As there the "dark/light mode" open is missing when you click on the "Books" tab from the tab. While when you choose "books" as the standard homepage, the switch button is there in the overview. 

This is the "Actions area" on the upper left when setting "books" as the standard homepage in the settings (with switch button):
grafik

Where as this is the "Action area" when clicking the main "Books" tab on the top bar (without switch option)
grafik

Then main contant is the same - just the landing page selected as "books" in the settings shows the switch button while when clicking the "books" main tab the option is not visible.

Just a proposal and observation.

Thanks for the great work!

PR #2023, opened 2020-04-11
fix Chinese translation error

jzoy:


ssddanbrown:
Thank you for these improvements @jzoy. I have merged them in, to be part of the next release.

PR #2047, opened 2020-04-16
update

liudandandear:
同步更新


ssddanbrown:
Thanks for offering a pull request but I'm afraid I'll have to close this as it's quite a random collection of changes that I don't think are suitable for all BookStack users. For example, I don't think everyone would appreciate `baidu` tracking scripts being added to their instance :sweat_smile:.

Issue #1717, opened 2019-10-15
Export .MD file in addition to HTML and PDF


caseydierking:
**Describe the feature you'd like**
The ability to export single or bulk pages in .MD format instead of HTML or PDF

**Describe the benefits this feature would bring to BookStack users**
Let me preface this by saying I love bookstack. 
I want to be able to backup my files to a flat file structure and share them as I need to at the file level. I know we have HTML and PDF options for exporting, but I would love to be able to export the file directly to markdown. 

A lot of times, I write documentation for my company in bookstack and transfer that information over to our other static site generator for our docs. This would be a handy feature for me.

ssddanbrown:
Hi @caseydierking,
Thanks for the request. At some point of the upcoming API build we'll look to handle conversion between HTML and Markdown on the back-end (Current conversion is done client-side only), At that point it should be a quick job to enable markdown as an export option. Could technically implement it sooner but I'd prefer to build the feature for all users rather than needing to add dynamic control to export options dependant on what page editor is in use.

Somewhat related to #1273.

caseydierking:
Thanks for the response @ssddanbrown . Also thank you for your time and effort into this project. I'm impressed at how well thought out your responses are to issues and feature requests. 

Is there a current roadmap on development for the API? Anything others can do to help contribute?

Thanks,

ssddanbrown:
@caseydierking Thanks!

No official roadmap specifically for the API. For the next release we'll probably have the auth mechanism and endpoints for one focused thing (Like just books) so it can be tested and feedback can be provided before we span it out to cover most BookStack actions. 

Just a case of me finding the time to get started on it.

XVilka:
I think the second, related option should be also "Export MD with attachments", that will produce Markdown file, add the attached images/other files, then put everything in ZIP for downloading.

alderson59:
Hey, new to this project, and already loving it! How are pages and books currently stored? I can't seem to find the directory in which they are stored. Is this all done in the MYSQL database?

Would it be possible to keep pages and books stored on a directory openly as .md files, as an option? I would love to store some important info using bookstack to manage, and would especially be useful if the files are not easily accessible in the event of a crash/whatever.

Not sure if this deserves its own FR, so pinging for some thoughts first :)

OliverEvans96:
I agree with @alderson59 - I would feel much more comfortable putting important content into bookstack if I know I can easily export it / back it up in a plaintext format

jacky35:
[...]

PR #2115, opened 2020-05-13
Markdown Export

nikhiljha:
I want to be able to export a full, organized, plaintext copy of a book like @XVilka [suggests](https://github.com/BookStackApp/BookStack/issues/1717#issuecomment-543028773) in #1717.

## Status

- [x] Export single .md files
- [x] Convert WYSIWYG HTML into Markdown on export
- [x] Make the export button show in the UI
- [x] Export zip files with proper directory trees

I know this is a cold PR, but hopefully there's something usable in here. General guidance for how/where to implement these features is appreciated.

## Help Needed
- [ ] Export media in the zip file
- [ ] Since this is a heavy operation, gate .zip export behind a permission

These are things that might be good to also add... not sure where to start on them though.

## Benefits

- Adopting BookStack is easier knowing that all the data stored there is in a standard format. 
- A lot of people like git sync (hi @dkess), so this makes that possible with external software.

### DIFF
--- a/app/Entities/ExportService.php
+++ b/app/Entities/ExportService.php
@@ -8,6 +8,7 @@ use Exception;
 use SnappyPDF;
 use League\HTMLToMarkdown\HtmlConverter;
 use Throwable;
+use ZipArchive;
 
 class ExportService
 {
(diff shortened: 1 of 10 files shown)


nikhiljha:
Played around with some things but I'm not really sure how to make codeclimate happy... :(

This PR also now introduces a dependency which could be used to properly do conversions between MD <--> HTML. After testing, the only thing that breaks are the "Success, Warning, Danger" boxes, and the code blocks use nonideal (but still working) syntax.

Zenahr:
If this is tested properly and get's through it would be such a biggie for me. @nikhiljha Thank you for putting in the effort.

A Markdown Import function would also be awesome somewhere down the road.

ssddanbrown:
Hi @nikhiljha, sorry for my very late response and thank you very much for offering this.

I've merge this in and followed this up with commits 57ea2e92ec07d4682466857d27bd706a30e39652 and 992f03a3c01759653594fc383917b9aa3affb5a2. 

Within those commits I've addressed some of the points mentioned on your previous comment, to get the callout boxes working (Or at least left as HTML) along with making some content tweaks so that code blocks function as expected.

I've also removed the ZIP solution for now to get the core functionality moving and merged. In regards to the ZIP component, it's a good idea but I think will need to be thought about more universally, as you'd maybe then expect ZIP options for the other formats. Maybe it'd be an option/parameter on the existing endpoints, rather than whole new endpoints. 

The core markdown export functionality will be part of the next feature release.

PR #2120, opened 2020-05-14
Add id tag to the registration form

mtannertdev:
Add an id tag to the registration form. An id tag already exists on the login form and this will allow both the login form and registration form elements be alterable from the "Custom HTML Head Content" option in Settings. We hide both of these forms because all of our users use federated authentication to log in. We do not want users creating accounts in the app via the registration form.

### DIFF
--- a/resources/views/auth/register.blade.php
+++ b/resources/views/auth/register.blade.php
@@ -8,7 +8,7 @@
         <div class="card content-wrap auto-height">
             <h1 class="list-heading">{{ Str::title(trans('auth.sign_up')) }}</h1>
 
-            <form action="{{ url("/register") }}" method="POST" class="mt-l stretch-inputs">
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @mtannertdev,
From the sounds of your description you may be better off with disabling registration within the system and instead using the `{SERVICE}_AUTO_REGISTER=true` option as detailed here:
https://www.bookstackapp.com/docs/admin/third-party-auth/

Alternatively, You don't need an ID to target this form. You could always do something like

```css
form[action$="/register"] {
 /** Your css here */
}
```

I could still merge this if you really wanted if the above do not suffice, but ID's on certain elements are not something I'll ever ensure are stable across releases.

ssddanbrown:
Will therefore close this.

PR #2133, opened 2020-05-22
Add search function to database layer

jeff-ong:
Add MySQL stored procedure to allow us to call the search function in the database layer. It lets us access the search feature when using the relational database with different frameworks.

Steps to create this SP in database:
1. First, ensure that search_terms table is a type of either InnoDB or MyISAM
2. Create a new index on the search_terms table: `CREATE FULLTEXT INDEX search_terms_fulltext_index ON search_terms(term)`
3. Create the stored procedure by executing the code inside ./database/storedProcedure/search-books.sql

### DIFF
--- a/database/storedProcedure/search-books.sql
+++ b/database/storedProcedure/search-books.sql
@@ -1,0 +1,43 @@
+DELIMITER //
+CREATE PROCEDURE `search_book`(
+	IN v_search_query VARCHAR(100),
+    IN v_limit INT
+)
+BEGIN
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this pull request @jeff-ong.

I can see how this many be useful to others, But I wouldn't really be looking to support database-specific functionality, such as this, as part of this repo to be honest and from the look of it there's no functional purpose of including this script within the core project.

ssddanbrown:
Due to the above reasons, I'll close this off.

PR #2157, opened 2020-06-16
fix the translate error

Honvid:
fix the translate error that there have a extra letter 'n'.


ssddanbrown:
Thank you for this @Honvid, Will merge in for the next feature release.

PR #2162, opened 2020-06-19
Update search index when updating a page

kabab:
PR for updating search index when updating a page.

### DIFF
--- a/app/Entities/Repos/PageRepo.php
+++ b/app/Entities/Repos/PageRepo.php
@@ -193,6 +193,8 @@ class PageRepo
 
         $page->save();
 
+        $page->indexforSearch();
+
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @kabab,
Thanks for offering this fix.

Pages are already indexed on this line:
https://github.com/BookStackApp/BookStack/blob/master/app/Entities/Repos/BaseRepo.php#L71
And hence page content should already be indexed when using the master branch.
There was a bug, due to the ordering of update actions, that meant content was not properly indexed. This was fixed in [this commit](https://github.com/BookStackApp/BookStack/commit/00c0815808ffd98ee32aef8be56f1ba2dba85ea5)  which will be in the next release.

Issue #1820, opened 2019-12-19
NPM cannot download node-sass binding for Node 13


Grawl:
**Describe the bug**

Because of `image: node:alpine` in `docker-compose.yml`, fresh Node.js v13 is used. But there is no `node-sass` bindings published for Node 13 on Alpine.

```
> node-sass@4.12.0 install /app/node_modules/node-sass
> node scripts/install.js

Downloading binary from https://github.com/sass/node-sass/releases/download/v4.12.0/linux_musl-x64-79_binding.node
Cannot download "https://github.com/sass/node-sass/releases/download/v4.12.0/linux_musl-x64-79_binding.node": 

HTTP error 404 Not Found

```

**Steps To Reproduce**

1. Clone this repo with `master` on this commit: https://github.com/BookStackApp/BookStack/commit/b93f8a4d4680cfe34e3b64db95f0f9683152c22a (last commit when I write this issue)
2. Run `docker-compose up -d --build`
3. Open `node` container logs

**Expected behavior**

All containers is up

**Your Configuration (please complete the following information):**

 - Exact BookStack Version (Found in settings): `v0.28-dev`
 - PHP Version: `7.3`
 - Hosting Method: Docker

**Additional context**

To fix this, I change `docker-compose.yml` to specify Node container version:

```diff
  node:
-    image: node:alpine
+    image: node:12-alpine
```

ssddanbrown:
Thanks @Grawl for reporting this.

I have just fixed this in `master` by updating the npm dependencies so the project uses a node-13 compatible version of `node-sass`. Have tested by bringing up the docker-composer setup and checked it built the SASS files. 

I thought about pinning the node image version as per your fix but I think I'll leave it for the mean-time since it was quite useful in this case; Using the latest version indicated that a dependency update was probably needed.

PR #2166, opened 2020-06-26
Replace node-sass with dart-sass

timoschwarzer:
This PR replaces `node-sass` with `sass`. `sass` is the [SASS/SCSS reference implementation](https://github.com/sass/dart-sass).

### Pros:
- No native dependencies (JS only)
- Faster availability of new features

### Cons:
- A bit slower, but not very noticable.

### Why?
`node-sass` broke again for me with Node 14 with some error that it couldn't download some binaries from GitHub.

```
node_1     | Downloading binary from https://github.com/sass/node-sass/releases/download/v4.13.1/linux_musl-x64-83_binding.node
node_1     | Cannot download "https://github.com/sass/node-sass/releases/download/v4.13.1/linux_musl-x64-83_binding.node": 
node_1     | 
node_1     | HTTP error 404 Not Found
node_1     | 
node_1     | Hint: If github.com is not accessible in your location
node_1     |       try setting a proxy via HTTP_PROXY, e.g. 
node_1     | 
node_1     |       export HTTP_PROXY=http://example.com:1234
node_1     | 
node_1     | or configure npm proxy via
node_1     | 
node_1     |       npm config set proxy http://example.com:8080
node_1     | 
node_1     | > node-sass@4.13.1 postinstall /app/node_modules/node-sass
node_1     | > node scripts/build.js
```

Issues like the above or https://github.com/BookStackApp/BookStack/issues/1820 shouldn't arise anymore with `sass`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -189,12 +189,6 @@
       "integrity": "sha512-NuHqBY1PB/D8xU6s/thBgOAiAP7HOYDQ32+BFZILJ8ivkUkAHQnWfn6WhL79Owj1qmUnoN/YPhktdIoucipkAQ==",
       "dev": true
     },
-    "abbrev": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
-      "dev": true
-    },
     "acorn": {
       "version": "6.4.1",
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Thank you again @timoschwarzer!

Done some rough timings of before and after for a development build (npm run build) and found that, on my system, there was an increase from about 2.7 to 3.5 seconds. Probably never too noticeable, like you say, but still about a 30% increase. I do agree that the dropping of native-deps is worth it though. 

Nethertheless, I've played about with the build system and moved SASS builds out of webpack: https://github.com/BookStackApp/BookStack/commit/540119f133c99dcfddf37d623735e2a0f9660c14

With those changes we're down to 2.1 seconds `npm run build` time on my machine. So now we should be faster _and_ have no native dependencies :tada:

PR #2184, opened 2020-07-10
Feature - Convert Code Language to Select instead of Links + Input

b-zurg:
Description

Converted the language select in the code editor to be a dropdown instead of a link select and free-text input field.  This is a common pattern for this UI element and makes a lot of sense for a code language.

Added language support for TypeScript and JSX/TSX to allow for React and Typescript documentation.

Added a change that would ensure that the line-endings are LF (for windows developers) to not have an issue with the developer docker setup.

Added an environment variable `APP_ENV` to docker-compose as it was necessary for my development environment to run.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,9 +3,6 @@
 # Run `php artisan key:generate` to generate a valid key.
 APP_KEY=SomeRandomString
 
-# Necessary to set if running docker-compose in a dev setting.
-APP_ENV=local
-
 # Application URL
 # Remove the hash below and set a URL if using BookStack behind
 # a proxy, if using a third-party authentication option.
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,9 @@
 # Run `php artisan key:generate` to generate a valid key.
 APP_KEY=SomeRandomString
 
+# Necessary to set if running docker-compose in a dev setting.
+APP_ENV=local
+
 # Application URL
 # Remove the hash below and set a URL if using BookStack behind
 # a proxy, if using a third-party authentication option.
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
-* text=auto
+* text=auto eol=lf
 *.css linguist-vendored
 *.less linguist-vendored
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,7 @@ services:
       MAIL_DRIVER: smtp
       MAIL_HOST: mailhog
       MAIL_PORT: 1025
+      APP_ENV: local
     ports:
       - ${DEV_PORT:-8080}:80
     volumes:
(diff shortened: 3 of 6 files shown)


JtheBAB:
Would it be possible to define a default language per user? Then you don’t need to choose all the time what. With this solution it would cost me at least 2 clicks. Before just one.

b-zurg:
@JtheBAB it should be pretty easy to cache the last selected language in the browser's local storage.  I feel like that would be an acceptable solution personally as it would simply take less code and be easier to maintain.  What do you think?

JtheBAB:
Worth to try. How long would it last in the local storage?

ssddanbrown:
Thanks for offering this pull request @b-zurg.

Would you be able to explain why a select list would be better? Personally I avoid select lists where possible and I generally find myself disliking them for code selection. I don't want to make a change for the sake of convention.

This change may also cause issues for those that previously used the text input, with a value not available in the dropdown.

b-zurg:
Interesting points.  Maybe I can outline the reason why I made this PR and we can consult about what would be a good UX solution.

The main concern I have with the current code block popup is the free text field.  This is a field with a limited set of possible values.  Those values are displayed above but also somehow editable.   I think you can appreciate how this would be confusing or prone to error.  What if I want to use a language but don't type it in correctly, or how would I even know what languages are available by typing them in?  Do I go and look at the source code and then look at codemirror's site?

So having a controlled input makes sense.  The question is what form does it take.  Do we take the current approach and ensure all available options are selectable, but make the input function as a display-only field?  Do we have a searchable drop-down?

I went for the select list / drop-down as I think that it would be good to support all of the possible languages that codemirror provides, but to do that in the current form would involve quite a wall of text, sort of overloading the user with input that could be confusing.  

In my opinion the ideal would be a searchable dropdown.

Does this make sense?

ssddanbrown:
The free text input is more a parallel of how the language would be set in the Markdown editor. Sure they might not know what options are available but nor do you with the Markdown editor input, at least until you get the visual feedback of your content being highlighted. This input also allows the display of options that may have come from the markdown editor, where accepted values are more lenient with multiple options per language. 

> I went for the select list / drop-down as I think that it would be good to support all of the possible languages that codemirror provides, but to do that in the current form would involve quite a wall of text, sort of overloading the user with input that could be confusing.

I don't load in all languages on purpose at the moment, as I really wanted to keep the loaded JS lean and not have loads of redundant options to sift through, Instead seeing what languages people actually request. I'll soon be looking at splitting out the codemirror code anyway, and potentially dynamically loading the modes to be a bit more lenient there in the future. 

I get that a select option is the go-to for something like this, but I really do detest inserting a code block in teams, due to the select dropdown. I really do like the quick press options or, if not wanting to find an option, the quick-type option in BookStack currently, albeit a little messy in UI.

ssddanbrown:
This does not seem to be going anywhere so I'll close this off. Thanks still for your enthusiasm and efforts.

PR #2214, opened 2020-08-06
fix test errors of comment 87a5340

tomleesm:


ssddanbrown:
Thanks for offering this pull request @tomleesm to make the tests pass.

That said, I think the tests are correctly reporting of an error in the flow, and the code needs to be updated to pass the tests instead of the tests being updated to suit the code.

ssddanbrown:
Will therefore close this off

PR #2227, opened 2020-08-14
Render included content when exporting books

### LINKED ISSUE #2228: Book export doesn't print included content
**Describe the bug**
When I export a page or a chapter, content that I include with `{{@123}}` tags is rendered in the output.  However, the literal tag is printed in the output when I export the book.

**Steps To Reproduce**
1. Create a page that includes content from another page (`{{@123}}`).
1. Export the book containing the page from step 1.
1. The exported file prints the literal include tag.

**Expected behavior**
The page and chapter exports both print the rendered, included content instead of the literal include tag.  I expected the book export to do the same.

**Screenshots**
Rendered page with an include tag:
image

Page source showing include tag:
image

Exported page:
image

Exported book:
image

 - Exact BookStack Version (Found in settings): 0.29.3 (screenshots are from the demo site)
 - PHP Version: unknown
 - Hosting Method (Nginx/Apache/Docker): unknown

mr-vinn:
Currently, if I create a page with included content (`{{@123}}`) then I see the rendered content when I export the page or its containing chapter.  But the book export prints the literal string `{{@123}}` in the output instead of the included content.

This change results in the included content being rendered no matter which object (book, chapter, page) is exported.

Fixes #2228

### DIFF
--- a/app/Entities/Managers/BookContents.php
+++ b/app/Entities/Managers/BookContents.php
@@ -41,7 +41,6 @@ class BookContents
 
     /**
      * Get the contents as a sorted collection tree.
-     * TODO - Support $renderPages option
      */
     public function getTree(bool $showDrafts = false, bool $renderPages = false): Collection
     {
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thank you @mr-vinn for this. Has been merged and testing has been added to cover this case. Will be part of the next patch release.

PR #2233, opened 2020-08-24
Updated callout link formatting

### LINKED ISSUE #303: Callout formatting
Desired Feature:

Improved formatting for callouts, including:

* Link colouring to match the style:

screen shot 2017-02-07 at 01 52 18

* Paragraphs, and other mark down to be suppported.

alexmannuk:
Updated callout links to use font colouring based on type, with bold text to denote link, instead of using the theme link colour. Closes #303.

SCSS has been updated, will need CSS (npm run dev) before merge.

### DIFF
--- a/resources/sass/_blocks.scss
+++ b/resources/sass/_blocks.scss
@@ -27,6 +27,7 @@
     border-left-color: $positive;
     @include lightDark(background-color, lighten($positive, 68%), darken($positive, 22%));
     @include lightDark(color, darken($positive, 16%), lighten($positive, 5%));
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for this @alexmannuk, Have merged in for v0.30.1.

Your pull request did make the text slightly green for all four different call-out colors, so I applied commit de2756dd95fb9378b220b284f4bbe7ebcfe80b20 afterwards to instead inherit the correct color and use underline instead of bold since that's generally more common for links. A user could still apply bold manually if they wished.

PR #2238, opened 2020-08-31
Complete and fix Czech l10n

jakubboucek:
## Notes from each commits:
**Fix Czech translations (email -> e-mail)**
In Czech language "email" does not means "email" but "enamel paint", correct is "e-mail".
See in Wikipedia:
- https://cs.wikipedia.org/wiki/E-mail
- https://cs.wikipedia.org/wiki/Email_(barva)

**Czech translations: Add new translations to cs, improve existing**

**Czech translations: Fix broken labels**

**Fix English translations**
- Fix obvious bug
- Reunite capitalisation


ssddanbrown:
Thank you for providing these updates @jakubboucek. These have been merged to be part of the next update.

Note, there were some conflicts with our official translations process via https://crowdin.com/project/bookstack so not all your changes would have made it to the current language files.

PR #2257, opened 2020-09-17
Decode HTML entities in excerpts

### LINKED ISSUE #2114: Text not rendering properly in the "Recently updated card"
**Describe the bug**
I have a page in one of my books that has the following text: `Authy -> AndOTP`. For some reason, when I save the page and go back to my Bookstack home page, the page shows up in the "Recently Updated Pages" card, but the text in the recently updated section doesn't render properly, it shows the escaped character for the greater than sign: `Authy -&gt; AndOTP`. Looks like when the text is retrieved from the DB, the escape characters aren't accounted for properly.

**Steps To Reproduce**
1. Create a page with a less than or a greater than sign near the beginning of the page.
2. Go back to Bookstack homepage.
3. Look at the "Recently Updated Pages" card and see that the less than and/or greater than sign isn't rendered properly.

**Expected behavior**
The text in the "Recently Updated Pages" card should properly render the greater than sign and read out `Authy -> AndOTP`.

**Screenshots**
My page looks like this:
image

The "Recently Updated Pages" card looks like this:
image

 - Exact BookStack Version (Found in settings): v0.29.0
 - PHP Version:  7.3.16
 - Hosting Method (Nginx/Apache/Docker): Docker

mr-vinn:
Decode HTML entities in the text content before returning an excerpt.  Fixes #2114.

### DIFF
--- a/app/Entities/Managers/PageContent.php
+++ b/app/Entities/Managers/PageContent.php
@@ -25,7 +25,7 @@ class PageContent
     public function setNewHTML(string $html)
     {
         $this->page->html = $this->formatHtml($html);
-        $this->page->text = $this->toPlainText();
+        $this->page->text = html_entity_decode($this->toPlainText());
     }
 
     /**


mr-vinn:
I think the other `getExcerpt()` functions need some attention too; I've marked this PR as a draft until I can review the other classes.

ssddanbrown:
Thanks for starting a pull request. I think the HTML entities should probably be decoded before the text is stored in the DB, that way you won't have to alter every point that the text is used.

mr-vinn:
>  I think the HTML entities should probably be decoded before the text is stored in the DB

I wasn't sure if anything else was relying on the stored content being encoded, but if you think that's safe then I'm all for it.  I'll replace this commit with a change that stores decoded text in the database.

ssddanbrown:
Hi @mr-vinn,
I agree, That current search behaviour is just due to the existing issue and therefore not correct in itself.

I'd say don't worry about decoding existing content in a migration. This issue is mainly cosmetic and I'd prefer to not have a fairly heavy operation in the migrations to just solve a cosmetic issue. I always try to minimise risk in the migrations where possible.

If a way to update all existing plain text content is really warranted I'd prefer a command to be added which a user could optionally run but I don't think it is.

mr-vinn:
> I'd say don't worry about decoding existing content in a migration.

That works for me; in that case this PR is ready to merge.

ssddanbrown:
Thanks again @mr-vinn, Testing added to prevent regression & merged to be in the next release.

PR #2268, opened 2020-09-22
Fix the `AddActivityIndexes` migration's `down()` method

gertjankrol:
This PR fixes `artisan migration:rollback`, which was broken due to incorrect index names in the `2020_09_19_094251_add_activity_indexes` migration:

```
SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP INDEX `key`; check that it exists (SQL: alter table `activities` drop index `key`)
```


gertjankrol:
Proposed a workflow to automatically test migrations in https://github.com/BookStackApp/BookStack/pull/2270

ssddanbrown:
Thanks for fixing this @gertjankrol, I really need to remember that `index` and `dropIndex` work in different ways. Merging in now for next patch release.

PR #2270, opened 2020-09-22
Add `test-migrations` workflow

gertjankrol:
This PR adds a workflow to automatically test migrations


gertjankrol:
The currently failing jobs are fixable by applying the fix proposed in https://github.com/BookStackApp/BookStack/pull/2268

ssddanbrown:
Thanks for this @gertjankrol, This is a very good idea!

PR #2272, opened 2020-09-22
Fixed canonical redirects on non-root url app instances

jakubboucek:
If BookStack instance is deployed to any non-root path, e.g. `http://example.com/wiki/`,
requests for `http://example.com/wiki/shelves/`
was redirected to `http://example.com/shelves`
instead of `http://example.com/wiki/shelves`.

`%{REQUEST_URI}` contains whole URL path, script path included.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -11,9 +11,10 @@
 
     # Redirect Trailing Slashes If Not A Folder...
     RewriteCond %{REQUEST_FILENAME} !-d
-    RewriteRule ^(.*)/$ /$1 [L,R=301]
+    RewriteCond %{REQUEST_URI} (.+)/$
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @jakubboucek,
Thanks for offering this.

Would it be possible to see if your setup works with this `.htaccess`?:
https://github.com/laravel/laravel/blob/master/public/.htaccess

I see that's it includes a similar change you yours, but it would be ideal to align with that linked one instead if it works so that BookStack is aligned with it's framework.

jakubboucek:
@ssddanbrown You have right, it works and your note does sense. 

Changed & rebased to current `master`.

ssddanbrown:
Thanks @jakubboucek, Will merge now to be part of the next patch release.

PR #2274, opened 2020-09-23
Fixed "Ubunto Mono" $mono type misspelling

abulgatz:
`$mono` CSS variable had misspelling of `Ubuntu Mono`.


ssddanbrown:
Thanks @abulgatz, Good find, Has gone unnoticed for at least 3 years :sweat_smile: 

Will merge for next patch release.

PR #2288, opened 2020-09-29
Added console command to sync with ldap server

injektion:
Current behaviour:

When LDAP enabled, if a user tries to log in and isn't found, credentials are checked and user is imported.

Issue:

We require users to be populated from the LDAP server before they visit the site the first time. Also want option to prevent users from being auto registered if we'd prefer to use the LdapSync.

As example, if you're using any sort of SSO or Server/Header authentication, the user won't exist for an auto login check, and will have to click Login before being logged into their account.  Very confusing for users who expect to just be auto logged into their own account.

Fix:

Use this console command to trigger a sync with an LDAP filter (LDAP_SYNC_USER_FILTER) that will pull in users and sync groups prior to the login request.  

Changes:

Added and stuck most of the logic in  /app/Console/Commands/SyncLdap.php command

Added two config variables in config/services.php -> LDAP which default to current behaviours of the app
        'sync_user_filter' => env('LDAP_SYNC_USER_FILTER', false),
        'sync_user_recursive_groups' => env('LDAP_SYNC_USER_RECURSIVE_GROUPS', true),

Added public function to app/Auth/Access/LdapService.php to getAllUsers() by filter [defaults to sync_user_filter]

Added an option to disable auto registering members when using LDAP (config/auth/auto-register = TRUE) which stops the createNewFromLdapAndCreds call in /app/Auth/Access/Guards/LdapSessionGuard.php

### DIFF
--- a/app/Auth/Access/Guards/LdapSessionGuard.php
+++ b/app/Auth/Access/Guards/LdapSessionGuard.php
@@ -79,7 +79,7 @@ class LdapSessionGuard extends ExternalBaseSessionGuard
             return false;
         }
 
-        if (is_null($user)) {
+        if ((is_null($user) && config("auth.auth_register"))) {
             try {
                 $user = $this->createNewFromLdapAndCreds($userDetails, $credentials);
             } catch (UserRegistrationException $exception) {
--- a/app/Auth/Access/Guards/LdapSessionGuard.php
+++ b/app/Auth/Access/Guards/LdapSessionGuard.php
@@ -79,9 +79,13 @@ class LdapSessionGuard extends ExternalBaseSessionGuard
             return false;
         }
 
-        if ((is_null($user) && config("auth.auth_register"))) {
+        if (is_null($user)) {
             try {
-                $user = $this->createNewFromLdapAndCreds($userDetails, $credentials);
+                if (!config("auth.auth_register")) {
+                    $user = $this->createNewFromLdapAndCreds($userDetails, $credentials);
+                } else {
+                    throw new LoginAttemptException("LDAP on and Auth_Register disabled");
+                }
             } catch (UserRegistrationException $exception) {
                 throw new LoginAttemptException($exception->message);
             }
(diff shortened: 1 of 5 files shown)


tiredofit:
This is great stuff! I just tried this out on my installation and this works out well. It's less taxing to setup the accounts ahead of time rather than having to scramble when someone newly logs in to give them the appropriate permissions to groups that are not passed forth in LDAP. 

The Auth register command is a nice touch! I can see that being useful for those using SAML and OAuth services who want to restrict users as well and don't have the capability from the IDP.

I do hope this is merged into master.

ssddanbrown:
Thank you very much for offering this @injektion and sorry again for my very late reply.

To be honest, I'm hesitant in regards to merging this. I can understand that some this will be very useful to some but:

- I really don't want to increase the surface area of auth functionality unless there's significant desire or impact to including such functionality in the core.
- This appears to have been little desire for this feature, and it's audience would be fairly low (Subset of an already smaller subset of people that use LDAP auth)
- I'm concerned this is a feature where configurability and customization of logic would be desired. This is reflected by already having the option to filter out specific email addresses. 

I'm wondering though if instead I could ensure that the right internal elements are exposed so this command could be added via the [logical theme system](https://github.com/BookStackApp/BookStack/blob/master/dev/docs/logical-theme-system.md). That way such a command could be added to an instance without altering core files, without me needing to maintain it in core (And the review/testing work needed to include it initially) and then people could customize the logic to their liking. 

I'm thinking we'd need to provide:

- A way to register custom commands (Think might already possible but just need to be sure and provide an example).
- A way to externally make a LDAP connection via the LDAP service (Any maybe we also return the underlying LDAP class instance and config options for convenience).

ssddanbrown:
Following on upon the above, within v21.11.1 I've now added support for registering custom commands via the logical theme system: https://github.com/BookStackApp/BookStack/blob/master/dev/docs/logical-theme-system.md#custom-commands

Therefore I'm going to close this off. Feel free to raise issues if there are problems with this custom command registration is needed, or if further guidance on the system is needed.

Issue #2223, opened 2020-08-13
Authenticate via header, e.g. Auth Proxy


ryanc-me:
Support for authentication via header would be a great low-cost addition to Bookstack. With this authentication mode, Bookstack would check for the request header `X-Webauth-User`, and attempt to log-in as that user automatically with no password/token. This mode leverages an external service, the *auth proxy*, to authenticate users and add the header. Sessions are stored externally also; the user remains logged-in while the header is present.

[Louketo](https://github.com/louketo/louketo-proxy) is a good example auth proxy that supports OIDC providers, like Keycloak, Dex, etc.

Grafana [has an authentication module](https://grafana.com/docs/grafana/latest/auth/auth-proxy/) that works in this way.

To be clear, I am planning to make a pull request to add these features. It would be great to get some feedback on the plan first though!

How it works
---

The login flow looks like:
 1. User visits `bookstack.example.com`
 2. The request comes through to an auth proxy, *not* Bookstack. For example:
     - Nginx configured with Basic Auth, or
     - Louketo, connected to an external OIDC IdP like Keycloak.
 3. The auth proxy authenticates the user (via a login page, an SSO token, etc)
 4. The auth proxy adds/populates the `X-Webauth-User` header, and proxies the request to Bookstack.
 5. Bookstack sees the header, trusts the auth proxy, and logs the user in automatically.
 6. User is redirected to the application, skipping the login page.

A variety of auth proxies can be used to authenticate and set the header, for example: [Basic Auth](https://grafana.com/docs/grafana/latest/auth/auth-proxy/#making-apaches-auth-work-together-with-grafanas-authproxy), [vouch proxy](https://github.com/vouch/vouch-proxy), [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy).

Benefits
---
 - Allows a variety of authn services to be used; e.g. Keycloak supports almost any OIDC provider.
 - Allows sessions to be managed externally (useful for strict security policies).
 - Allows seamless SSO; users only presented with login page once, then have immediate access to other applications.
 - Highly secure; leveraging an auth proxy project whose sole focus is on security.
 - Easy to implement; minimal code to get it working.
 - The Keycloak/Louketo setup is popular in Kubernetes, and would allow Bookstack to fit into existing setups very nicely.
 - Permissions still handled within Bookstack; the auth proxy *only* handles authentication.

Considerations
---

[...]

Issue #2180, opened 2020-07-09
$_SERVER_REMOTE_USER authentication


tiredofit:
**Describe the feature you'd like**
Adding the ability for HTTP headers to authenticate a user via REMOTE_USER or AUTH_USER 

**Describe the benefits this feature would bring to BookStack users**
This would round out the authentication features of this application allowing systems that rely on central authentication to automatically login users based on the value included in the above header. A bonus would be to allow auto registration for those not in the user table similar to the other authentication plugins.

ssddanbrown:
Hi @tiredofit,
Thanks for the suggestion.

Could you confirm what systems or standards use these headers? I'm apprehensive to support further authentication mechanisms due to the focus they divert from enhancing the core platform.

tiredofit:
Hi there. REMOTE_USER or a similar variable is used in quite alot of intranet style setups and typically found as a companion to LDAP in software applications. Basically it rounds out those who are running their own authentication backend to quickly connect to 3rd party systems in an enterprise level. It's used both on Windows and Linux server environments. 
I have seen other systems that are built with Laravel utilizing the [ADLDAP2](https://github.com/Adldap2/Adldap2-Laravel/blob/a8ccbd5ecd97965761e5bb30a4bdf5f572f49797/docs/auth/middleware.md) module supporting it.

In short, how it works is that the web server passes the $_SERVER_REMOTE_USER variable to the PHP engine which would then circumvent the requirement for logging in. While I can certainly support the other methods that are in use (I am using SAML now) 
I have an interesting use case that may make a bit more sense as to why it's requested.

Server would have 8 shelves - and one of the shelves would be publicly viewable. For those who were already logged into the central authentication system, the header would be passed and upon their visit to the site, they would see all 8 shelves. Yes, they would be able to click the login button, however it would be a source of confusion to the users to go that extra step. (Elementary school learners). 

You could also pass additional variables such as `groups` which would be similar to how you are fetching groups from LDAP and SAML to map against the Bookstack applications.

I could provide access to a development authentication server, however the easiest way to emulate this would be to add a value in the nginx php configuration such as `fastcgi_param REMOTE_USER (username);`

ssddanbrown:
Thanks for explaining.
I'l therefore leave this open for a while to get an idea of need/support otherwise I may close it off after a period of time. With these kind of additions, It's effort of social/support/maintenance required, rather than the effort of initial implementation, which makes me concerned, based upon my experience from the other auth options.

[...]

Issue #1500, opened 2019-06-20
LDAP Auto Login


MKCPC:
**Describe the feature you'd like**
I'd like a way for LDAP to support auto login, potentially using NTLM to pass the current users credentials enabling our users to be able to open the bookstack web page and automatically be logged in.

I'm not sure if this feature already exists but after looking through other feature/requests bugs i've not been able to find anyone else talking about it.

**Describe the benefits this feature would bring to BookStack users**
Ease of use for internal users not having to sign in every day to access documents.  Links given to team members would be instantly available rather than having to login each time.

ssddanbrown:
Thanks for writing up this request @MKCPC.

I can see how this may be useful. To be totally open, I'm not looking to spend much more efforts on the LDAP system unless it's an simple addition. Also, I'm not really familiar with NTLM in any way but from a quick search it appears to be microsoft/windows specific which would make me nervous about including support in this project. 

If the re-logging-back-in is the main pain point here, you may be able to quickly alleviate that by setting the following in your .env file:

```
SESSION_LIFETIME=240
```

The number is the lifetime of a user session in minutes. The default is 120.

Mant1kor:
@MKCPC , seems like you want to use SSO(Single sign-on) auth mechanism.
@ssddanbrown , NTLM is not recommended to use. It old and a bit not secure.
Better way is using Kerberos authentication. But all component must understand Kerberos tickets:
1. Web server
2. BookStack
3. System

Apache can do this after installing additional modules. Nginx you can build with third-party spnego module.
BookStack - don't support Kerberos.
Debian and RHEL based distributive supports Kerberos after some configuration.
So integration Kerberos-based SSO mechanism is not so easy as it seems.
P.S. If BookStack working on Windows server with IIS - 1 and 3 components support Kerberos auth by default.

MKCPC:
@Mant1kor Yes that's correct, a much more elegant way of explaining it!  I wasn't aware of the security issue around NTLM so Kerberos makes much more sense.  SSO is exactly what I would like to request, all of our team are logging on via a windows domain account so being able to pass the details of the current logged in user would make for a nice seamless user experience.

@ssddanbrown Thanks for session lifetime setting, i've updated that in the mean time to provide a session time spanning 8 hours so once a member of our team logs in once they will not need to login for the rest of the day at least.

ethanmdavidson:
[...]

PR #2289, opened 2020-09-29
Middleware to handle SSO variables

injektion:
Simple middleware to check if SSO variables are set, and auto login if they are.

As per issues #2223 #2180 #1500

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -13,6 +13,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \BookStack\Http\Middleware\TrimStrings::class,
         \BookStack\Http\Middleware\TrustProxies::class,
+        \BookStack\Http\Middleware\SsoLogin::class
     ];
 
     /**
--- a/app/Http/Middleware/SsoLogin.php
+++ b/app/Http/Middleware/SsoLogin.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace BookStack\Http\Middleware;
+
+use Closure;
+use Illuminate\Support\Facades\Auth;
+use BookStack\Auth\User;
+
+class SsoLogin
+{
+    /**
+     * Check if SSO variable is set and auto login if they are.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+
+        // auto login if SSO env variable is set (ie: APP_SSO_KEY=HTTP_MAIL which will look at $_SERVER['HTTP_MAIL'])
+        //  note: remember to "artisan config:clear" when updating the .env file
+
+        // from .env
+        // APP_SSO_KEY=HTTP_MAIL   <-- this is the $_SERVER variable to check
+        // APP_SSO_DISCOVER=email  <-- this is the field to check exists in Users table
+
+        $sso_key = env("APP_SSO_KEY", false);
+        $sso_discover = env("APP_SSO_DISCOVER", false);
+
+        if ($sso_key && $sso_discover) {
+            if (Auth::check()) {
+                // they're already logged in so no need to check for SSO match.
+            } else {
+                // if they're not already logged in try and look up an existing user
+                $uid = @User::where($sso_discover, $_SERVER[$sso_key])->first()->id;
+                if ($uid) {
+                    Auth::loginUsingId($uid, true);
+                }
+            }
+        }
+
+        return $next($request);
+    }
+}


ricardojlrufino:
This can be important for integrating automatic authentication for help pages in third-party software

ssddanbrown:
Thanks for offering this @injektion and apologies for my massively late reply.

I've been hesitant to merge this since, while it does provide some likely desired functionality, it would not be enough to close any of the linked issues. They all would need a slightly different implementation and/or would desire a much expanded set of options.

This is an area where configurability and exact implementation may vary wildly between use-cases so I'm not sure it's going to be something we can easily cater for by default.

Instead, the [logical theme system](https://github.com/BookStackApp/BookStack/blob/release/dev/docs/logical-theme-system.md) could be utilised to setup a custom logic implementation as needed. I've ensured this has a couple of middleware-based events included with the functionality of this PR in mind. For example, To roughly translate your pull request code, a `functions.php` file within the theme system would look like this:

```php
<?php

use BookStack\Access\LoginService;
use BookStack\Facades\Theme;
use BookStack\Users\Models\User;
use Illuminate\Http\Request;

Theme::listen(ThemeEvents::WEB_MIDDLEWARE_BEFORE, function (Request $request) {

    // The name of the header which will contain the value to identify
    // the authorized user. You need to make sure this header is managed by the web server
    // or auth proxies before the request hits BookStack, so end users can't control the
    // value of this header.
    $authHeaderName = 'http-mail';

    // The BookStack user property that we'll match the header value
    // with to identify the user to authentication.
    $authUserAttribute = 'email';

    $authVarValue = $request->headers->get($authHeaderName);
    if ($authVarValue && !auth()->check()) {
        $user = User::query()->where($authUserAttribute, '=', $authVarValue)->first();
        if ($user) {
            $loginService = app()->make(LoginService::class);
            $loginService->login($user, 'custom_header_auth');
        }
    }
});
```

This can be implemented on an existing, current BookStack install without altering core app files. This kind of approach then allows customization and configurability where needed.
Full details on setting this up can be seen in the logical theme system docs: https://github.com/BookStackApp/BookStack/blob/release/dev/docs/logical-theme-system.md

_**Note**: Logic within the logical theme system is unsupported and unofficial. It has not been fully tested, nor have all angles of security been considered. BookStack updates may interfere with, or break, logic used._

I'll therefore close off this PR but if you have any objections to the above or any feedback regarding the logical theme system approach feel free to continue to comment.

injektion:
Just circled back to this months later, and your template based solution was much better than middleware clog - switched our installs over to this instead of the SSO middleware and it's working smoothly... thank you for the pointer.

ssddanbrown:
Thanks for confirming @injektion. Good to know this theme-based solution works for people!

Namron10410:
Hello,
I want to perform an autologin on Bookstack with LDAP / SSO
I want to do it with HTTP / REMOTE_USER requests
I did not succeed following your advice above
Could you please explain to me the procedure to follow (in more detail)
thanks in advance

cjs59:
> Responding to https://github.com/BookStackApp/BookStack/pull/2289#issuecomment-857164663

For anyone using this code with more recent versions of BookStack, change the first use line to:

use BookStack\Users\Models\User;

pghope:
> For anyone using this code with more recent versions of BookStack, change the first use line to:
> 
> use BookStack\Users\Models\User;

Huge thanks for adding this here @cjs59.

I was seeing failed new login attempts in version 23.06 onwards - got a 500 error and generic error page 'An unknown error occurred'.

ssddanbrown:
I've updated [my example above](https://github.com/BookStackApp/BookStack/pull/2289#issuecomment-857164663) to align with current BookStack versions (tested on v23.10.2).
This also updates the logic to log-in via the app's `LoginService` which allows this to work with some other system features where desired (MFA, Audit-log login tracking, theme system, email confirmation).

cjs59:
The updated example only works for HTTP request headers - if you need a server header like REMOTE_USER, you need this line instead:

$authVarValue = $request->server->get($authHeaderName);

mig5:
This works great, the only issue is I found that there could be a redirect loop if you had MFA enabled on your account in Bookstack. I had to disable it to avoid a strange 302 redirect on the /mfa/verify route. (Fortunately I can set up MFA with my SSO/IDP provider).

JasperE84:
> This can be implemented on an existing, current BookStack install without altering core app files. This kind of approach then allows customization and configurability where needed. Full details on setting this up can be seen in the logical theme system docs: https://github.com/BookStackApp/BookStack/blob/release/dev/docs/logical-theme-system.md
> 
> _**Note**: Logic within the logical theme system is unsupported and unofficial. It has not been fully tested, nor have all angles of security been considered. BookStack updates may interfere with, or break, logic used._
> 

Hi @ssddanbrown, any chance we can make this procedure also update the LDAP group memberships for the logged in user? I've successfully implemented the SSO by header using the sample provided, but once the user made his first manual login and the user exists in bookstack, the LDAP group memberships do not seem to be updated from the AD anymore. (Considering that I also guess the user will remain a bookstack user with associated rights once the LDAP backend user is deleted)

JasperE84:
> Hi @ssddanbrown, any chance we can make this procedure also update the LDAP group memberships for the logged in user? 

This idea seemed rather complex to implement, I've now moved to SAML with my bookstack Entra App registration (with Entra application proxy, providing an authed Entra reverse proxy to my LAN docker container (just like cloudflared proxy can). No need to open ports on my router this way and with SSO from Microsoft Entra seamless sign on. Works quite nice, sill trying to figure out now how to not require the user to click the login link top right at all now. (Once login is clicked it auto chooses SAML using the env variable, but why require the user to click login at all)

ssddanbrown:
> Works quite nice, sill trying to figure out now how to not require the user to click the login link top right at all now. (Once login is clicked it auto chooses SAML using the env variable, but why require the user to click login at all)

@JasperE84 Is that with the `AUTH_AUTO_INITIATE=true` option enabled? That option auto-starts login if there's only a single SSO auth option. If public access is enabled though, the user will have to first navigate to the login screen since they may be considered a guest/non-logged-in user. If public access is disabled, this should be fairly seamless, as users are redirected to the login screen automatically (if not already logged in), where authentication will be auto initiated.

JasperE84:
> If public access is disabled, this should be fairly seamless, as users are redirected to the login screen automatically (if not already logged in), where authentication will be auto initiated.


Thats excellent, I should have thought of that. Thanks! I will disable public access tomorrow and try that out!

PR #2296, opened 2020-10-01
Fix build:js:watch not building at first launch in Docker

timoschwarzer:
This PR adds the `--initial` flag to `chokidar`. This ensures that the JS resources get built immediately after starting the `build:js:watch` script. I also passed the `SHELL` environment variable down to node in the node entrypoint because `chokidar` does not run without it.

tl;dr: this makes it possible again to clone the repository, follow the Docker steps in the README and having a working dev installation after that :)

### DIFF
--- a/dev/docker/entrypoint.node.sh
+++ b/dev/docker/entrypoint.node.sh
@@ -5,4 +5,4 @@ set -e
 npm install
 npm rebuild node-sass
 
-exec npm run watch
+SHELL=/bin/sh exec npm run watch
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks @timoschwarzer once again for this. Good tip on the chokidar command, had not come across the affects of no initial build yet that but I know that would have been a future annoyance. And thanks for maintaining the docker setup, Gave it a rebuild and run and all seemed to work as expected.

PR #2298, opened 2020-10-01
Install composer dependencies in Docker entrypoint

timoschwarzer:
I think it is an unnecessary step to install composer dependencies manually when using the Docker development environment. NPM dependencies also get installed automatically so this is just one less thing to think of :)

I also noted that you need to set `APP_ENV` to `local` because database migrations won't run otherwise. (Command is skipped in production)


ssddanbrown:
Thanks @timoschwarzer, All seems to work well when tested, especially with composer v2 speed improvements, and it's nice to streamline the experience.

PR #2299, opened 2020-10-01
Allow users to create roles when creating users

timoschwarzer:
I often create a new role for each user to have granular control over user permissions. This PR makes this easier by adding an option to create a new role on the fly while creating a new user:

image

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -2,6 +2,7 @@
 
 use BookStack\Auth\Access\SocialAuthService;
 use BookStack\Auth\Access\UserInviteService;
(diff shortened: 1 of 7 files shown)


ssddanbrown:
Thanks @timoschwarzer for providing this and apologies for my late feedback.

While I can definitely see the use-case/benefits for some people at the moment, this is not a feature that I'd specifically want to add. Realistically per-user roles are a workaround for a lack of user-level item permissions which is something I'd still like to attempt later on in the roadmap. Additionally, encouraging users to potentially create one-off roles where not explicitly needed could lead to performance considerations, especially with large user-counts. Therefore I'm going to close this off, Sorry.

Issue #493, opened 2017-08-31
[FEATURE] Header/Footer for PDF Export


sirup:
### For Feature Requests

For PDF etc. Export it would be nice to set a global Header/Footer to identify where the file is from etc. Maybe also wih Dynamic Fields like Version / Date / Creator.

Chris-ZA:
I would support this request. I was thinking about this the other day when I exported a book. 

Our current Microsoft Word document template have this and it helps a lot to have this information on a PDF or printed document.

aljawaid:
This would be ideal. Shame we can't currently add the logo or other bits and have the option to remove the 'last edited by' information. Page numbers and a possible optional front cover page would be good too.

lynchpatrickj:
Agree, good request.

PWA-GouldA:
I also support this feature.

The feature should have items such as:
- page numbers
- chapter title/number
- book title
- draft (when not finalised)
- logo
- date/time
- author/editors/etc

badsmoke:
is there any news on this topic?

injektion:
Hey @ssddanbrown - I've pushed a branch ( [https://github.com/injektion/BookStack/tree/exportheaders](https://github.com/injektion/BookStack/tree/exportheaders) ) which might help with this... For exports, it'll display a page number in the footer and allow users to create a page to include as a header on every page.

Most of the juice is in [partials/export-header-footer.blade.php](https://github.com/injektion/BookStack/blob/exportheaders/resources/views/partials/export-header-footer.blade.php) which is now included in all the /views/[books|chapters|pages]/export.blade.php and checks for a couple .env variables

```
EXPORT_HEADER_PAGE_SLUG=header
EXPORT_SHOW_PAGE_NUMBERS=true
```

There's some extra CSS added to partials/export-styles.blade.php which sets up a counter for pages, and if the .env is true will drop it into a footer div

```
        .pagenum:before {
            content: counter(page);
        }
```

As for the header, I'm currently letting the user define a page as their header, and then in [partials/export-header-footer.blade.php](https://github.com/injektion/BookStack/blob/exportheaders/resources/views/partials/export-header-footer.blade.php) using the slug of that page to import a header....

```
@if(env("EXPORT_HEADER_PAGE_SLUG"))
<div class="header">
    @php
        $PAGE = new \BookStack\Entities\Page();
        $page = $PAGE::where("slug", "=", env("EXPORT_HEADER_PAGE_SLUG"))->first();
[...]

PR #2311, opened 2020-10-06
added custom page header and page count footers to export

injektion:
#493 

[submitting as a PR for visibility/feedback]

Hey @ssddanbrown - I've pushed a branch ( https://github.com/injektion/BookStack/tree/exportheaders ) which might help with this... For exports, it'll display a page number in the footer and allow users to create a page to include as a header on every page.

Most of the juice is in partials/export-header-footer.blade.php which is now included in all the /views/[books|chapters|pages]/export.blade.php and checks for a couple .env variables

```
EXPORT_HEADER_PAGE_SLUG=header
EXPORT_SHOW_PAGE_NUMBERS=true
```

There's some extra CSS added to partials/export-styles.blade.php which sets up a counter for pages, and if the .env is true will drop it into a footer div

```
        .pagenum:before {
            content: counter(page);
        }

```
As for the header, I'm currently letting the user define a page as their header, and then in partials/export-header-footer.blade.php using the slug of that page to import a header....

```
@if(env("EXPORT_HEADER_PAGE_SLUG"))
<div class="header">
    @php
        $PAGE = new \BookStack\Entities\Page();
        $page = $PAGE::where("slug", "=", env("EXPORT_HEADER_PAGE_SLUG"))->first();
        $page->html = (new \BookStack\Entities\Managers\PageContent($page))->render();
    @endphp
    {!! $page->html !!}
</div>
@endif
```

It's working fine to allow the user to create their own headers, add some page numbering, and export, but I think there's probably a better flow to how to get the user to define the page to use as their header (configuration page?)... Do you have any recommendations on where you'd like to see the header get defined, or do you think this will work for those who want it?  Additional features or configurations you'd need to include this?

### DIFF
--- a/resources/views/books/export.blade.php
+++ b/resources/views/books/export.blade.php
@@ -20,6 +20,8 @@
         ul.contents ul li {
             list-style: circle;
         }
+
+
         @media screen {
             .page-break {
                 border-top: 1px solid #DDD;
@@ -30,6 +32,9 @@
     @include('partials.custom-head')
 </head>
 <body>
+
+@include('partials.export-header-footer')
+
 
 <div class="page-content">
 
--- a/resources/views/chapters/export.blade.php
+++ b/resources/views/chapters/export.blade.php
@@ -23,6 +23,8 @@
 </head>
 <body>
 
+@include('partials.export-header-footer')
+
 <div class="page-content">
 
     <h1 style="font-size: 4.8em">{{$chapter->name}}</h1>
--- a/resources/views/pages/export.blade.php
+++ b/resources/views/pages/export.blade.php
@@ -33,6 +33,8 @@
 </head>
 <body>
 
+@include('partials.export-header-footer')
+
 <div id="page-show">
     <div class="page-content">
 
--- a/resources/views/partials/export-header-footer.blade.php
+++ b/resources/views/partials/export-header-footer.blade.php
@@ -1,0 +1,27 @@
+{{--
+
+    Fills in the header and footers styled in partials/export-styles.blade.php
+
+    Relies on the following in .env
+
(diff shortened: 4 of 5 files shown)


ssddanbrown:
Hi @injektion, Sorry for me very late reply.  I've just tried to test and finish this off to get it implemented but come across some challenges. 

I done some changes in 370d07fd60110c4ad8e006ab960387421340d831 to change how the header & footer is set, I changed things to use the theming system to set custom header/footer content instead of env options and existing pages. One of the challenges was getting the footer/header heights working with the page margin. Would be awkward but we could expose the options for these in some default header/footer content.

The main issue for this is WKHTMLtoPDF compatibility. WKHTMLtoPDF uses a completely different system for headers and footers. I did try to work around this (By extracting out headers/footers from the content and re-feeding in the support way) but It's even trouble to share the footer content from the DOMPDF solution since the margin control and page numbering are done in very different ways. I'm not really happy to implement a solution that just works for DomPDF, especially since we often guide people to use WKHTMLtoPDF in the event of issues. A random additional issue that I came across was that WKHTMLtoPDF completely fails to render any footer/header without using a patched QT version of the library. 

Not sure where to go with this. Might look to see what WKHTMLtoPDF-only support would look like but it's still not ideal. Will have a think on it.

injektion:
@ssddanbrown Admittedly when I was piecing together the header/footer solution, I was working in a DOMPDF environment.  Makes sense not to go that route if it doesn't work in the recommended WKHTMLtoPDF export.

You've got a better grasp on the project as a whole (and the changes since this was done back in last Oct), so if something comes together in your thoughts that you want to offload and throw your ideas this way, let me know and I can help take it off your plate.

PR #2320, opened 2020-10-12
Import thumbnail photos when LDAP users are created.

jasonhoule:
My BookStack instance is using Active Directory for authentication. We wanted to be able to automatically import the user photos that we already have stored in Active Directory.

This change, controlled by config values in the config/services.php file, allows for user avatars to be imported from the LDAP thumbnailPhoto attribute. Users can still update their images from the profile page.

### DIFF
--- a/app/Auth/Access/Guards/LdapSessionGuard.php
+++ b/app/Auth/Access/Guards/LdapSessionGuard.php
@@ -117,7 +117,17 @@ class LdapSessionGuard extends ExternalBaseSessionGuard
             'password' => Str::random(32),
         ];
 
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Thanks a bunch for this @jasonhoule. Now merged for the next feature release.

I followed this up with some changes in df0e03cd07db5ff2dbd16a27ccfcc2ca38402e46. I removed the `LDAP_IMPORT_THUMBNAIL_PHOTOS` option and now the functionality will run depending on whether the `LDAP_THUMBNAIL_ATTRIBUTE` option has been defined. I also tweaked the logic to import on every login if the user does not have an image already set.

PR #2328, opened 2020-10-16
Fixes from laravel-microscope package

imanghafoori1:
This PR adds some fixes discovered and applied by laravel-microscope package.

### DIFF
--- a/app/Auth/Permissions/PermissionService.php
+++ b/app/Auth/Permissions/PermissionService.php
@@ -3,11 +3,8 @@
 use BookStack\Auth\Permissions;
 use BookStack\Auth\Role;
 use BookStack\Entities\Book;
-use BookStack\Entities\Bookshelf;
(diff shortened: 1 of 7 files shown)


ssddanbrown:
Thanks for these fixes @imanghafoori1.

Note, I removed the inclusion of the package in 82e671a06d2d18757a9ff0be83f11e0a7053b3d5. I couldn't seem to successfully run all checks when giving the package a go. It would error on the controller/route checks due to the helpers file. I had a quick look at your package's code but could not see why. The helpers file was included in a fairly forced/awkward manner so I changed that up (Seen in the above commit) to use composer to load the file instead but that did not seem to help prevent the error.

imanghafoori1:
I think you need to check for the existence of the functions like laravel core does:
```php
if (!function_exists(...)) {
 function ...(){
     ... 
  }
}
```

PR #2333, opened 2020-10-19
Added translation for Norwegian (Bokmål)

Swoy:
This will add translations for Norwegian to BookStack. It is identified by the langID no_NB


Swoy:
had a small issue that caused it to fail

Issue #2349, opened 2020-10-31
Automatic Restored Revision Changelog Summary Text


rondaa:
**Describe the feature you'd like**

When using the "Restore" feature to create a new revision that's a clone of a previous revision, a changelog summary should automatically be applied, which reads `Restored from #{revision number} "{cloned revision's original changelog summary}"`. I am open to a suggested alternative template for a restored revision changelog summary.

This replaces the current behavior, where restored revision changelog summaries are blank.

**Describe the benefits this feature would bring to BookStack users**

Every revision, according to best practices, should have a changelog summary. Any missing changelogs mean that a click on the revision changes or preview links are necessary to see what the revision contains, which defeats the purpose of having a changelog property available. With this feature, users' revision histories will be 100% in line with what is achieved through following best practices.

**Additional context**
![image](https://user-images.githubusercontent.com/1410433/97770678-ce9e4080-1b0b-11eb-88b5-ef3a6da4884b.png)
In this image, revisions 3 and 4 are restored from edits 1 and 2, respectively. This is not made clear in the changelog summary, nor will it be easily verifiable from looking at the "Changes" view for each respective revision. These would be unacceptable submissions if they were user-edited revisions.

![image](https://user-images.githubusercontent.com/1410433/97770720-276dd900-1b0c-11eb-863c-60e29d96a925.png)
This history log of a Wikipedia article shows an instance where an edit (middle row) is restored (top row) to a previous version (bottom row). Underlined text, which was automatically generated, shows which version the article was restored from. This feature is what inspires my feature request.

rondaa:
![image](https://user-images.githubusercontent.com/1410433/97952995-2982a280-1d6d-11eb-911d-4b2d4ea26269.png)

Working locally. 2 new lines and a localization string. Plan to run tests and PR tomorrow. Haven't heard a response but I'm still open to a different templating

ssddanbrown:
As per #2353, this is now merged to be in the next release.

PR #2353, opened 2020-11-03
Automatic Restored Revision Changelog Summary Text

rondaa:
I made this in support of my request in #2349 

- [x] Ran phpcs (PSR-2)
- [x] PHPUnit

image


ssddanbrown:
Thanks for this @rondaa.

While testing this PR i noticed the revision saving, upon restore, was actually incorrect and that the revision, which the summary was saved against, would be the pre-restored content. Have updated this in 024b0d8a64059d6c125c0a91411a0f7a492a9d6a and added testing to cover that and testing for this change summary text. 

PR #2376, opened 2020-11-18
LDAP: Added TLS support

Body4:
A new option is available in `.env` file: LDAP_START_TLS=true|false

It is necessary to add the option `TLS_CACERT <path_to_bundle.pem>` in `/etc/openldap/ldap.conf` to make it work. Because, unfortunately, the LDAP options is simply ignored by (my) PHP (i don't know why). Maybe yours it will work?


ssddanbrown:
Thanks for this @Body4 and sorry for the delayed response,
I followed this up with some further changes in af032f89932290b9f1217e80733447da6c039f55.

I found luck in my certificate being verified by importing the CA certificate into the BookStack host system's trusted certificate store. 

This is now in master, to be part of the next feature release.

Issue #2348, opened 2020-10-28
Featue: support for meta embeds


codemicro:
**Describe the feature you'd like**
Currently, BookStack has no support for meta embeds that are commonly used on social media sites, similar to this:

![DiscordPTB_2020-10-25_23-25-41](https://user-images.githubusercontent.com/25284325/97381810-c8b51f00-18c1-11eb-9872-d8ca9a94b1fd.png)

Perhaps this is something that would fit into BookStack? For example, the meta embed for a page of text could contain info like the page title, the first image (if any) in it, and a small amount of the content. The embed for a book could show its cover image and some info about the pages inside of it. Et cetera, et cetera.

Currently, the only option is to use some custom header content, which results in a fixed embed for every page.

**Describe the benefits this feature would bring to BookStack users**
This gives BookStack links that're shared on services like Discord and Twitter more credibility and makes them look a bit better... okay, it's basically entirely an aesthetical thing. You have to admit they do look nice, though.

**Additional context**
[CSS tricks: The Essential Meta Tags for Social Media](https://css-tricks.com/essential-meta-tags-social-media/)
[The Open Graph protocol](https://ogp.me/)
Example meta tags from my site:
```html
<meta name="description" content="UK based student and programmer">
<meta property="og:title" content="Thomas Pain">
<meta property="og:site_name" content="tdpain.net">
<meta property="og:description" content="UK based student and programmer">
<meta property="og:url" content="https://www.tdpain.net/">
<meta property="og:image" content="https://cdn.tdpain.net/avatar.png">
```

ssddanbrown:
Thanks for the request with clear details @codemicro.

I’d be very much open to having something like this within BookStack. There have been prior similar requests but they have all requested additional controls to set such tags with custom values. As long as this was based on existing content instead of into reducing additional inputs, this would be something I’d be happy to be within the platform.

codemicro:
Hi Dan!

Great news! I don't think there should necessarily be any extra configurable options just for the meta embeds, I imagine the existing data that exists would be plenty to populate one.

I'd submit a PR for it, but unfortunately I don't know PHP at all well. Go is more my thing.

james-geiger:
Hey @ssddanbrown, been quiet a while but can pick this one up if you've not started on it already.

@codemicro This is a really well documented issue, so I think there's a good place to start.  Are there any tags that would be "must haves" versus just "helpful" for the content types: Shelf, Book, Chapter, and Page?

[...]

PR #2393, opened 2020-12-03
Create content meta tags for Shelf, Book, Chapter, and Page

james-geiger:
Draft PR to implement #2348.

To Do:

- [x] Page
- [x] Chapter
- [x] Book
- [x] Shelf
- [x] Default


james-geiger:
This PR implements the feature request in #2348 for media embeds that adhere to Open Graph specifications.  Most sites that support Open Graph require a title, URL, description, and image.

A new method on the Page model has been created to return either the first image in the HTML body of the page or the book cover to which the page belongs.

The default for each image relies on the book default image, which returns as base64.  I'm not sure how well this will be supported across different sites, so I welcome additional feedback.

`phpunit` tests are failing, so might need some additional guidance.  Some make sense, i.e `test_iframe_js_and_base64_urls_are_removed()` but not others.

app/Entities/Models/Page.php
```diff
@@ -126,4 +126,18 @@ public function forJsonDisplay(): Page
...
+
+    public function getCoverImage(): string
+    {
+        $dom = new \DomDocument();
+        $dom->loadHTML($this->html);
+        $images = $dom->getElementsByTagName('img');
+
+        try {
+            $cover = $images->length > 0 ? $images[0]->getAttribute('src') : $this->book->getBookCover();
+        } catch (Exception $err) {
```
Abijeet, line 138: We should either add a `use \Exception;` statement at the top or put `\Exception` here.

app/Entities/Models/Page.php
```diff
@@ -126,4 +126,18 @@ public function forJsonDisplay(): Page
         $refreshed->html = (new PageContent($refreshed))->render();
         return $refreshed;
     }
+
+    public function getCoverImage(): string
+    {
+        $dom = new \DomDocument();
```
Abijeet, line 132: I would put the logic to fetch the first page image in `PageContent` class, if that returns null, we can return the book cover.

resources/views/chapters/show.blade.php
```diff
@@ -6,6 +6,11 @@
     option:entity-search:entity-type="chapter"
 @stop
 
+@push('social-meta')
+    <meta property="og:description" content="{{ Str::words($chapter->description, 50, '...') }}">
```
Abijeet, line 10: Humm, as per this link, the og description can be set to 110 characters: https://stackoverflow.com/a/35817780/903324

resources/views/pages/show.blade.php
```diff
@@ -1,5 +1,10 @@
 @extends('tri-layout')
 
+@push('social-meta')
+    <meta property="og:description" content="{{ Str::words($page->html, 50, '...') }}">
```
Abijeet, line 4: Instead of `html`, what do you think of using the `text` property instead? That would avoid tog:description` having HTML content.

resources/views/base.blade.php
```diff
@@ -11,6 +11,13 @@ class="{{ setting()->getForCurrentUser('dark-mode-enabled') ? 'dark-mode ' : ''
     <meta name="base-url" content="{{ url('/') }}">
     <meta charset="utf-8">
 
+    <!-- Social Cards Meta -->
+
```
Abijeet, line 15: minor nit: get rid of this newline

Abijeet (changes requested):
Thank you for your contribution. Please see my comments.

Abijeet:
I think if we replace HTML with text (as mentioned here: https://github.com/BookStackApp/BookStack/pull/2393#discussion_r563317284) , we can fix the failing tests.

resources/views/chapters/show.blade.php
james-geiger: Changed to utilize `Str::length` helper with length set to 100.

resources/views/pages/show.blade.php
james-geiger: Agreed, updated to use `text` property.

app/Entities/Models/Page.php
james-geiger: @Abijeet See changes in #48587d2 and let me know your thoughts.

app/Entities/Models/Page.php
james-geiger: @Abijeet See changes in 48587d2 and let me know your thoughts.

james-geiger:
Unit tests continue to fail even after changing code to utilize `text` property of pages... Not sure if it has to do with this branch.

ssddanbrown:
Thanks for this @james-geiger.

I've now merged and followed this up with 265f5db03f3f08665bb113191f83c49e5dae53a7 with some tweaks and testing.

I've scaled back the implementation a bit and removed an logic that guesses images, uses default backup options or uses images of related content. I'd prefer to keep this logic minimal and specific to content that has been specifically chose by the user. Would prefer omission of content rather than any potential confusion. Hope that's understandable. 

james-geiger:
@ssddanbrown Absolutely understand the scale back.  I think it's worth noting that, because different sites implement this spec differently, many require an image to be "valid" and won't display without an image.  This is prevalent on the "social" sites, could be different on others.

I mention this not to suggest that a default be enforced, but so that users can be aware.

PR #2410, opened 2020-12-13
PHP 8 Support

### LINKED ISSUE #2388: PHP 8 support
Now that PHP 8 has been released we should update the app to work on php8. 

Laravel framework v6.20 adds PHP8 support: https://github.com/laravel/framework/releases/tag/v6.20.0
This does bump the minimum version of php to 7.2.5: https://github.com/laravel/framework/pull/34928

- [x] Update framework.
- [x] Go through dependencies and update where require.
- [x] Update GitHub actions to cover php 8.
- [x] Test locally

Due to php version change we'll need to add a notice to the update docs on release.

ssddanbrown:
#### Outstanding dependency issues

* doctrine/dbal
   * PHP8 support added in 2.12+, PHP7.2 dropped from 2.11+
   * Support across same stable version, could `or` the versions.
* onelogin/php-saml
    * Current lacks PHP8 support, [Mentioned here](https://github.com/onelogin/php-saml/issues/453).
* fzaninotto/faker (Dev)
    * Think we'll need to look to use https://github.com/fakerphp/faker instead. Check against current Laravel defaults for v6 & latest branches.
* phpunit/phpunit (Dev)
    * 8.5 installable but not full supported (No code coverage) on PHP8. 9.x branch does support php 8 but requires 7.3.
    * Could `or` the versions here.

#### Other Considerations

* Have removed laravel-stats package due to lack of use and change of license terms that I'm not confident we can meet.
* I think we'll need to remove the `config.platform` option in `composer.json` to ensure PHP 8+ systems and PHP 7 get their supported packages.
* Will we have to remove the composer.lock file? Since now we're likely to see different dependencies per platform? If so, Will need to double check versions in composer.json are accurately locked. May also lead to some instability or future unexpected problems, especially as user installs may be harder to replicate.
   * If removed we'd then need to change how updates are done since a `composer update` may be needed from second update onwards, due to a local non-changing composer.lock file.
* Could it be easier to do an extra php version change for BookStack, to get us past this. Jump to php7.4 soon. Might be worth seeing what Laravel is jumping to for next LTS then align sooner than expected.

---

Closes #2388


Abijeet:
Note to also update the Dockerfile: `dev/docker/Dockerfile` to use PHP 8

ssddanbrown:
Closing since this became superseded by commits in https://github.com/BookStackApp/BookStack/issues/2648#issuecomment-803405790

PR #2416, opened 2020-12-17
Fix issue with viewing export dropdown list

shubhamosmosys:
There is issue with viewing all the export list.

Step to reproduce the issue:
1. Login to bookstack.
2. Open any book or page.
3. Click on export.
4. There are three list in the dropdown but only two can be visible third one is hidden.

### DIFF
--- a/resources/sass/_layout.scss
+++ b/resources/sass/_layout.scss
@@ -272,6 +272,7 @@ body.flexbox {
     min-height: 50vh;
     overflow-y: scroll;
     overflow-x: hidden;
+    height: 100%;
     scrollbar-width: none;
     -ms-overflow-style: none;
     &::-webkit-scrollbar {
@@ -281,9 +282,6 @@ body.flexbox {
   .tri-layout-middle-contents {
     max-width: 940px;
     margin: 0 auto;
-  }
-  .tri-layout-right-contents {
-    height: 1000px;
   }
 }
 
--- a/resources/sass/_layout.scss
+++ b/resources/sass/_layout.scss
@@ -282,6 +282,9 @@ body.flexbox {
     max-width: 940px;
     margin: 0 auto;
   }
+  .tri-layout-right-contents {
+    height: 1000px;
+  }
 }
 
 @include smaller-than($l) {


ssddanbrown:
Thanks for attempting a fix for this @shubhamosmosys.

Could `height: 100%` be used instead of a fixed px height please? The use of `1000px` results in the sidebar increasing the page height on smaller pages. Additionally, could this be applied to the `.tri-layout-left-contents, .tri-layout-right-contents` rule at line 267 within the same file since this would apply to both of the sidebars and saves us an extra rule being defined.

shubhamosmosys:
Hi @ssddanbrown,
We can't use `height: 100%` because that will violate the sticky behaviour of `.tri-layout-left-contents,` and  `.tri-layout-right-contents` on scrolling page.
On scrolling page, left and right side content should be sticky in nature until they get scroll.
 
Issue can be seen below:
image

Expected Behaviour on scrolling page:
image

Also since we can have many pages inside the book. So, the size of `.tri-layout-left-contents` should not be fixed.
Hence, we're fixing only `.tri-layout-right-contents` size.

ssddanbrown:
Hi @shubhamosmosys,

With a quick test in chrome and firefox, adding `height: 100%` [about here](https://github.com/BookStackApp/BookStack/pull/2416/files#diff-55b7897471a39ff633308eae3ff5ad1e10f2f0116ef037447549802c93b25a0aR268) did not affect the sticky behaviour at all. Are you testing in a different browser or perhaps testing the style elsewhere?

>  Also since we can have many pages inside the book. So, the size of .tri-layout-left-contents should not be fixed.
Hence, we're fixing only .tri-layout-right-contents size.

Both of the sidebars have the potential to be taller or shorter than the viewport. In that regard I don't know why we need to treat them differently.

shubhamosmosys:
Hi @ssddanbrown,
Thanks for the suggestion. Adding `height:100%` is working fine.
Updated the PR.

ssddanbrown:
Thanks @shubhamosmosys, Will merge in for the next release.

Issue #2454, opened 2021-01-06
Show difference between two page revision


johnroyer:
**Describe the feature you'd like**
It is good to add a function which can show the difference between two revision.

This is how Mediawiki do below:

![revision diff tool](https://user-images.githubusercontent.com/114491/103752788-1f1ea680-5045-11eb-8b38-f412ae392c89.png)

----

**Describe the benefits this feature would bring to BookStack users**
It take a lot of time using preview page to know what is changed in revisions. If there is an diff tool, review changes is a lot more easier.

ssddanbrown:
Hi @johnroyer,

There's already a "Changes" option when looking at the revisions of a page. Is there something additional to that which you need?

johnroyer:
oh .... maybe I have not grade yet. Give some timeto upgrade.

soerry.

johnroyer:
I know why now.

The is a button called `change`. But it translate to `說明 (instruction)` in Chinese. That why I did not found the function.

----

I will send an PR to fix this problem.

johnroyer:
Issue closed. Thanks a lot !

ssddanbrown:
@johnroyer No problem, thanks for reporting and providing the updated translations!

PR #2460, opened 2021-01-07
Fix wrong translation

johnroyer:
fix issue #2454

update translate that make misunderstanding.


johnroyer:
Update translation on Crownin. PR close.

Issue #2467, opened 2021-01-09
Healthcheck Endpoint


modem7:
#### API Endpoint or Feature

A healthcheck endpoint where it returns an OK/200 if all is good.

#### Use-Case

Docker and Traefik both allow healthchecks to be performed, this is useful if there is a failure and the container needs rebooting/alerting that something isn't right.

ssddanbrown:
Thanks for the request @modem7.

I'm thinking that this being part of the official API system may cause a lot more complexity than required (As you'd need to pass the right auth headers).

I think we could just add this as a publicly accessible, still JSON, endpoint. As long as we're not exposing anything private.
On request, We should probably perform:

- Simple DB query to ensure connection.
- Cache random value put/get check.

Return 200 status code unless any failure, in which case 500.

modem7:
The publicly accessible endpoint would definitely be useful, that way we could also have it on loadbalancers with reverse proxies as well as docker/grafana etc.

ssddanbrown:
Now included within 6eda1c1fb28142d432bf0ca05d7eaba9bf16f0d0, To be part of the next feature release (v0.32).

PR #2475, opened 2021-01-11
Made `ApiController` non-abstract

gertjankrol:
This fixes `php artisan route:list` (controllers can't be abstract because then they're not instantiable)
I haven't tested the full API functionality, as I don't use this (yet). I don't think this will cause any problems though.

### DIFF
--- a/app/Http/Controllers/Api/ApiController.php
+++ b/app/Http/Controllers/Api/ApiController.php
@@ -5,7 +5,7 @@ use BookStack\Http\Controllers\Controller;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Http\JsonResponse;
 
-abstract class ApiController extends Controller
+class ApiController extends Controller
 {
 
     protected $rules = [];


gertjankrol:
In my setup, I added a status endpoint to the API. This also broke due to the class not being instantiable :)

ssddanbrown:
Hi @gertjankrol,
Thanks for offering this.

Just to confirm though, as far as I can see the error which this fixes would only occur due to an route being added that utilises this ApiController class; This does not fix a bug that exists in the release code. Is that correct?

gertjankrol:
Hi, it also fixes `artisan route:list`, which is now failing due to the class not being instantiable

ssddanbrown:
@gertjankrol But only when you add your own routes? When testing on my system `artisan route:lists` works fine.

gertjankrol:
Oh, I'm sorry. Just confused it a bit, the problem is not in the release code :)

ssddanbrown:
@gertjankrol Okay, Thanks for confirming. In that case, I'd probably prefer to leave this class abstract as I really intend for it to act as a base for all other API controllers.

You mentioned adding your own status endpoint, If you have any input to provide on https://github.com/BookStackApp/BookStack/issues/2467 that would be much appreciated, then we can get something officially built-in.

gertjankrol:
Sure, I can help out with that. I'll try to get a PR ready for review tonight

PR #2481, opened 2021-01-12
Added status (health check) endpoints to the API

gertjankrol:
Related to issue #2467


gertjankrol:
I see codeclimate is failing, I could break it up in separate functions?

ssddanbrown:
Thank you very much for this @gertjankrol.

To be honest, this implementation was very extensive whereas I'd prefer a simpler implementation without the addition of any options. I've added this implementation in 6eda1c1fb28142d432bf0ca05d7eaba9bf16f0d0.

Thanks again and sorry that I did not use your code.

PR #2484, opened 2021-01-13
Disable autocomplete on the change password field

l1n:
This seems to catch people up on my instances a lot, since you can't upload a new profile picture with the change password fields filled in.

### DIFF
--- a/resources/views/users/form.blade.php
+++ b/resources/views/users/form.blade.php
@@ -74,7 +74,7 @@
             <div class="grid half mt-m gap-xl">
                 <div>
                     <label for="password">{{ trans('auth.password') }}</label>
-                    @include('form.password', ['name' => 'password'])
+                    @include('form.password', ['name' => 'password', autocomplete => 'new-password'])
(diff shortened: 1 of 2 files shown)


resources/views/users/form.blade.php
```diff
@@ -74,7 +74,7 @@
             <div class="grid half mt-m gap-xl">
                 <div>
                     <label for="password">{{ trans('auth.password') }}</label>
-                    @include('form.password', ['name' => 'password'])
+                    @include('form.password', ['name' => 'password', 'autocomplete' => 'new-password'])
```
Abijeet, line 77: TIL https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#preventing_autofilling_with_autocompletenew-passw

Abijeet (approved):
Looks good to me!

ssddanbrown:
Thanks for this @l1n and thanks @Abijeet for confirming.

Issue #2462, opened 2021-01-07
Show tags for search results


burnoutberni:
**Describe the feature you'd like**
I'd like to be able to see all tags of a given search result. This could look something like this:

![Screenshot from 2021-01-07 16-24-15](https://user-images.githubusercontent.com/1640933/103910202-d75f5400-5104-11eb-8cdf-ea23b7c45391.png)

One might also consider adding either a global or a search setting for this in order to make it optional, but personally I consider this useful and unintrusive enough to have around all the time.

**Describe the benefits this feature would bring to BookStack users**

We use BookStack as a internal knowledge base for our organization and have important metadata stored in our tags (such as datestamps if pages are related to an event). Showing this information would make it easier to pinpoint which search results are relevant. Probably that's the case for other users too.

**Additional context**
This feature request should be really simple to implement, but I wanted to double-check whether it has any chance of being added before starting a PR. Also, I have no experience with Laravel/Blade, so my implementation would probably be very basic (read: no extra settings, just adding the tags). If that sounds good to you, please let me know and I'll go ahead.

On a similar note, it would be great to use this tag metadata in other contexts too such as for sorting the search results (I already mentioned that in https://github.com/BookStackApp/BookStack/issues/577).

ssddanbrown:
Thanks @burnoutberni! Closing as per #2487.

PR #2487, opened 2021-01-15
Show tags of all search results

burnoutberni:
This is a basic implementation of #2462.

It's rather similar to the `$showPath` property that you can attach to any `entity-list`. Now there's another property called `$showTags` that adds all tags to the list items. I've only enabled it for the search results page, but this might come in handy in other circumstances too.

Obviously the search results link to the corresponding shelf/book/chapter/page, but the tags also contain links. This is an issue since nesting `<a>` elements is not allowed. To avoid this I've added a `$disableLinks` property to `tag-list`, which replaces the tag links with `<span>` elements.

### DIFF
--- a/resources/sass/_blocks.scss
+++ b/resources/sass/_blocks.scss
@@ -227,7 +227,7 @@
   border: 1px solid #CCC;
   overflow: hidden;
   font-size: 0.85em;
-  a, a:hover, a:active {
+  a, span, a:hover, a:active {
     padding: 4px 8px;
     @include lightDark(color, #777, #999);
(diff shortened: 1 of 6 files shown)


ssddanbrown:
Thanks for this @burnoutberni, All merged for the next feature release!

I did make some changes as described in 3eaf03a7ac77d3d2765541659312d45e10890824, these primarily tweaked the design a bit just so that the tags are a little more subtle, and hence less distracting, when listed within entities. I also changed the position since I felt the tags were really secondary to the name, path and hint text, hence should come afterwards. 

Screenshot from 2021-05-24 16-23-13

Issue #2217, opened 2020-08-08
Uploading a file with a dot in filename returns an error


mhulet:
**Describe the bug**
Uploading an image with filename `Screenshot 2020-08-04 at 21.42.34.png` returns an error. 
Uploading an image with filename `Screenshot 2020-08-04 at 214234.png` succeeds.

**Steps To Reproduce**
1. Edit a page
2. Click on the image icon in the WYSIWYG editor
3. Try to upload an image with the filename above

**Expected behavior**
Uploading an image with dot(s) in filename succeeds.

**Screenshots**

<img width="194" alt="Screenshot 2020-08-08 at 23 21 05" src="https://user-images.githubusercontent.com/293355/89719953-db5c3e00-d9cd-11ea-8295-352c6a3b0aed.png">

Screenshot is in French: _"file must have only one filename extension"_

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): 0.29.3
 - PHP Version: 7.3.17
 - Hosting Method (Nginx/Apache/Docker): Docker

ssddanbrown:
Hi @mhulet,
Thanks for submitting the clear request.
This is done on purpose since we try to keep the original file name in the image path for context, and having multiple dots in a filename can open a system to a certain range of vulnerabilities on various webserver setups.

Maybe we should change this to instead remove all but the last dot automatically?

ssddanbrown:
This has been incorporated as part of #2611 thanks to @Hecke29 and will be part of the next feature release.

Issue #2337, opened 2020-10-20
Can't upload macOS screenshots with default filename due to no_double_extension check


abulgatz:
**Describe the bug**
Mac screenshots taken with <kbd>control</kbd> + <kbd>command</kbd> + <kbd>shift</kbd> + <kbd>3</kbd> (or <kbd>4</kbd>) are automatically saved to the desktop as a png with filename: `Screen Shot YYYY-MM-DD at HH.MM.SS AM.png`.

The extra periods in the filename fail the `no_double_extension` check the Bookstack runs on all uploaded images, so macOS screenshots can't be uploaded to Bookstack without renaming them.

**Steps To Reproduce**
1. On macOS
2. Take a screenshot with <kbd>control</kbd> + <kbd>command</kbd> + <kbd>shift</kbd> + <kbd>3</kbd> or <kbd>control</kbd> + <kbd>command</kbd> + <kbd>shift</kbd> + <kbd>4</kbd>
2. Edit any page in Bookstack 
3. Insert image
4. Click to upload
5. Try to upload the screenshot just taken from the Desktop
6. Error: "The file must only have a single file extension"

**Expected behavior**
Bookstack shouldn't care about extra dots in a filename.

**Screenshots**
<img width="200" alt="image" src="https://user-images.githubusercontent.com/1438193/96538723-d124ae80-125e-11eb-8576-639e356657de.png">

**Additional context**
I don't think there's any reason that Bookstack should care about periods in a filename. If Bookstack needs an extension, it's easy enough to grab any text after the last dot. Is there something that would break if the `no_double_extension` was just removed? I can't think of any other web or desktop apps that reject extra periods in a filename.

ssddanbrown:
Thanks for reporting this @abulgatz.
Since this is already open under #2217 I'm going to close this off as a duplicate.

The `no_double_extension` check is done on purpose to avoid a range of potential path-based attacks in addition to avoiding some potential attacks based off of how Apache can interpret some file name/extension formats. 

It was a rather hastily added patch though when needing to do so in the name of security. As said in #2217, We could probably do better by just replacing/removing any additional periods within the name.

PR #2500, opened 2021-01-20
Allow image filename has more than one dot (".")

### LINKED ISSUE #2306: Ensure "images with dots in file-name" error message shows localised message
**Describe the bug**
A graphic drawn in diagrams.net and exported as png file cannot be inserted as an image. It is always said that it is a wrong file extension. Unfortunately I can't describe it better because it works strangely with other png images.

Would be important to fix it quickly.

**Steps To Reproduce**
Steps to reproduce the behaviour:
1. Go to page
2. Click on insert image
3. Insert png picture 

**Screenshots**

grafik

 - Exact BookStack Version v0.30.2
 - PHP Version: PHP/7.3.22
 - Hosting Method (Nginx/Apache/Docker): docker and the image is `linuxserver/bookstack:v0.30.2-ls106`

johnroyer:
I found an image validator in `Controller.php`. The rule denied double filename extension (filename which has more then one ".").

This patch should resolve #2306 and #2217.

### DIFF
--- a/app/Http/Controllers/Controller.php
+++ b/app/Http/Controllers/Controller.php
@@ -159,6 +159,6 @@ abstract class Controller extends BaseController
      */
     protected function getImageValidationRules(): string
     {
-        return 'image_extension|no_double_extension|mimes:jpeg,png,gif,webp';
+        return 'image_extension|mimes:jpeg,png,gif,webp';
     }
 }


johnroyer:
Unit test failed: `ImageTest::test_files_with_double_extensions_cannot_be_uploaded`

I notice that there is a unit test testing filename can not have double extension name on 193e2ffeb . Is there any reason to avoid multiple file extension name ?

ssddanbrown:
@johnroyer Thanks for offering this but yeah, this is done on purpose.

Please see the following comments for more info:
- https://github.com/BookStackApp/BookStack/issues/2217#issuecomment-687571191
- https://github.com/BookStackApp/BookStack/issues/2337#issuecomment-742850834

johnroyer:
@ssddanbrown Thanks. Is there any request is hanged? 

If no, I planed to add a block as footer.

PR #2510, opened 2021-01-26
Docker: Fix permission with node service by adding node as user

Abijeet:
See: https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user

Fixes the following error:

glob error:
[Error: EACCES: permission denied, scandir '/root/.npm/_logs'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path: '/root/.npm/_logs'
}

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,6 +39,7 @@ services:
   node:
     image: node:alpine
     working_dir: /app
+    user: node
     volumes:
       - ./:/app
     entrypoint: /app/dev/docker/entrypoint.node.sh


ssddanbrown:
Hi @Abijeet, Good to see you!

I could verify the issue but this seemed very strange to me. That `/root/.npm/_logs` path should be within the container so in theory should not be affected by users/groups on the host system. 

Looking into the node docker image, It looks like them may make use of a `node` user to run actions, hence why it may not have permissions to write to the `/root` path, coming from the user that's running the entry-point script. 

I found that setting a `user: node` option on the node image, within  `docker-composer.yml` would solve this problem without the need of additional steps or `.env` changes. Would you be able to test that on your system and let me know if that also works for you?

Abijeet:
> Hi @Abijeet, Good to see you!

Hi @ssddanbrown; Good to see everything is going well.

> I found that setting a user: node option on the node image, within docker-composer.yml would solve this problem without the need of additional steps or .env changes

Thanks for the hint, that does fix the issue.

Issue #1381, opened 2019-04-09
Navigation at the End of the Page


blackshrub:
It is better if there is a prev-next button at the end of every page rather than to navigate from the left sidebar. Prev-next button is more mobile-friendly.

ssddanbrown:
Thanks for the suggestion @blackshrub.
I'm surprised this has not been requested before tbh.

I guess the last page within a chapter should have a next button back to the chapter. Same for last chapter within a book, but returning to the book-level.

ssddanbrown:
# Merged Matching Issue

Originally posted by @Wookbert: 

**Describe the feature you'd like**
The name and UI concept clearly implies to organize your information in books with chapters and pages. Reading a book however comes with the experience and convenience of continuous reading through a simple page turn. This does not only apply for paper books, but also has been mimicked on any eBook reader.

On BookStack I'm missing this convenience however so far. Yes, we do have the Book Navigation in the left side panel next to the current page (hereunder; which BTW only follows your scrolling and thus is always in instant reach, if the browser window is wide enough, so that there’s enough space for a right side panel as well), but that’s just not as convenient as a simple page turn.

![image](https://user-images.githubusercontent.com/3104762/64079875-9d4e9880-cced-11e9-9eb4-b687fd7a9f5d.png)

I therefore suggest to additionally place four buttons between the end of the page and the comments section:

- Previous chapter (jumps to first page thereof)
- Previous page
- Next page
- Next chapter (jumps to first page thereof)

Here’s a quick mockup of how that could look:

![image](https://user-images.githubusercontent.com/3104762/64079956-c58ac700-ccee-11e9-9160-41dafc1b3022.png)


**Further behaviour** 
- Buttons are ghosted if there is no previous/next chapter/page.
- Buttons could be hidden entirely if there is just a single page (UX consistency?). (Hide chapter buttons if there aren’t any chapters??)

**Discussion: Floating navigation?**
Another concept might be to have this navigation with its 4 buttons visible not only when reaching the end of the page, but instead visible and thus accessible at all times, so one can e.g. jump to the next page, even if one’s only half-way through a long page. The preferable place for such a floating navigation would be the bottom of the window IMO.

The problem I see in this approach however is the waste of screen real estate. A solution to this problem could be to make the navigation only appear (instantly) when a) the mouse pointer gets moved to the lower something pixels at the window or b) one scrolls to the very end of the screen.

ssddanbrown:
[...]

PR #2511, opened 2021-01-27
Add prev and next button to navigate through different pages

shubhamosmosys:
- Added feature to navigate through the different pages.
- This will help the user to navigate through different pages without scrolling above and clicking on the page.
- User can directly switch to the next page from the bottom of the page.
- This feature will be more helpful if the user using mobile/tab then, the user doesn't have to switch to the info tab from the content tab to view the next page.
bookstack1

- After the implementation of this feature user can view the next or previous page without switching to the info tab from the content tab.

bookstack2

Prev and Next button functionality can be seen in the attached video below:

https://user-images.githubusercontent.com/57482557/105948778-98804680-6091-11eb-9eff-893d7b3482e6.mp4

### DIFF
--- a/app/Entities/Models/Page.php
+++ b/app/Entities/Models/Page.php
@@ -126,4 +126,13 @@ class Page extends BookChild
         $refreshed->html = (new PageContent($refreshed))->render();
         return $refreshed;
     }
+    /**
+     * Get the parent chapter ID.
+     */
+    public function getParentChapter()
+    {
+        $chapterId = $this->chapter()->visible()
+        ->get('id');
+        return $chapterId;
+    }
 }
--- a/app/Entities/Repos/PageRepo.php
+++ b/app/Entities/Repos/PageRepo.php
@@ -467,4 +467,10 @@ class PageRepo
             ->where('page_id', '=', $page->id)
             ->orderBy('created_at', 'desc');
     }
+    /**
+     * Get page details by chapter ID.
+     */
+    public function getPageByChapterID(int $id){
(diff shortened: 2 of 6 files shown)


Wookbert:
I'm actually using both my own, simple implementation and @james-geiger ’s implementation from #1381 in parallel for quite some time now, and have come to the conclusion that previewing the page titles as he does, is really handy.

image

ssddanbrown:
Thanks for this @shubhamosmosys,
I have now reviewed this and made a series of changes in 0cfff6ab6f840e1e8892ff8d138880ae40f9de12, with some testing added in ed6ec341dfc1dff765911b5b283884c56ef747fe.

The design is now as shown below, This will appear on both page and chapter display pages.

Screenshot from 2021-05-29 12-54-19

ssddanbrown:
Also, Thanks @Wookbert for your input on the various implementations.

Wookbert:
@ssddanbrown Excellent! Any chance you could add an optional single pixel frame (as @james-geiger did) or shaded background, so it has more of a button character rather than just a text link. Me and my colleague both think it makes the feature more obvious and more likely to be used.

ssddanbrown:
@Wookbert I've purposefully try to keep these links rather muted by default as to not be distracting or result in a series of boxes at the bottom of the page. They will currently show an outline when moused over, which could be force enabled by default by a line of custom head css if desired.

Wookbert:
Ah, OK. Will inspect and if „necessary“ comment once released.

PR #2512, opened 2021-01-27
Add the option to use ldap:// with StartTLS

lahx:
ldaps:// seems to be depreacted and ldap:// with StartTLS is the preferred way.
In case a ldap server makes the use of StartTLS mandatory (-ZZ option on cli) php needs to execute ldap_start_tls() on the connection.


ssddanbrown:
Thanks for offering these changes @lahx.
This functionality has now been achieved as part of #2376 and will be part of the next feature release.

PR #2513, opened 2021-01-27
Fix French translations

Baptistou:


Baptistou:
I would like to highlight the last point in commit [f40c389](https://github.com/BookStackApp/BookStack/pull/2513/commits/f40c389a1586102521babd2241f20591b064ace9).
The button ```pages_edit_set_changelog``` causes a wrapping which is fixed with the new translation.
Is it an intended behaviour ? If not, this issue might appear in an other language.

Before
image

After
image

ssddanbrown:
Thanks @Baptistou, Will merge this in for the next feature release.

Issue #852, opened 2018-05-20
Add estimated read time


mendiromania:
Just a cool feature request, fit an estimated read time somewhere

TBK:
+1

Example:
![image](https://user-images.githubusercontent.com/858296/40860948-76b70f18-65e7-11e8-9ccb-f676f8cf5906.png)

Code example:
https://gist.github.com/mynameispj/3170442

aljawaid:
Not only cool but very useful guide too considering this is a book-type software which is designed for reading :)

guillaumehanotel:
Hi,
as mentioned in the PR #2514 , the lead dev isn't convinced of this feature. As he recommended it to me, I attach here a piece of code to put in the "Custom HTML Head Content" section in the settings. Feel free to adjust it at your own.

This will allow you to have an estimated time under the title of a page (cf this [image](https://imgur.com/zo1t0eR))

```javascript
<script type="text/javascript">

	function renderPageReadingTime() {
		const page = document.querySelector('.page-content');
		if(page !== null) {
			const pageTitle = page.getElementsByTagName('h1')[0];
			const pageContent = pageTitle.parentElement.innerText;
			const words = (pageContent.replace(/(<([^>]+)>)/gi, "")).split(' ').length;
			const readingTimeInMinute = Math.floor(words / 200);
			const readingTimeEl = document.createElement('p');
			readingTimeEl.appendChild(document.createTextNode(readingTimeInMinute + " min read"));
			pageTitle.after(readingTimeEl);
		}
	}
	
	setTimeout(renderPageReadingTime, 100);

</script>
```

ssddanbrown:
There's been relatively little demand for this since opening so I'm going to close it off.
If desired, upon the above JS solution, this could also likely be added using our [visual theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/visual-theme-system.md).

PR #2514, opened 2021-01-29
Add reading time to page display

guillaumehanotel:
Hello,

as referred in issue #852, i lay the first stone for this feature by adding a estimated read time below a page title.
Any advice is welcome to help me integrate this feature correctly in the project.

Thank you

### DIFF
--- a/app/Http/Controllers/PageController.php
+++ b/app/Http/Controllers/PageController.php
@@ -133,6 +133,10 @@ class PageController extends Controller
 
         $pageContent = (new PageContent($page));
         $page->html = $pageContent->render();
+
+        $words = str_word_count(strip_tags($page->html));
+        $readingTimeInMinute = floor($words / 200);
+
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Thanks for offering this PR @guillaumehanotel, But i think the value of such a feature would really need to be confirmed before inclusion as I'm personally not too convinced in the benefit of this. If BookStack was primarily for articles/blog-posts then maybe this would be ideal for all users otherwise this would need to be configurable visible, which is something I'd prefer to also avoid. Either way, the main page content is generally reserved for content only, no meta data.

I saw you originally had a JS based solution. What might be useful is if you tweak that so it can be added via the "Custom HTML Head Content" setting, then share that on the original issue, so that others can add it to their instance if desired.

guillaumehanotel:
Thank you for your feedback, 
I did as you told me and shared my piece of code in the issue.

Issue #1742, opened 2019-10-24
Sort Books within Shelves


sazora:
When I enter through the shelves area and then click on a topic within the shelves, it would be helpful if we could have the sort helper option like there are in other sections.

It would make it uniform across the platform.

![Image 2019-10-24 at 5 21 30 PM](https://user-images.githubusercontent.com/54324712/67526475-df0f1600-f682-11e9-83a6-163e29444d3c.png)

sunnythaper:
I would like to second this feature request!

judokan9:
I would also like to have this feature

pablfr:
+1 :)

sorinsfirlogea:
+1

ssddanbrown:
Thanks to @guillaumehanotel this functionality has now been merged and will be part of the next feature release so I'll therefore close off this request.

PR #2515, opened 2021-01-29
Sort Books within Shelves

guillaumehanotel:
Hello,

as referred in the issue #1742 , I've added the possibility to sort books within shelves.

Please give me your feedback

### DIFF
--- a/app/Http/Controllers/BookshelfController.php
+++ b/app/Http/Controllers/BookshelfController.php
@@ -101,6 +101,9 @@ class BookshelfController extends Controller
         $shelf = $this->bookshelfRepo->getBySlug($slug);
         $this->checkOwnablePermission('book-view', $shelf);
 
+        $sort = setting()->getForCurrentUser('books_sort', 'name');
+        $order = setting()->getForCurrentUser('books_sort_order', 'asc');
+
         Views::add($shelf);
         $this->entityContextManager->setShelfContext($shelf->id);
         $view = setting()->getForCurrentUser('bookshelf_view_type', config('app.views.books'));
@@ -109,7 +112,9 @@ class BookshelfController extends Controller
         return view('shelves.show', [
             'shelf' => $shelf,
             'view' => $view,
-            'activity' => Activity::entityActivity($shelf, 20, 1)
+            'activity' => Activity::entityActivity($shelf, 20, 1),
+            'order' => $order,
+            'sort' => $sort
         ]);
     }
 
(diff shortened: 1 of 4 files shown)


resources/views/shelves/show.blade.php
```diff
@@ -9,19 +9,30 @@
...
+                    'created_at' => trans('common.sort_created_at'),
+                    'updated_at' => trans('common.sort_updated_at'),
+                ], 'order' => $order, 'sort' => $sort, 'type' => 'books'])
+            </div>
+        </div>
+
         <div class="book-content">
             <p class="text-muted">{!! nl2br(e($shelf->description)) !!}</p>
-            @if(count($shelf->visibleBooks) > 0)
+            @if(count($shelf->visibleBooksByCustomSorting($sort, $order)) > 0)
```
ssddanbrown, line 26: Since this call is ran twice it's going to re-query the database. This should probably be done once at the controller level and passed into the view as an extra value.

resources/views/shelves/show.blade.php
```diff
@@ -9,19 +9,30 @@
...
-        <h1 class="break-text">{{$shelf->name}}</h1>
+
+        <div class="grid half v-center">
+            <h1 class="break-text">{{$shelf->name}}</h1>
+            <div class="text-m-right my-m">
+                @include('partials.sort', ['options' => [
+                    'name' => trans('common.sort_name'),
+                    'created_at' => trans('common.sort_created_at'),
+                    'updated_at' => trans('common.sort_updated_at'),
+                ], 'order' => $order, 'sort' => $sort, 'type' => 'books'])
```
ssddanbrown, line 20: Shelves currently have user-defined ordering. There's no option here to make use of the existing, user-defined ordering.

app/Http/Controllers/BookshelfController.php
```diff
@@ -101,6 +101,9 @@ public function show(string $slug)
         $shelf = $this->bookshelfRepo->getBySlug($slug);
         $this->checkOwnablePermission('book-view', $shelf);
 
+        $sort = setting()->getForCurrentUser('books_sort', 'name');
+        $order = setting()->getForCurrentUser('books_sort_order', 'asc');
```
ssddanbrown, line 105: Within `show.blade.php` changes I've added a comment regarding user-defined ordering. Once that's added we won't be able to share the `books_sort` setting with the books listing since there would be differences in available ordering options. We'd probably need to add a `shelf_books_sort` option, which defaults to the current user-defined sort.

ssddanbrown (changes requested):
Thanks for offering this PR @guillaumehanotel.

I've added my thoughts during initial review of the changes. We'll also need to ensure these changes are covered with some level of testing.

gdrnclaeys:
Looking forward to this change

ssddanbrown:
Thanks for attempted some of those changes.

I've finished this off and added testing in 5c9c1d1a4bf71c04e802cbdb39f2a1dff3ff6515.

This is now merged, to be part of the next feature release.

PR #2521, opened 2021-01-31
Add footer block to place contact information or something else

johnroyer:
there are something could place in the footer:
- `Powered by BookStack`
- send email to `admin@bookstack` if vulnerability found
- Creative Commoms - Attribution-NonCommercial-ShareAlike 4.0 International


ssddanbrown:
Thanks for offering this @johnroyer but as of yesterday we merged in #1973/#854, Which I feel achieves the same goal really. I would not want two new options for footer content.

As an added note, If creating a PR you don't need to update all language files, Only English, Crowdin will automatically create all the language variants as required.

johnroyer:
It is a good news. Thanks !

PR #2522, opened 2021-01-31
Docker: Fix PHP tests 

Abijeet:
This creates another mysql_testing database during db service setup

Replace server with env tags in phpunit.xml in order to force
override certain parameters when tests are run. See:
sebastianbergmann/phpunit#2353 for more
information.

Rename primary developer Docker database from bookstack-test to
bookstack-dev. bookstack-test is used as the mysql_testing database

### DIFF
--- a/app/Config/database.php
+++ b/app/Config/database.php
@@ -81,10 +81,11 @@ return [
         'mysql_testing' => [
             'driver'    => 'mysql',
             'url' => env('TEST_DATABASE_URL'),
-            'host'      => '127.0.0.1',
+            'host'      => $mysql_host,
             'database'  => 'bookstack-test',
             'username'  => env('MYSQL_USER', 'bookstack-test'),
             'password'  => env('MYSQL_PASSWORD', 'bookstack-test'),
+            'port'      => $mysql_port,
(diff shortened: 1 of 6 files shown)


ssddanbrown (changes requested):
Thanks for this @Abijeet.

I really don't want to have to alter an app files just to suit the docker setup unless absolutely required, Especially those changes in `phpunit.xml` as I'm sure I'll  end up spending time in the future trying to understand the difference in variables or I'll get caught out when trying to override them in testing.

The `mysql_testing` database connection takes a `TEST_DATABASE_URL` env option, This should take priority over other options and takes the format of `mysql://user:password@host/table` so that could be used if needed.

With this, test database migration is done automatically during init but seeding is done manually via command. I think it'd be better, and consistent with the non-docker flow, if both of these were done manually. Auto migrating alongside the non-test DB could cause confusion if errors are thrown and will slow down start-up and increase the output.

ssddanbrown:
Thanks again @Abijeet.

I made the changes advised above within commit 43c2fc3c3780c758c154aca8b56e63080fa49cc4.

Issue #2491, opened 2021-01-15
Feature request: Pandoc integration


maggie0002:
Hi @ssddanbrown,

I was thinking Pandoc integration as an optional module. It would add some efficiencies to the various exports by keeping the assets seperate as discussed above (and potentially resolve some other outstanding issues), but also provide a bunch of additional options, such as EPUB (#1949), Word doc, video export support (#883; #2412) and a bunch more.

Here are a few shortcuts to try it out:

1. Here is Pandoc: https://pandoc.org
2. In most repositories so` apt-get install pandoc` or `brew install pandoc `should do the trick (if installing in a docker container, may need to install build-essential and/or curl).
3. An example Markdown I have tested with:

test.md
```
# Test file
Test MD File.

[![Build Status](https://cdn.vox-cdn.com/thumbor/zEZJzZFEXm23z-Iw9ESls2jYFYA=/89x0:1511x800/1600x900/cdn.vox-cdn.com/uploads/chorus_image/image/55717463/google_ai_photography_street_view_2.0.jpg)](https://travis-ci.org/joemccann/dillinger)
Dillinger is a cloud-enabled, mobile-ready, offline-storage, AngularJS powered HTML5 Markdown editor.

  - Type some Markdown
  - Convert some Markdown

![](https://www.learningcontainer.com/wp-content/uploads/2020/05/sample-mp4-file.mp4)

# New Features!

  - sdfsdf
  - sdfsdvldkvnc
 
You can also:
  - send
```

Execute the command:

`pandoc test.md -o example2.html --extract-media ./assets`

More info relating to this originally discussed in: https://github.com/BookStackApp/BookStack/issues/2412

maggie0002:
@ssddanbrown in response to the last comment over in https://github.com/BookStackApp/BookStack/issues/2412, indeed, these ostensibly simple things often get more complex very quickly. 

In terms of workflow, after giving it some thought perhaps a similar integration as WKHTMLTOPDF. The user installs Pandoc manually, using the Pandoc docs for their environment (apt-get Pandoc for example in Ubuntu). Then adds in a `PANDOC=True` variable to the .env file so that BookStack doesn't have any responsibility for the Pandoc install. 

When PANDOC=True there could be some new fields in the export dropdown menu: EPUB; HTML Archive (or something more logically named instead of HTML Archive. 

Hopefully then passing the same content being pulled for the current export features to Pandoc on the system locally, followed by a return of the output to download. 

[...]

Issue #2459, opened 2021-01-07
wkhtmltopdf generating wrong TOC links


dyzdyz010:
**Describe the bug**
I'm using [linuxserver/bookstack](https://hub.docker.com/r/linuxserver/bookstack) docker image. When I export the page to pdf with `wkhtmltopdf` and open it, click one of the `TOC` links, it doesn't take me to the correct location and showing meaningless link address, like `file:///tmp/knp_snappy5ff65ed5491ae7.13561238.html`.

**Steps To Reproduce**
1. Create a page with some headers
2. Configure BookStack to use `wkhtmltopdf` for pdf-generating
3. Export and open
4. Click one of the TOC links

**Expected behavior**
Clicking a TOC title link should take me to the corresponding content position.

**Screenshots**

![2021-01-07 09 44 24](https://user-images.githubusercontent.com/1324420/103840704-078b0080-50cd-11eb-83d5-e919fa13c574.jpg)

![2021-01-07 09 44 29](https://user-images.githubusercontent.com/1324420/103840713-0d80e180-50cd-11eb-9bb4-0d9a9bd18a2c.jpg)

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): v0.30.7
 - PHP Version: 7.3.25
 - Hosting Method (Nginx/Apache/Docker): Docker

**Additional context**
Also no bookmarks for TOC.

maggie0002:
I am experiencing the same issue. Also using the linuxserver image. I noted an issue open over there, but appears to be for an unrelated point but wanted to flag it: https://github.com/linuxserver/docker-bookstack/issues/76

maggie0002:
I tried this in the development environment today, and didn't have any issues. I then tried it in the Solidnerd Docker container, and did have the issue.

Process of elimination has narrowed this down to the version of wkhtmltopdf running:

Development environment: wkhtmltopdf 0.12.6 (with patched qt)
Linuxserver: wkhtmltopdf 0.12.5
Solidnerd: wkhtmltopdf 0.12.5

1. So I updated the Linuxserver image to the 0.12.6 (actually I bumped the Alpine container version to .13, which then pointed it to the new repository which had the 0.12.6 in it). But the issue still occurred. 

2. Then I pulled the latest build directly from the website (https://wkhtmltopdf.org/downloads.html) into the Solidnerd container, and the issue was resolved. 

[...]

PR #2524, opened 2021-02-01
Add Pandoc to API

maggie0002:
This is a proposed endpoint to resolve: https://github.com/BookStackApp/BookStack/issues/2491

As can be seen from the conversation in that thread, adding more complex means of exporting HTML (i.e. via Pandoc or other third party means) may be complex to integrate and to maintain. 

Instead, this API endpoint provides the HTML file with its original HREFs that point back to the server hosting BookStack. It provides a much lighter weight file due to images not being encoded into the file, and provides users more options to build on BookStack with their own third party solutions to processing exported content than can be achieved with contained HTML.

My personal use case, as an example, will be to then process this output via Pandoc or WGET to generate an offline version which I have more control over the formats/content/sizes. It will also allow me greater control over extracting embedded videos. 

Note: This is for discussion over whether it would be merged. The following further commits are required at a minimum before ready:

- [x] Integrate the new endpoint in to the API docs (appears to auto generate them, haven't identified how yet).
- [ ] Consider the endpoint names. My suggestion would be to rename the current exportHtml (which is the original contained HTML endpoint) to `exportContainedHtml` and then have this new endpoint as `exportHtml`.

Would be helpful hear whether this is within the scope of the project before putting in these other changes.  

I would welcome edits from maintainers directly into the repo if you so desire.

### DIFF
--- a/app/Entities/Tools/ExportFormatter.php
+++ b/app/Entities/Tools/ExportFormatter.php
@@ -71,6 +71,54 @@ class ExportFormatter
     }
 
     /**
+     * Convert a page to a plain HTML file with original HREFs.
+     * Includes required CSS.
+     * @throws Throwable
+     */
+    public function pageToHtml(Page $page)
+    {
+        $page->html = (new PageContent($page))->render();
+        $pageHtml = view('pages.export', [
+            'page' => $page,
+            'format' => 'html',
+        ])->render();
+        return $pageHtml;
+    }
+
+    /**
+     * Convert a chapter to a plain HTML file with original HREFs.
+     * @throws Throwable
+     */
+    public function chapterToHtml(Chapter $chapter)
+    {
+        $pages = $chapter->getVisiblePages();
+        $pages->each(function ($page) {
+            $page->html = (new PageContent($page))->render();
+        });
+        $html = view('chapters.export', [
+            'chapter' => $chapter,
+            'pages' => $pages,
+            'format' => 'html',
+        ])->render();
+        return $html;
+    }
+
+    /**
+     * Convert a book to a plain HTML file with original HREFs.
+     * @throws Throwable
+     */
+    public function bookToHtml(Book $book)
+    {
+        $bookTree = (new BookContents($book))->getTree(false, true);
+        $html = view('books.export', [
+            'book' => $book,
+            'bookChildren' => $bookTree,
+            'format' => 'html',
+        ])->render();
+        return $html;
+    }
+
+    /**
      * Convert a page to a PDF file.
      * @throws Throwable
      */
--- a/app/Http/Controllers/Api/BookExportApiController.php
+++ b/app/Http/Controllers/Api/BookExportApiController.php
@@ -36,6 +36,17 @@ class BookExportApiController extends ApiController
     }
 
     /**
+     * Export a book as a plain HTML file with original HREFs.
+     * @throws Throwable
+     */
+    public function exportPlainHtml(int $id)
+    {
+        $book = Book::visible()->findOrFail($id);
+        $htmlContent = $this->exportFormatter->bookToHtml($book);
+        return $this->downloadResponse($htmlContent, $book->slug . '.html');
+    }
+
+    /**
      * Export a book as a plain text file.
      */
     public function exportPlainText(int $id)
--- a/app/Http/Controllers/Api/ChapterExportApiController.php
+++ b/app/Http/Controllers/Api/ChapterExportApiController.php
@@ -40,6 +40,17 @@ class ChapterExportApiController extends ApiController
     }
 
     /**
+     * Export a chapter as a plain HTML file with original HREFs.
+     * @throws Throwable
+     */
+    public function exportPlainHtml(int $id)
+    {
+        $chapter = Chapter::visible()->findOrFail($id);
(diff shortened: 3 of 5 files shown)


ssddanbrown:
Thank you for offering this @maggie0002. That said, I don't think I'd want to include this. This would be a slippery slope in my view, of creating new endpoints for somewhat subtle changes in export requirements. Nor would I want a mass of options to build up on a single endpoint.

Additionally, this content already provided in the page endpoint is the plain content HTML; If anything, this adds export specific content. If you have a real need for this I could try to find some time to collate the content used in an export to get you pretty much the same using the normal page GET API? Is a case of going through the export views and gathering the required styles and HTML used.

maggie0002:
Hi @ssddanbrown,

As you can probably tell, I'm maxing out on what I can do here, and it is indeed something I have a real need for. Any help would be hugely appreciated. 

The API script you shared (https://github.com/BookStackApp/api-scripts/blob/19f9a5bcf215fa84010c72dc99bff5c199120bf8/php-book-to-static-site/book-to-static.php#L238) does look hopeful, but I would need to be able to export all three, pages, chapters and books like the API endpoint I tried to add here. I would also need to be able to use it without having to provide the tokens too, i.e. if the user is already logged in then it shouldn't require them. That is something the additional endpoint would allow, but I'm not sure how I would pass on the tokens through a separate PHP page, which ideally would be hosted on a server users could call.

Ultimately, the workflow will be:

1. User is shown a lightweight minimalist interface that displays all their available books, chapters and pages that they can download (this interface will be compiled from the API) (this is part of a vue.js web interface).
2. When user clicks the link to download an item, it will download the book or chapter or page as HTML with all the images and videos included with it so it can be used offline (separate rather than embedded images for space saving). 

The idea of having an API endpoint that displays the HTML with original HREFs is it would be easy enough to then create the HTML with separate videos and images from it.

ssddanbrown:
@maggie0002 Since it sounded like you were doing things front-end I done a quick example in JavaScript:
https://gist.github.com/ssddanbrown/6e19f39684696ea01002e4d155c0ffd2

Should be able to just paste this into the browser console to trial it (As long as you're logged into BookStack); No token-auth as requested but you may need to consider CORS if your JS app is on another domain.
It's most of the content that makes up the export layout, apart from the metadata footer. I've trimmed the styles a bit but they could probably be trimmed further with more time.

maggie0002:
Thanks @ssddanbrown for the time putting this together, really appreciate it. 

The CORS thing is a bit of an issue having to allow it on the server for the endpoint, but understandable. I think the bigger challenge is it involves so much code to maintain. Would need to add on to that more code to cycle through all the pages in a chapter or book to create them and format the HTML accordingly, and then still build on top the process that takes that HTML and creates a package of all the content for offline use. Especially when most of what is needed is already in Bookstack and can be tapped into directly without too significant a change to the endpoints rather than placing parts in-between to redo the same task.

That's not your problem of course, I pitched the pull request suggesting the endpoint to provide a means to tap in directly to the content in more malleable forms and was a swing and a miss. I'm wondering if you could take a look at this pull request though and point me in the direction of how to resolve the issue with the API docs not generating since the changes I apply? I will have to give the issue some more thought, but may end up forking the BookStack repo and adding in these changes, would be good to know how to get it working.

ssddanbrown:
@maggie0002 Docs work fine for me with these changes.

Try adding this to your `.env` file:

```bash
APP_ENV=development
```

The docs are cached in production mode (Using the app version number within the key) so it doesn't have to re-scan them from source each time ([can see this done here](https://github.com/BookStackApp/BookStack/blob/master/app/Api/ApiDocsGenerator.php#L27)).

maggie0002:
@ssddanbrown, thanks for the pointer. Indeed somewhere along the way I had switched back to prod to early. The docs do appear to work ok.

I am going to have a look at integrating Pandoc into the endpoint in this pull request at some point. If you're interested in taking a look either for interest, future reference etc. then I will keep adding here. Otherwise feel free to close the pull request. 

Thanks again

maggie0002:
This hasn’t been forgotten, just a bit backed up. 

After some research it seems like a Pandoc wrapper would be the way to go: https://github.com/ueberdosis/pandoc

It would align with the approach used in the PDF export which is nice, and is PhP so no need for hacking away. Been meaning to have a go at experimenting with it, but wanted to check in and see if this is something you think may be viable for merging as a Pandoc integration? If so would be good to get input as it would entail designing an endpoint that can receive the options through the URL, or to have just a few export options available hardcoded in, such as a PDF and HTML option. I go back and forth on what may be best.

maggie0002:
Hoping to pick this up again this weekend. Any thoughts on the API endpoints?

That said, having given it some more thought, how would you feel about this as an optional exporter similar to how wkhtmltopdf is currently integrated? The wrapper is proving effective: https://github.com/ueberdosis/pandoc

Would also help resolve some other issues that I don't think we will find a way around:

https://github.com/linuxserver/docker-bookstack/issues/80
https://github.com/BookStackApp/BookStack/issues/2459

maggie0002:
This pull request has become a mix of different buts of code and objective. Going to close this pull request and move feature request to: https://github.com/BookStackApp/BookStack/issues/2491

PR #2530, opened 2021-02-03
Update TrashCan.php

i4j5:
bookshelf


ssddanbrown:
Thanks for the fix @i4j5.

This was cherry-picked into the v0.31.x branch and will be soon published as part of the next patch release.

PR #2533, opened 2021-02-03
Fix German translation string

benediktvolke:


ssddanbrown:
Thanks @benediktvolke,
Will merge this in for the next feature release.

Issue #2525, opened 2021-02-01
/user/{id} shouldn't be sequential  


maggie0002:
**Describe the bug**
Whenever a user is created they are assigned a UID of `last UID +1`. I.e, Admin user is user 1, the next user created is 2, the next 3 and so on and so forth. 

This means that you can query the database and build an entire replica of every user (including profile pictures) by simply visiting url.com/bookstack/user/1, url.com/bookstack/user/2 and so forth. There isn't any email addresses or login details exposed, but I would suggest this isn't particularly good practice exposing so much data in this way.

**Expected behavior**
Each user should be provided a non-sequential UID, a random generated number of reasonable length.

ssddanbrown:
Thanks for raising @maggie0002,

I agree, I can see that potentially being a probably in non-private instances. Will mark for next feature release.

ssddanbrown:
Extra thought: This would also apply to search filters.

ssddanbrown:
Now added into master to be part of the next feature release.

I've used a unique slug version of the user's name, since that keeps it non-sequential, contextual and uses information that will be known when clicking through on that link (Does not expose anything new).

It does mean that link could possibly break old references on user name change but I imagine it's relatively uncommon that a BookStack profile will be linked to in an important manner. 

Thanks again for raising @maggie0002

kirillnad:
The same with pages nums!
It would be greate each page has a long random number.
Then, it would be possible share page by it's link without permission issues

PR #2561, opened 2021-02-14
Search owned by me

### LINKED ISSUE #2529: Search books owned by me
Thank you for the great feature #2436 !
Now that we have a owner field, it would be a very nice feature to better identify books that we own.
For now, I found only two options available in the search page which are "Created by me" and "Updated by me".
I was thinking about 3 possible ideas :
- Add an option "Owned by me" in the search page
- Display the owner field in the list/grid view next or under book title.
- Set an automatic book tag "own" that we could search, or even better that could be displayed #2528 

Better identify books that we own.

benediktvolke:
This adds the feature to search entities by owner to resolve #2529.

image

### DIFF
--- a/app/Entities/Tools/SearchRunner.php
+++ b/app/Entities/Tools/SearchRunner.php
@@ -290,6 +290,17 @@ class SearchRunner
         $query->where('updated_by', '=', $input);
     }
 
+    protected function filterOwnedBy(EloquentBuilder $query, Entity $model, $input)
+    {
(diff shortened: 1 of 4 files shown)


ssddanbrown:
Thanks @benediktvolke!

---

_Just a reminder to myself to consider #2525 when reviewing this._

ssddanbrown:
Thanks again @benediktvolke.
Now merged for the next feature release.

PR #2581, opened 2021-02-23
Add socialite provider for gitea

moqmar:
http://gitea.io/ is a self-hosted Git service like GitLab, but it uses slightly different OAuth2 endpoints. This PR makes it possible for us at Codeberg.org (which is the reason why I have created this PR) as well as many other people, organizations and companies to use BookStack together with their Gitea instances.


ssddanbrown:
Thanks for offering this @moqmar, But I've come to the decision I don't want to grow out the supported third-party social provides much further, at least without a high level of benefit/request from the user-base. It's not a large amount of work to initially implement but it gets troublesome attempting to support people here when issues eventually arise.

PR #2586, opened 2021-02-24
Fixing default config lookup for bookshelves_view_type in BookshelfController

philjak:
A wrong setting key (`bookshelf_view_type`) was used. So the configured default in `.env` was not used. This PR fixes this issue.

How to reproduce:

- Set `APP_VIEWS_BOOKSHELVES=list` in `.env` File
- Use a new user and open a random bookshelf (/shelves/{slug})

Expected: See books as list.

What actually happens: Default grid view is shown.

### DIFF
--- a/app/Http/Controllers/BookshelfController.php
+++ b/app/Http/Controllers/BookshelfController.php
@@ -103,7 +103,7 @@ class BookshelfController extends Controller
 
         Views::add($shelf);
         $this->entityContextManager->setShelfContext($shelf->id);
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering a PR @philjak but this was not a bug or incorrect value used; Those are two different views with two different preference keys. There's currently no `.env` option for the default shelf view option.

philjak:
@ssddanbrown Ahh, ok. Sorry. I just recognized that the default behavior changed during the last updates

PR #2607, opened 2021-03-03
Implementation of Nextcloud OAuth authentication

artur-shaik:
As https://github.com/SocialiteProviders/Providers/pull/632 is already merged, this implementation use it for authentication with Nextcloud.

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -187,6 +187,13 @@ TWITTER_APP_SECRET=false
 TWITTER_AUTO_REGISTER=false
 TWITTER_AUTO_CONFIRM_EMAIL=false
 
+# You must include index.php in NEXTCLOUD_BASE_URI if pretty URL is not configured.
+NEXTCLOUD_BASE_URI=false
+NEXTCLOUD_APP_ID=false
+NEXTCLOUD_APP_SECRET=false
+NEXTCLOUD_AUTO_REGISTER=false
+NEXTCLOUD_AUTO_CONFIRM_EMAIL=false
+
 # LDAP authentication configuration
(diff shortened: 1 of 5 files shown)


artur-shaik:
Looks like repo needs `composer update` command to make it work?

Tealk:
@artur-shaik Is this done via oAuth2?

artur-shaik:
> @artur-shaik Is this done via oAuth2?

Exactly.

Tealk:
> Exactly.

Why don't you name it oAuth2, then it would be more general and users would know that they can include other services as well.
I for example would use it for MailCow ;)

artur-shaik:
I wish to have generic OAuth provider in this app, too. But authentication implemented with third party library for php called Socialite. So I decide to implement Nextcloud support in that library to easily integrate it in Bookstack.

I think, you can try to substitude your app credentials and see what happen, with this patch.

Tealk:
I will definitely try that out, then all services I have would be accessible via one instance.

artur-shaik:
> I will definitely try that out, then all services I have would be accessible via one instance.

I just realized that this will not work. There are field access values that are unique for Nextcloud.

So, my advise for you is to implement your provider as Socialized plugin and use it on your BookStack instance as I do on my own. Actually you can just copypaste base classes and change it regarding you provider.

ssddanbrown:
Thanks for offering this PR @artur-shaik, But I've come to the decision I don't want to grow out the supported third-party social provides much further, at least without a high level of benefit/request from the user-base. It's not a large amount of work to initially implement but it gets troublesome attempting to support people here when issues eventually arise.

PR #2616, opened 2021-03-06
Fixed german formal translation

geins:
-fixed mismatch of informal language in formal german file
-fixed 4 residues of english in formal german file


ssddanbrown:
Thanks @geins, Have merged in for the next feature release.

Issue #2631, opened 2021-03-11
Pages not rendering base64 blob images


awarre:
**Describe the bug**
Pages not rendering images after pasting images into WYSIWYG editor when original source is a base64 image blob.

**Steps To Reproduce**
1. Create a New Page
2. Copy data from a web site including an image that is a base64 data blob. (I used the pixel graphic from http://www.techerator.com/2011/12/how-to-embed-images-directly-into-your-html/ for this test)
3. Paste the data into the WYSIWYG editor
4. The text and image will appear on the Editing page.
5. Save the page
6. The image will not appear on the page. The image tag is completely stripped.
7. Edit the page again.
8. The image again renders on the Editing page, and was preserved during the save.

**Expected Behavior**
The image should render both on the editor, and on the actual page.

**More info**
This is very close to working. The image does get uploaded, and can be accessed directly and viewed through the editor. The path generated is in the format of: `blob:https://example.com/aabb1234-1234-aabb-bbcc-ffffeeeeffff`. As stated above, this is the expected uploaded image and can be accessed directly.

It appears the page does not generate the <img> element at all. If it did, everything should work as expected.

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): 0.31.7
 - PHP Version: 7.3.19
 - Hosting Method (Nginx/Apache/Docker): IIS

ssddanbrown:
Thanks for reporting @awarre with the provided detail. 

Just tried this in the WYSIWYG and markdown editors on my local instance using the same image and in both instances the image data uploaded into BookStack as expected. On Firefox/Fedora.

Could you confirm the browser in use when this occurs?

awarre:
Using the latest build of Chrome on Windows 10. I just tested with Edge and see the same (though that's not much of a difference).

I can replicate this issue on the BookStack Demo site as well. It doesn't *look* like a browser rendering issue, because the `<img>` element is entirely missing from the page.

`<p id="bkmrk-cool%2C-that%E2%80%99s-eight-b">cool, that’s eight-bit me! And no linked image required!</p>
--
  | <p id="bkmrk-"></p>
  | <p id="bkmrk-if-you%E2%80%99re-interested">If you’re interested in trying this on your own without a pre-made base64 conversion tool, pay special attention to</p>
`

Where the empty <p> tags, there should be the image.

awarre:
The Editor page shows this code:

```

<textarea id="html-editor"  name="html" rows="5"
--
[...]

PR #2647, opened 2021-03-19
Allow Pages to render base64 blob images

awarre:
Modify badIframe xpath query to allow rendering data:image blobs.

#2631


ssddanbrown:
Closing as per https://github.com/BookStackApp/BookStack/issues/2631#issuecomment-803165054

PR #2656, opened 2021-03-26
Consolidate .env.example and .env.example.complete

setpill:
No more crossreferencing files and crawling through git history to see which of the conflicting duplicate documentation is the most recent and should be taken as canonical.

Options that were in `.env.example.complete` but not in `.env.example` were commented out, so you retain the benefits of the slimmed down `.env.example` (you can skip over any commented out options).


ssddanbrown:
Thanks for offering the PR but this is not a change I'd look to include since I like having clear distinctions of the required, and complete set of options, with the default `.env` used by users being a simplified file.

> No more crossreferencing files and crawling through git history to see which of the conflicting duplicate documentation is the most recent and should be taken as canonical.

`.env.complete` is designed to be a reference, so just use that instead. `.env.example` should rarely change in a significant manner and, if it does in a manner you'd need to be aware of, it'd be detailed on the [updates page](https://www.bookstackapp.com/docs/admin/updates/).

setpill:
Makes sense, I just did the consolidation for myself/my own learning process and figured I'd offer it up as a contribution :)

In that case you might want to look e.g. at the documentation of the `MAIL_DRIVER` option - the `mail` option was removed from `.env.example` but not from `.env.example.complete`. Seemed to me that `.env.example` is leading there, but here you state the opposite.

ssddanbrown:
@setpill Thanks for notifying on that, Have opened #2657 as a reminder to update those bits.

Issue #2594, opened 2021-02-26
Recycle Bin - parent information request


gdrnclaeys:
Currently the recycle bin only shows us the "Deleted Item name", "Deleted By" and "Deletion Time". When a single page has been deleted and the parent chapter or books also gets deleted a short time after, there is currently no way to see the parent information of the page item that is deleted in the first place.
So when you try to restore the single page you currently get the message "The parent of this item has also been deleted. These will remain deleted until the parent is also restored." So not knowing what the parent is makes it from a users point of view impossible to restore this page. If we had visibility of the parent information on a per item we would be able to restore the parent and subsequently the page.

arjvand:
There is 2 options :


1. Including parent name on Recycle Bin :
![screen1](https://user-images.githubusercontent.com/6619045/114594818-eecd5080-9ca2-11eb-9b33-88c575632b2f.jpg)

2. Including restore parent button on restore confirm page (in order to direct user to the parent restore confirm page):
![screen2](https://user-images.githubusercontent.com/6619045/114595906-2ab4e580-9ca4-11eb-8048-8d4985ec4d55.jpg)

ssddanbrown:
Thanks for the the request @gdrnclaeys.

As per #2682, code has now been merged to show the immediate parent item in the recycle bin list view, in addition to there being an option to go to the restore page for the deleted parent item upon this scenario occuring.

These additions will be part of the next feature release.

PR #2682, opened 2021-04-17
Including parent information in recycle bin

arjvand:
According to issue #2594 when a child item has been deleted and the parent item also gets deleted, there is no way to see the parent information in order to recover child item. This PR includes parent information in recycle bin to help user recover child elements.

screen1


ssddanbrown:
Thank you very much for offering this @arjvand.

I don't think this alone would have solved all scenarios of the original request, since you'd still have limited visibility on pages, within chapters, where the parent book has been deleted (Since it's an extra hop away) so I've followed this up with 3a402f6adc78fa939d898369e425088fed141544 to support your addition to provide the user an immediate jump to the actual deletion upon restore. 

All now merged, This will be part of the next feature release.

Issue #566, opened 2017-10-20
Diagram/graph feature on wiki via markdown syntax


tuedang:
### For Feature Requests

Desired Feature:
It would be nice if we can support to transform chart/graph from markdown, it's perfect for developer in case of drawing diagram for visualize the things.

##### Expected Behavior
I have tried to play around on BookStack, and it somehow will be like this

![markdown-graph](https://user-images.githubusercontent.com/7925077/31802764-26c463ca-b57a-11e7-93bf-eb4cfe1fcae6.png)

Technically, it we worry about performance, it should be applied lazyloading to load the graph resource.

andunix:
This could be done by integrating [mermaid](https://mermaid-js.github.io/mermaid/#/). But I don't know, if it is only client side (rendering in browser).

henri9813:
This feature would be awesome, we need this functionnality, currently, we draw network scheme on draw.io and after save it on the bookstack, but it's difficult to maintain ...

cdrfun:
Mermaid syntax is supported via diagrams.net https://www.diagrams.net/blog/mermaid-diagrams

ravensorb:
This would make it a lot easier to work with Bookstack.  Would love to see this implemented -- maybe integration with something like https://kroki.io/ could help simplify (and add support for a number of different visualizations)?

recalcitrant-matter:
Hello, we currently have implemented a simple work-around that allows us to render mermaid diagrams client side using the mermaid syntax.
![image](https://github.com/BookStackApp/BookStack/assets/33179984/76c8c415-c96b-4d4b-aa1f-a4d006728635)


To get this working we imported the mermaid rendering module using the **Custom HTML Head Content**  option under customization according to the documentation https://mermaid.js.org/config/usage.html#using-mermaid.

![image](https://github.com/BookStackApp/BookStack/assets/33179984/c7eebb02-dd6b-42c9-a1b7-811dbfa74b37)

Now just put your mermaid syntax in a `<pre class='mermaid>` element like this:
```
<pre class="mermaid">
flowchart TD
    A[Christmas] -->|Get money| B(Go shopping)
    B --> C{Let me think}
    C -->|One| D[Laptop]
    C -->|Two| E[iPhone]
    C -->|Three| F[fa:fa-car Car]
</pre>
```
and it should get rendered.

[...]

Issue #495, opened 2017-09-05
[Feature Request] Math & UML


Zottelchen:
Desired Feature: Math & UML

I would really like a simple way to write down Math and UML. There are enough libaries which support Markdown and WYSIWYG (at least Math, not sure about UML). Also if Math is going to be included please don't forget an easy way to paste special symbols.

noxify:
For the UML part, do you mean something like this?

* https://github.com/plantuml/plantuml / http://plantuml.com/tinymce
* https://github.com/knsv/mermaid

Zottelchen:
Yes, exactly. Both seem to give enough options to explain a process visually.

TBK:
For the math part of your feature request would https://www.mathjax.org/ suffice?

noxify:
For Mathjax, here the MD / TinyMCE implementation:

* https://hiltmon.com/blog/2017/01/28/mathjax-in-markdown/
* https://github.com/SoftwareMansion/tinymceMathjax

DagW:
+1 for Mermaid!

TBK:
@ssddanbrown could this be a v0.27.0 target? 🙏

JHenneberg:
I also would appreciate this very much :)

nicobbg:
Mermaid is available in Gitlab and is easy to use. It would be great to have this feature in Bookstack as well.

ltDan79:
+1 :)

brutecode:
+1 mermaid looks super dope

ThomasF34:
+1 for Mermaid features :)

eceforge:
For the math side of this might I recommend KaTeX [https://katex.org/](https://katex.org/) instead of mathjax -- it's significantly faster and has no dependencies. I've had partial success in fact simply adding the required javascript include to the customer header section of bookstack. The problem comes when attempting to bind the auto-rendering to a javascript event as it seems to break the image insertion as described in #1588

Lerke:
Another +1 from me. Having embedded UML support via Mermaid or PlantUML instead of having to render it locally and upload would be a great feature!

gkostkowski:
+1 for Mathjax or KaTeX

Zeigren:
> Another +1 from me. Having embedded UML support via Mermaid or PlantUML instead of having to render it locally and upload would be a great feature!

The included draw.io has support for [PlantUML](https://about.draw.io/use-plantuml-in-draw-io/)

[...]

PR #2714, opened 2021-04-27
Added Mermaid to MarkdownIt editor

cadavre:
This is really a simple PR with addition of [Mermaid](https://mermaid-js.github.io/) into MarkdownIt editor, that BookStack is using.

It allows to create Mermaid diagrams via Markdown with specific syntax.

This is a preview of how it works:

Zrzut ekranu 2021-04-27 o 14 19 00

This addresses #495 and #566 .

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,4619 @@
 {
+  "name": "app",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
+  "packages": {
+    "": {
+      "dependencies": {
+        "@liradb2000/markdown-it-mermaid": "^0.4.2",
+        "clipboard": "^2.0.8",
+        "codemirror": "^5.60.0",
+        "dropzone": "^5.9.2",
+        "markdown-it": "^11.0.1",
+        "markdown-it-task-lists": "^2.1.1",
+        "sortablejs": "^1.13.0"
+      },
+      "devDependencies": {
+        "chokidar-cli": "^2.1.0",
+        "esbuild": "0.7.8",
+        "livereload": "^0.9.3",
+        "npm-run-all": "^4.1.5",
+        "punycode": "^2.1.1",
+        "sass": "^1.32.8"
+      }
+    },
+    "node_modules/@braintree/sanitize-url": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/@braintree/sanitize-url/-/sanitize-url-3.1.0.tgz",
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Thanks for offering this PR @cadavre but I'd really need my [concerns listed here](https://github.com/BookStackApp/BookStack/issues/495#issuecomment-628919106) addressed before considering support.

In addition, I'm fairly sure that on the latest version of BookStack this will only show the mermaid output within the preview window, not within the the actual page output after save, unless I'm missing something?

cadavre:
Hah, I'm newbie to BookStack dev and looks like I went kind of too excited after Mermaid started to work in preview. Haven't thought that you're maintaining two different Markdown rendering engines. :o That seems like a pain in the ass. :/

> The current use-case for UML/Mermaid is weaker since we now have draw.io which has a lot of overlap and is well-suited to the use-case of multi-ability environments. I can see why those technically inclined may like UML/Mermaid, kind of want to use it myself tbh, but it's focusing on a narrow audience with a custom text syntax, which kind of opposes the idea/spirit of the project.

Markdown itself sounds like "focusing on a narrow audience with a custom text syntax" in comparison to WYSiWYG. I can see nothing wrong in adding Mermaid which can be accessed only via MD.

I believe we could add Mermaid rendering same as it is done for draw.io – graphs can be exported to .png for browsing mode. If I follow up with this kind of implementation, would it be ok for you to merge it?

ssddanbrown:
> Markdown itself sounds like "focusing on a narrow audience with a custom text syntax" in comparison to WYSiWYG. I can see nothing wrong in adding Mermaid which can be "accessed" only via MD.

But now we're providing a narrow feature for an already narrowed audience, and this works against the alignment I'm intended to work towards with the editor to get to a place where you switch between modes easier.

> I believe we could add Mermaid rendering same as it is done for draw.io – graphs can be exported to .png for browsing mode. If I follow up with this kind of implementation, would it be ok for you to merge it?

I'd imagine people would still want to edit via mermaid text, This is going to introduce some significant complications to achieve.  In addition it introduces some further complications if accepted in the editor markdown since you now have mis-alignment with the API unless you're able to do things back-end.

To be honest, I can't imagine an implementation I'd be happy to merge at this time, especially as we look to change up the editors later this year and align the editing modes. Will mean wasted effort if done now. I'd rather look to see what's required, after the editor work, to allow people to optionally add "Mermaid" via our customization options (Accepting the limitations/caveats). 

I'll close off this PR for now though.

mwatermolen:
Any chance we can re-evaluate this?

PR #2723, opened 2021-04-30
Improvement of the search user experiance

mo7zayed:
- Added words highlighting for the matched words in the search results﻿


ssddanbrown:
Due to the above I'll close this but feel free to re-open a fresh PR if one can be put together that addresses the concerns above.

Issue #2701, opened 2021-04-21
User management API so you are able to use user provisioning for automatic user updates


Mynster9361:
#### API Endpoint or Feature

API to access edit and maintain users to be able to use user provisioning and administer it locally from example Azure or other external services that can handle authentication to bookstack through user management API with provisioning.

I am almost certain that Azure is not the only service that provides this feature i could imagine something like Okta or other SSO providers also support this.

#### Use-Case

To be able to integrate bookstack even further with bookstack.
Currently you can bind groups from bookstack to groups in Azure and make them sync that way and remove users from groups.
But what happens when a user leaves and you delete them from Azure.
The user will still remain in Bookstack until you delete it manually and move the ownership.
(Maybe the content ownership should/could be moved to a default group instead)

#### Additional Context

I have found some more information on what it does and how user provisioning works for Azure since that is what i use with my bookstack instance
https://docs.microsoft.com/en-us/azure/active-directory/app-provisioning/user-provisioning
https://docs.microsoft.com/en-us/azure/active-directory/app-provisioning/how-provisioning-works

ssddanbrown:
The work in #3238 has now been merged which adds a CRUD user actions to our REST API which allows many task to be automated. Therefore I will close this off.

This issue hinted as SCIM but I didn't know if SCIM was the fundamental requirement of this request. The new API endpoints should allow some level of "SCIM adapter" to be created externally if desired. 

I wouldn't really want to implement SCIM directly into BookStack at this stage. I've reached the limit of authentication mechanisms I'm willing to support right now since they are difficult to fully understand and tricky to support for the user-base. I don't want to be spending any more of my own time:

- Reading through multiple RFCs
- Signing up to auth system trial accounts
- Trying to understand where Microsoft have made things awkward and where differences may lie between cloud and on-prem functionality.
- Supporting once implemented and attempting to understand what configuration is needed vs what is actually just some strange expectations due to an arcane environment. 

Simply put, it's not fun work, and it provides no benefit to the vast majority of users while having a disproportionate negative affect on maintenance and my mental well-being.  I'd be happy to reconsider if an enterprise account was willing to work with me while also willing to invest :dollar:  to make it worth my time. 

To pre-emptively answer the question "Would you be willing to accept a PR for SCIM"? No, as it does not solve the above. I would still have to understand the standards & requirements while having the responsibility of stability & maintenance. A PR would only really cover the most fun part, writing the code.

[...]

PR #2734, opened 2021-05-06
DRAFT: API endpoint for user management

itbane:
Hi,

I've found the need to get users (and their email) from the API for automation purposes. As there isn't an endpoint for user management yet, I thought I'd give it a shot and tried to implement it.

I'd like to get some feedback on whether I'm on the right track and should continue working on this.

**Disclaimer**: I haven't worked with Laravel ever before and my last contact with php was ~10 years ago. I'm grateful for any advice and pointers you can give me to improve the PR.

Tests and `CREATE`, `UPDATE`, `DELETE` operations are still WIP. I've seen someone request SCIM-Support (#2701) - I'd have to look into that.

**Current state:**
* GET: `/api/users` - Route (only accessible with `users-manage` permissions); lists all users + their roles
* GET: `/api/users/{id}` - Route (only accessible with `users-manage` permissions); lists details of a specific user

**Quick overview on my changes / thought process:**
- Add new routes for `users`
- Add new Controller `UserApiController`
- Use as much from existing `User` and `UserRepo`-classes as possible
  - obstacle: `apiListingResponse` requires a `Builder`-object, `User::getAll()` returns a collection
    - If I understood correctly, the collection contains all the data already, while the builder only contains the databasequery
    - solution: Implemented a new Function within `UserRepo` that returns a `Builder`-object
  - obstacle: some fields in the `User`-object are hidden, but need to be displayed, such as `email`
    - solution: added a new optional parameter to `apiListingResponse` that takes an array of fields that should be returned anyways (using `data->makeVisible()` in `ListingResponseBuilder::to_response`)

### DIFF
--- a/app/Api/ListingResponseBuilder.php
+++ b/app/Api/ListingResponseBuilder.php
@@ -10,6 +10,7 @@ class ListingResponseBuilder
     protected $query;
     protected $request;
     protected $fields;
+    protected $hiddenFields;
 
     protected $filterOperators = [
         'eq'   => '=',
@@ -24,11 +25,12 @@ class ListingResponseBuilder
     /**
      * ListingResponseBuilder constructor.
      */
-    public function __construct(Builder $query, Request $request, array $fields)
+    public function __construct(Builder $query, Request $request, array $fields, array $hiddenFields )
     {
         $this->query = $query;
         $this->request = $request;
         $this->fields = $fields;
+        $this->hiddenFields = $hiddenFields;
     }
 
     /**
@@ -40,6 +42,7 @@ class ListingResponseBuilder
 
         $total = $filteredQuery->count();
         $data = $this->fetchData($filteredQuery);
+        $data = $data->makeVisible($this->hiddenFields);
 
         return response()->json([
             'data' => $data,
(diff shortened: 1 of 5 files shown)


ssddanbrown:
Looks good so far @itbane. The create/update endpoints will be the trickier ones, Lots of conditional logic tied up on those requests. Don't worry about covering every use-case for the initial implementation, simplicity is preferred. 

If you haven't already worked it out, Comments on the controller methods are automatically used as the description for the endpoint in the docs. We'll also need to ensure requests & responses exist in the [dev/api folder](https://github.com/BookStackApp/BookStack/tree/master/dev/api) which are also used in the docs. Will also need tests to at least cover the main positive case for each endpoint.

itbane:
Great! It'll take a couple of days, but I'll keep adding endpoints, documentation and tests.

xinin:
Please, we need those features as soon as posible.
We are trying to use the API to create automatically our users and set their permissions.

It would be awesome to be able to change permissions set of a book, shelves ... using your API

ssddanbrown:
Hi @itbane, How is this going?  No massive rush from my side but if you get to a point where you can't progress, or can't devote additional time for any reason, then that's fine but just let me know and I'll consider this as available to pick up & continue.

ssddanbrown:
Closing this in favour of progress being continued in #3238, due to lack of response/progress here.
Thanks @itbane for your original starting efforts here.

PR #2735, opened 2021-05-06
Fix table width styling on pages rendered in markdown

### LINKED ISSUE #2732: display of entries in Chrome/Edge(based on Chromium) are not right
Hi,

today I opend bookstack on a other computer with the browser Chrome and Edge on it. There I opened an entry and it looks bad:

After I saw this, I opened the same entry on opera and all was okay:
2021-05-05 11_31_24-Veeam Backup Plan _ Knowledgebase - Vivaldi

dopyrory3:
PR to resolve #2732 
Tables generated by the markdown renderer don't honour the max-width property without applying word-break styling to the td elements. 
This has shown to fix the table overflow tested in Chrome, Edge, Opera, and Firefox

### DIFF
--- a/resources/sass/_pages.scss
+++ b/resources/sass/_pages.scss
@@ -122,9 +122,6 @@ body.mce-fullscreen, body.markdown-fullscreen {
     max-width: 100%;
     height: auto !important;
   }
-  td {
-    word-break: break-word;
-  }
 
   // diffs
   ins,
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this @dopyrory3.
If we're going to be adding this as a default style, it should be on both `th` and `td` elements along with the other default styles: https://github.com/BookStackApp/BookStack/blob/master/resources/sass/_tables.scss#L8

ssddanbrown:
Thanks @dopyrory3, Will merge in now for next patch release.

PR #2755, opened 2021-05-19
Add Keycloak socialite provider

cyber-simon:
Simple change to add the Keycloak provider from the existing Socialite providers.

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -187,6 +187,11 @@ TWITTER_APP_SECRET=false
 TWITTER_AUTO_REGISTER=false
 TWITTER_AUTO_CONFIRM_EMAIL=false
 
+KEYCLOAK_APP_ID=false
+KEYCLOAK_APP_SECRET=false
+KEYCLOAK_AUTO_REGISTER=false
+KEYCLOAK_AUTO_CONFIRM_EMAIL=false
+
 # LDAP authentication configuration
 # Refer to https://www.bookstackapp.com/docs/admin/ldap-auth/
 LDAP_SERVER=false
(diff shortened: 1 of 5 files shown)


ssddanbrown:
Thanks for offering this @cyber-simon, But I've come to the decision I don't want to grow out the supported third-party social provides much further, at least without a high level of benefit/request from the user-base. It's not a large amount of work to initially implement but it gets troublesome attempting to support people here when issues eventually arise.

Meteorus:
@ssddanbrown this is really a sad decision. I'd like to use this feature for OpenID. So i would be happy too, if you merge this feature into the master branch.

etgocode:
That unfortunate, this feature would have saved me at least 2 working days. :(

cyber-simon:
@ssddanbrown Hey Dan! I tried your new solution, and this seems to be a nice way of doing this. Unfortunetly i ran into one bigger problem:

The Keycloak socialite doesn't work, without a scope set (in the getDriver method):
```php
if ($driverName === 'keycloak') {
    $driver->scopes(['openid']);        
}
```
If this one is not set, the redirect parameter scope will be empty, which violites the OpenID Connect standard. Keycloak won't accept the request.

Would it be okay, to take these lines in? Could be configurable like `config('services.keycloak.scope')`. Or would it be better, to define a getDriver hook in the funtions.php?

ssddanbrown:
Thanks for the feedback @cyber-simon.
I've opened #2759 to allow you to do that via some level of `configureDriver` (Or similar) callback using the theme system. Have assigned to the next patch release.

ssddanbrown:
Hi @cyber-simon,
I've just released v21.04.6 with the additional option to configure the driver. 
Please see the very bottom example on the updated docs page: https://github.com/BookStackApp/BookStack/blob/master/dev/docs/logical-theme-system.md

Just an advisory, in the future I may be looking to include https://github.com/wikimedia/composer-merge-plugin as a method to handle multiple composer.json's so you don't have to manually import the required classes from the Socialite provider, just have them installed within your theme's composer usage. This should be better in the long run but could be a breaking change.

PR #2768, opened 2021-05-24
[sec] Fixes a few minor vulnerabilies when using target="_blank" on links (RSPEC-5148)

CorruptComputer:
Using `target="_blank"` on links without adding `rel="noopener"` can be a minor vulnerability. 
See [RSPEC-5148](https://rules.sonarsource.com/html/RSPEC-5148) for more info on the specifics of the issue.

This is not really an issue on most up to date browsers (Chrome 88+, Firefox 79+ or Safari 12.1+) though I think its always best to program defensively since you don't know what browser the end users will end up using. 

If you don't see any benefit in adding this feel free to close.


ssddanbrown:
Thanks for this @CorruptComputer. Most of these are internal, hence wouldn't really matter, but there are a few potential external ones in there that could be prone to such scenarios. Will merge in now for the next feature release.

Issue #2760, opened 2021-05-21
Searching for CJK works only with keywords that put inside quotation marks


dgwxx:
**Describe the bug**
Searching for CJK works only with keywords that put inside quotation marks

**Steps To Reproduce**
I added some text in CJK like '一部販売店がHGST製品に対して Seagate の同等製品',
 - tried to search for 'HGST' -> No matches
 - tried to search for '"HGST"' -> 1 match
 - tried to search for 'Seagate' -> 1 match
 - tried to search for '販売店' ->  No matches
 - tried to search for '"販売店"' -> 1 match

**Expected behavior**
To match the keywords without quotation marks.

**Your Configuration (please complete the following information):**
 - BookStack v21.04.5
 - MariaDB　10.4.19
 - Apache/2.4.47 (Win64)
 - PHP/8.0.6

I checked the source code in BookStack/app/Entities/Tools/SearchRunner.php and I find the code near line 152
`$query->orWhere('term', 'like', $inputTerm .'%');`
the '%' is not in pairs. I fixed this code, and all keyword can be matched.
`$query->orWhere('term', 'like', '%'. $inputTerm .'%');`

ssddanbrown:
Thanks for raising @dgwxx.
This is a known issue and already open under issues #1037 and #778 so I'm therefore going to close this off as a duplicate.

Issue #1037, opened 2018-09-29
Searching for word only works if put inside ""


tp199314:
**Describe the bug**
When searching for String i did not receive any results. I will get results, when i am searching for "String" (added " before and after in the search box).
This issue is simmilar to https://github.com/BookStackApp/BookStack/issues/531
I am on BookStack v0.24.1, but i tried to regenerate the search index using the command "php artisan bookstack:regenerate-search" in /var/www/bookstack

**Steps To Reproduce**
1. Go to BookStack
2. Click into Search Field
3. Search for String
4. No Result.
5. Search for "String"
6. Got the correct Result.

**Screenshots**
![image](https://user-images.githubusercontent.com/2600276/46244484-af3e0200-c3df-11e8-80ee-fc2345fe5cb1.png)
![image](https://user-images.githubusercontent.com/2600276/46244489-c250d200-c3df-11e8-8b5a-41cf674b75b1.png)

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): BookStack v0.24.1
 - PHP Version: 7.2.10-0ubuntu0.18.04.1
 - Hosting Method (Nginx/Apache/Docker): Apache

tp199314:
The string i am searching for is part of a headline (Header Tiny). When i put the string in normal text, the search is working. Is it possible to add search in headlines to the search function?

bandresen:
I have the same issue. I thought it was related to numbers in my search term but after putting them in quotes I was able to find them.

Turns out the numbers I was looking for were part of a Tiny Header and a Link.

bm751613:
Same Issue here. I have tested this with text in a Medium Header as well as paragraph text within a table. The same issue seems to happen with multi-word but not with every search.

scureza:
I have a page of about 500 lines, movie titles, in some cases the search finds the page i'm looking for only if i put the word in quotes. Usually happens with the first word of the line. I also tried on the demo site and the same thing happens.

mojaedukacja:
Here the same.
Any new solution or workaround?

Please help.

Langlais115:
BookStack v26.03.3

I've noticed this issue when I search for the first word in bullet and numbered lists.

### Example:
If I search for **server** the search will not find the page
If I search for **car**, the search will find the word (notice that I've add a space before **car**)
If I search for **"server"**, the search will find the word (notice the double quote around the **server** word into the search bar)

_Bookstack page content:_
```
- Printer
- server
-  car
[...]

Issue #778, opened 2018-03-31
Chinese search cannot find words in the middle of a sentence.


jasoncheng7115:
### For Feature Requests

Desired Feature:

### For Bug Reports

* BookStack Version: v0.20.0

When the word I'm looking for is the first word, or there's a space in front of it, it's ok.
![i01](https://user-images.githubusercontent.com/30381035/38159052-a175106e-34d3-11e8-875d-304da7a1bbaa.png)

But if the word is in the middle of a sentence, it cannot be found.
![i02](https://user-images.githubusercontent.com/30381035/38159056-aad016fe-34d3-11e8-83b0-31db2080f243.png)

Whether this is a full-text retrieval of related issues?

Thanks!

alexwyl:
The same problem in version: v0.25.1, I have just tried BookStack...

lotustalk:
The same problem in version: v0.24.3, I use a docker

linpengxuan:
still the same problem in v26.4. hope it could get solved. thanks

sosize:
you can use "成功" for search,  maybe the word segmentation has the bug, hope fix it

LeonLiuY:
Confirmed this issue still in v0.27.5
One of my team member is hesitating because of this. Would like to see it fixed.

hlj:
Hope fix this issue soon.

ssddanbrown:
Sorry about this issue. It essentially stems from my unfamiliarity with non-English text.

At the moment BookStack splits up page content, on certain characters such as spaces and some punctuation, into terms which are put in the database for indexing then a "Starts With" match of those are checked against on a normal search.

As @sosize has mentioned, you can wrap a search in quotes, at which point BookStack will perform a "contains" against the content directly instead of the above "Starts With". This is not the default simply due to performance. ("Starts With" searches can use indexes much more effectively than "Contains").

I'm not really sure how we could utilise the "Starts With" system for such characters. Perhaps the search should default to a "Contains" search if such characters are found in a term?

sosize:
@ssddanbrown Can this be set as config control ？Select “Starts With” or “Contains” for search type.

More is hope full-text search.

 Or how to quickly modify the code?

lishuai199502:
can i replace all the "startWith" with "contains",or how to modify the source code ,sorry ,i'm a noob

lishuai199502:
Hi,all the guys,I fixed this problem in v0.28.3.Just add a '%' in SearchService.php.
[...]

PR #2776, opened 2021-05-31
Fix for CJK works only with keywords that put inside quotation marks

dgwxx:
Issue #2760,  #1037 and #778.


ssddanbrown:
Thanks for offering this @dgwxx but this will prevent the database from using indexes to handle searches.
If we implemented this, we might aswell delete the whole search term system and search against the content directly. 
Ideally I'd want search actions to be indexed by default. Therefore I'm going to close this off but if you have a solution or idea that would achieve this in an indexed manner feel free to put it forward.

There's a class which splits out page content into "terms". One alternate solution may be to update that logic to split CJK content down further.

Issue #1464, opened 2019-05-24
Attachment are always triggerd to download


JtheBAB:
**Describe the bug**
When clicking to a attachment it always offers to download instead of using the default browser settings.

**Steps To Reproduce**
1. Go to a page with an attachment (pdf)
2. Click on the attachment link
3. Attachment is offered as a download instead of respecting the default behavior of the browser

**Expected behavior**
Bookstack should respect the default behavior of the browser. In my case open the PDF with the inbuild PDFReader from FireFox.

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): 0.26.1
 - PHP Version: 7.3
 - Hosting Method (Nginx/Apache/Docker): Apache

ssddanbrown:
Thanks for the report @JtheBAB.

To be honest this is purposeful, to standardise how all attachments are handled. As soon as you stop forcing this behaviour it becomes unreliable as different file types will behave differently then this is multiplied by how browsers will handle  those different file types in a different manner.

JtheBAB:
Hi  @ssddanbrown 

I see your point. But i think it would be better when the user settings are respected.

It came just now in my mind:

I don't know how complicated it is but why not add a setting on the "edit profile" page to control if it should be forced (default) or free?

ssddanbrown:
> I see your point. But i think it would be better when the user settings are respected.

The trouble is, that it's often not up to the user. It's often up the browser and the overall users' system.

> I don't know how complicated it is but why not add a setting on the "edit profile" page to control if it should be forced (default) or free?

Would not be complicated to add but extra settings can be a burden on maintenance & support while opening the floodgates to more settings. Additionally, I'd imagine that admins in many multi-user environments would prefer consistency over user preference.

Perhaps we can look at achieving this in a different manner. The attachment link behaviour could stay as-is but have a little dropdown icon next to the attachment to select download type, Or a key modifier so a ctrl+click (Default open in new tab shortcut) opens it using browser behaviour.

JtheBAB:
Sounds good :)

mattkerrison:
Please implement this! PDFs just download instead of opening in a new tab, it’s so painful

APlueckthun:
It would be extremely (!) appreciated by all of the users of BookStack in our place, if PDFs can be viewed inline, as opposed to being dowloaded. 
[...]

Issue #1270, opened 2019-02-09
Full PDF Wrapper


recklessnl:
**Describe the feature you'd like**
Bookstack is a Wiki-like app, a perfect place to store manuals. A ton of manuals are stored as PDFs. It would be so much easier if Bookstack had native PDF Wrapper support enabled for users to instantly load the PDF file within the Bookstack page. 

**Describe the benefits this feature would bring to BookStack users**
Pretty self-explanatory, it makes Bookstack even more viable for small businesses or families.

**Additional context**
Currently you can only download PDFs to your PC. There are other docker apps like Grocy who have full PDF wrapper support and it really makes it so much easier. Really hope this can get added.

ssddanbrown:
Hi @recklessnl,
Thanks for the request.

This seems very similar to #705, Do you see this issue as different to that one or can I close this as a duplicate?

recklessnl:
It's different, because we shouldn't want a 'preview', we should have a full wrapper that enables users to directly 'stream' the PDF file in question. This way you can embed the PDFs directly into the overall Bookstack page, alongside all the other tools you are currently have implemented. It's not so much a preview as a fully functioning PDF embed / wrapper. I can't think of any downside to this.

ssddanbrown:
Ah, Okay, Thanks for confirming the difference. 

> I can't think of any downside to this.

- Web PDF renders can have varying and heavy performance due to the complexities in rendering PDFS.
- Web PDF renders can often have quirks and not fully support all features of a PDF.
- This would be extra functionality to implement & maintain, Especially if built into the editors, Using up more maintainer time.
- This could encorouge users leaving content in PDF format, instead of page content, where it becomes much harder for BookStack to index for search/discovery and revision tracking.

recklessnl:
1) A PDF wrapper is optional feature users can add to their pages. If it doesn't perform up to standard, they can use the current system.
2) Same as above, although I have to say I love the one Grocy is using. Works fine for my needs.
3) Can't comment on this, all I can say is it would make bookstack a lot better and more useable
4) So what? More power to the users. I want some pages to be just one big PDF manual page. Who are you to tell me how to use it? Again, you don't HAVE to use it if you don't want it.

Must have feature for me.

ssddanbrown:
> 4\. So what? More power to the users. I want some pages to be just one big PDF manual page. Who are you to tell me how to use it? Again, you don't HAVE to use it if you don't want it.

[...]

Issue #705, opened 2018-02-18
[Feature Request] Preview for PDF Attachments


ghost:
### Feature Request

Hey Everybody,
i am very happy to see the rapid progress about the App - great Work 
But i miss the function, that I can get a preview of attached pdf-Files, before I downloading it. 
It would be very nice, when these feature where implement in the next versions of Bookstack.

Yoginth:
Change it from 
`'Content-Type' => 'application/octet-stream',` 
**to**
`'Content-Type' => 'application/pdf',`
**in** 
`app/Http/Controllers/PageController.php` (`exportPdf()` Method)

lommes:
@Yoginth 
I think you got the question wrong. It is not about the generation of PDF-files.

@ITforHome 
Can you please give some details on how the preview should look? 

Currently the attachements are only downloaded. In case of PDF there are severall possibilities to implement a "preview". There can be a real preview in form of a generated image including only the first page of the pdf, change the 'Content-Type' for this download (as suggested by @Yoginth) so the browser uses the defined configured preview or embedd a preview using pdfjs or something. In the last 2 cases the file technically is already downloaded to your computer.

Yoginth:
no @lommes if you use `application/pdf` it will be viewable, not downloadable

lommes:
@Yoginth yes and no.
File attachments are handled by AttachmentController.php (not by the PageController) and Downloads by the [get-method](https://github.com/BookStackApp/BookStack/blob/master/app/Http/Controllers/AttachmentController.php#L188). If you change the Content-Type for all filetypes this will most likely result in errors when you attach and later download a zip, exe, ... file.

Yoginth:
fine got it!

ITforHome:
@Yoginth 
i have changed the Content-Type to `'Content-Type' => 'application/octet-stream'`  in the PageController.php file, but is not working. When i click on an attached pdf-File, i can only downloaded it, but i can't get a preview. 

@lommes 
i have changed the Content-Txpe to `'Content-Type' => 'application/pdf'` in the AttachmentController.php file, but is not working too. When i also click on an attached pdf-File, i can only downloaded it, no preview.

General:
In my question i mean a attached pdf-File NO the generated PDF-File of a page in Bookstack

ITforHome:
When I open the Bookstack App on Mobile Safari or Chrome I get a preview about attached PDF-Files.

albertmatyi:
[...]

PR #2791, opened 2021-06-06
Attachment serving without forced download

ssddanbrown:
This allows Ctrl/Cmd clicking of attachments to open them with an `?open=true` query string which responds with an `inline` content-disposition and does some content-type sniffing to set an expected `Content-Type`. 

Related to #1464, #705, #1270


erenirmak:
I used linuxserver.io docker container of bookstack (v25.02) but cannot ctrl+click on attachment to open it without downloading..should I change image ?

Issue #2809, opened 2021-06-14
Feature Request - New Update Available Alert in Admin dashboard 


Cave-Johnson:
**Describe the feature you'd like**
An indicator next to the version in the bookstack admin panel stating that a new version is available. Similar to Gitlabs "Update Available / Update ASAP"
![image](https://user-images.githubusercontent.com/19388624/121964993-47cd6980-cd64-11eb-96ab-db776d4f4886.png)

![image](https://user-images.githubusercontent.com/19388624/121965126-7c412580-cd64-11eb-94b8-d8394185c19a.png)

**Describe the benefits this feature would bring to BookStack users**
This would allow bookstack admins to easily see that a new version was available and if a new release is out that contains critical security fixes, would help admins know that there is an update out.

**Additional context**
This would help admins not signed up to newsletters / email updates and who don't check git regularly

siegerhansma:
Took a first stab at it in https://github.com/BookStackApp/BookStack/pull/2859

<img width="869" alt="Screenshot 2021-07-28 at 21 41 23" src="https://user-images.githubusercontent.com/3443339/127385983-744e64d7-5e29-41c4-9659-fc9a23d9be54.png">

ssddanbrown:
BookStack is now [managed on Codeberg](https://codeberg.org/bookstack/bookstack).

As part of this change, we're closing all existing GitHub issues and pull requests.
You'll instead find this issue on Codeberg here: 
https://codeberg.org/bookstack/bookstack/issues/2809
  
Any further discussion for this will now be done on Codeberg.
Therefore this GitHub thread will be locked.

PR #2859, opened 2021-07-28
Feature - Version check on Settings page

siegerhansma:
I took a first stab at this feature in issue https://github.com/BookStackApp/BookStack/issues/2809. It uses the Github API to check the latest release tag for this repository and uses version_compare to do a comparison. I only tested it with valid release versions (so no -dev). 

@ssddanbrown 

* I added a new translation, what would be the correct way to add this to the repo? 
* Do you think this fits well with the rest of the application design wise? 

Screenshot 2021-07-28 at 21 41 23

Please let me know if there is anything unclear or missing. Thanks!


ssddanbrown:
Thanks for this @siegerhansma, All appears to work nicely from a quick test on my dev instance.

> I added a new translation, what would be the correct way to add this to the repo?

Please just add only the English translation value for the initial implementation. Our translation system will automatically create entries for all other languages once this is merged into master. So just editing and adding an entry to the `resources/lang/en/settings.php` file for this one. 

> Do you think this fits well with the rest of the application design wise?

Yeah, I'm happy with the simplicity, We don't need anything fancy here.

A few additional things it would be good to solve before I merge:

- [ ] Could you please remove any changes to the `package-lock.json` from this? Since there are no dependency changes to be made here.
- [ ] Please remove the `if (app()->runningUnitTests()) {` part. I will want to test this. It's fine to leave tests failing due to this, I'll look to mock the httpFetcher maybe globally for tests.
- [ ] There's [an option defined here](https://github.com/BookStackApp/BookStack/blob/master/app/Config/services.php#L14) that we'll need to honour in this since we're calling an external service; So always assume latest if `config('services.disable_services')` is true.

ssddanbrown (changes requested):
As per previous comment, marking this as changes requested.

ssddanbrown:
Hi @siegerhansma,
Would you still be looking to address the above to get this feature in? If you could at least do the first it would mean I could merge this while still attributing you, otherwise it makes things a little complicated while ensuring correct attribution.
Otherwise, Please let me know if you won't be able to make changes and I'll close this off.

ssddanbrown:
Will therefore close this off. Feel free to re-submit this with the above requests in the future if not already addressed.

PR #2877, opened 2021-08-08
Concurrent page editing: more warnings

MatthieuParis:
Currently, warnings are displayed if two users are editing the same page at the same moment. More precisely, a warning is displayed if a second user starts editing a page when a first user has already opened the editor AND saved a first draft. If those two people open the page at the same moment (thus no draft is saved yet), they get no warning.

This patch adds a warning each time a draft is saved, when several users are editing the same page (with a draft already saved for each of them) or when the edited draft is older than the last page version.

### DIFF
--- a/app/Entities/Tools/PageEditActivity.php
+++ b/app/Entities/Tools/PageEditActivity.php
@@ -26,6 +26,7 @@ class PageEditActivity
      */
     public function hasActiveEditing(): bool
     {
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Thanks for contributing this @MatthieuParis.

I followed this up with f99af807d044b55ef8ec6c43d8ad8c7d61b47146 to add testing and make various other changes. All described in the commit message. 
Now merged, will be part of the next patch release.

PR #2902, opened 2021-08-30
Fixed SAML logout for ADFS

theodor-franke:
This PR. adds some options to the .env file for configuring the logout with SAML. It also sets the correct NameId in the logout request. 

This PR enables:
- Signed Logout SAML-Assertions (default is disabled)
- Sets the correct email address in the logout assertion
- Enables the encoding of lowercase URLs (default is disabled)
- some little tweaks in the logout process so that it works with ADFS (theses changes must be enabled in the .env file, without any changes in the .env file the logout process is not altered.)


theodor-franke:
In ADFS you need to have a Rule like this:
```
c:[Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress"]
 => issue(Type = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier", Issuer = c.Issuer, OriginalIssuer = c.OriginalIssuer, Value = c.Value, ValueType = c.ValueType, Properties["http://schemas.xmlsoap.org/ws/2005/05/identity/claimproperties/format"] = "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress", Properties["http://schemas.xmlsoap.org/ws/2005/05/identity/claimproperties/spnamequalifier"] = "https://<FQDN_OF_BOOKSTACK>/saml2/metadata");
```

and in your `.env` file you need to add the following lines:

```
SAML2_SP_CERTIFICATE=MIIFtT... (your cert)
SAML2_SP_PRIVATEKEY=MIIJRAIBADANBgkqhkiG... (your private key)
SAML2_SP_NAME_ID_Format=urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress
SAML2_SP_NAME_ID_SP_NAME_QUALIFIER=https://<FQDN_OF_BOOKSTACK>/saml2/metadata
SAML2_RETRIEVE_PARAMETERS_FROM_SERVER=true
SAML2_LOGOUT_REQUEST_SIGNED=true
SAML2_LOGOUT_RESPONSE_SIGNED=true
SAML2_LOWERCASE_URLENCODING=true
```

ssddanbrown:
Thanks for offering this @theodor-franke.

To be honest it'll probably take some time before I review this deeply, since I know it's going to consume about a day's worth of effort for me to go through each of the 8 added options to understand their functionality so I can assess the backwards compatibility and testing we'll need to add where possible.

At a first glance, This seems like a lot of new options just to fix ADFS SLS. Are all these changes required to make ADFS SLS work? Or could you describe exactly what was failing for ADFS SLS? If this goes beyond I'd prefer to keep the scope limited to just fixing the ADFS SLS issues unless there's significant demand/response to go outside of that.

In regards to the implementation, I have not functionally checked anything, but one thing I've seen from a quick look is that any `env()` usages outside of the `Config/` directory will need to be altered to use the config system instead like the existing options set. Using `env()` outside this directory can cause issues, especially for those who cache their config for performance.

theodor-franke:
I only tried this on our coporate ADFS-Server (windows 2016 so I think its called ADFS4.0) and without these fixes the logout would not work. 
The first important thing is the Certificate, ADFS requires the logout-assertion to be signed. The other important thing ist that the correct EntityID is set in this SAML-Assertion. I made all  these Settings and not only one big `USE_ADFS` Setting because I think that other IDP may need these configs as well.

regarding the `env()` problem: If you think that this is a big problem I will try to find a solution for this. In my opinion this isnt a big deal because you will only ajust these ones and than (hopfully) never touch them again.

nomnombearclaw:
I've been trying to get SAML working with ADFS running on Windows Server 2019 and not having much luck. I've combed the documentation and still haven't had much luck. Since I'm running Windows Server 2019, are these new features/settings required? Does anyone have their working setup documented so I could go off something that is known to work? Any help would be appreciated.

theodor-franke:
I can't share my Documentation because I dont work any longer at the place where I installed this. However the docs are a great starting point (https://www.bookstackapp.com/docs/admin/saml2-auth/) and if you want you can write me when you have Problems.

PR #2904, opened 2021-08-31
Allow mastodon certification for footer label

MickGe:
Add footer tag relationship `me` if label contains 'Mastodon' (insensitive case).
It allows the certification of mastodon profile.

### DIFF
--- a/resources/views/common/footer.blade.php
+++ b/resources/views/common/footer.blade.php
@@ -1,7 +1,13 @@
 @if(count(setting('app-footer-links', [])) > 0)
 <footer>
     @foreach(setting('app-footer-links', []) as $link)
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @MickGe,
Did you mean to open this as a PR to the repo? 

If so you might need to explain the use-case of this a bit further as I'm not sure what this is doing.
If this is for verification of a non-related platform, I don't really want to be adding extra bits of code to maintain for unrelated processes of other projects and services.

ssddanbrown:
Thanks for providing additional information @MickGe.

I'm going to go ahead and close this since I wouldn't really be looking to support these odd bits just to suit the process of external websites. Thanks for still offering a contribution back to the project though.

You don't have to handle patching the system for this kind of thing, You could use our built-in theme system which allows you to override any view without altering core app files. Details on this can be found here:
https://github.com/BookStackApp/BookStack/blob/master/dev/docs/visual-theme-system.md

MickGe:
Thanks!

I didn't think about the Visual Theme System. That is perfect.

See you soon for a better contribution… 😉

PR #2960, opened 2021-10-06
OIDC implementation

ssddanbrown:
Continuation of #2169

## TODO

- [x] Look to address PHP8+ compatibility. Either choose different libraries or assist the upgrade of upstream projects.
- [x] Add basic-level auto-discovery to handle key & endpoint fetching.
- [x] Update and ensure coverage of automated testing.
- [x] Test with multiple auth providers
   - [x] Okta
   - [x] Keycloak
   - [x] Auth0
 - [x] Sponsor phpseclib

## References

- [OIDC Implementers Guide](https://openid.net/specs/openid-connect-basic-1_0.html#IDToken)
- [OIDC Discovery Spec](https://openid.net/specs/openid-connect-discovery-1_0.html)
- [steverhoades/oauth2-openid-connect-client](https://github.com/steverhoades/oauth2-openid-connect-client)
- [Okta Autodiscovery Example](https://dev-377627.oktapreview.com/.well-known/openid-configuration)
- [phpseclib](https://phpseclib.com/) - Potential solution for key handling/verification. 

## For Docs

- Mention default cache time of 15 mins for auto-discovery.
- Advise that only RS256 id token signing is currently supported.

### DIFF
--- a/app/Auth/Access/Oidc/OidcProviderSettings.php
+++ b/app/Auth/Access/Oidc/OidcProviderSettings.php
@@ -149,7 +149,7 @@ class OidcProviderSettings
 
         if (!empty($result['jwks_uri'])) {
             $keys = $this->loadKeysFromUri($result['jwks_uri'], $httpClient);
-            $discoveredSettings['keys'] = array_filter($keys);
+            $discoveredSettings['keys'] = $this->filterKeys($keys);
         }
 
         return $discoveredSettings;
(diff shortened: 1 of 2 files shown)


sg3s:
Some of your dependencies you added here have overlapping logic/functionalities, and you might be able to abstract some heavy lifting of handling keys and doing checks may be outsourced to other/better dependencies.

The library phpseclib/phpseclib seems to have some overlap with lcobucci/jwt (sub dependency of league/oauth2-client). I've personally not used phpseclib but am not / was not impressed by lcobucci/jwt (though the v4 is much better, but it took a while to release).

We've adopted the following which helped us get rid of more boilerplate key retrieval/matching logic and signature & claim checks. It might help to reduce logic here too. Especially the checking a token against a keyset url is powerful and supported algorithms is pretty useful.
https://web-token.spomky-labs.com/

Of course make your own judgement, I must admit I've not been able to go through all code in detail.

PR #3060, opened 2021-11-18
Add Sorting functionality within the chapter

Shubhamt04:
Added sorting functionality with the chapter:
- Able to sort the pages within the chapter based on selected filter.
- Able to save/update the sorting pattern for sorting.
- Added Chapter sorting entities for `en` language.

image

image


ssddanbrown:
> I'm kinda stuck now. So, can you suggest me an approach to save sorting pattern for chapter?

You should be able to re-use the existing `applySortUpdates` function instead of re-creating a new chapter-specific one for saving the new sort order. You're going to need to update the `getChaptersInvolvedInSort` function you got to select the correct chapter references out of the sortMap.

resources/js/components/chapter-sort.js
```diff
@@ -0,0 +1,189 @@
...
+    pagesFirst: function(a, b) {
+        const aType = a.getAttribute('data-type');
+        const bType = b.getAttribute('data-type');
+        return (aType === 'page' ? -1 : 1);
+    },
+    pagesLast: function(a, b) {
+        const aType = a.getAttribute('data-type');
+        const bType = b.getAttribute('data-type');
+        return (aType === 'page' ? 1 : -1);
+    },
```
ssddanbrown, line 29: These sort types don't appear to have any purpose.

resources/js/components/chapter-sort.js
```diff
@@ -0,0 +1,189 @@
...
+            parentChapter: parentChapter,
+            type: type,
+            chapter: chapterId
+        });
+
+    }
+
+}
+
+export default ChapterSort;
```
ssddanbrown, line 189: This entire class shares a massive amount of logic with the `book-sort.js` file. Could this not instead be made into a generic `content-sort.js` file? I just don't want duplicate classess/functions to maintain if possible.

resources/lang/en/entities.php
```diff
@@ -166,6 +166,15 @@
     'chapters_permissions_active' => 'Chapter Permissions Active',
     'chapters_permissions_success' => 'Chapter Permissions Updated',
     'chapters_search_this' => 'Search this chapter',
+    'chapters_sort' => 'Sort Chapter Contents',
+    'chapters_sort_named' => 'Sort Chapter :chapterName',
+    'chapters_sort_name' => 'Sort by Name',
+    'chapters_sort_created' => 'Sort by Created Date',
+    'chapters_sort_updated' => 'Sort by Updated Date',
```
ssddanbrown, line 173: Can just use the existing book-based string instead of creating new translation strings with the same content.

resources/lang/en/entities.php
```diff
@@ -166,6 +166,15 @@
...
+    'chapters_sort_pages_first' => 'Pages First',
+    'chapters_sort_pages_last' => 'Pages Last',
```
ssddanbrown, line 175: Should be no need for these strings.

resources/views/chapters/parts/sort-box.blade.php
```diff
@@ -0,0 +1,26 @@
...
+    <h5 class="text-chapter entity-list-item no-hover py-xs pl-none">
+        <span>@icon('chapter')</span>
+        <span>{{ $chapter->name }}</span>
+    </h5>
+    <div class="sort-box-options pb-sm">
+        <a href="#" data-sort="name" class="button outline small">{{ trans('entities.chapters_sort_name') }}</a>
+        <a href="#" data-sort="created" class="button outline small">{{ trans('entities.chapters_sort_created') }}</a>
+        <a href="#" data-sort="updated" class="button outline small">{{ trans('entities.chapters_sort_updated') }}</a>
+        <a href="#" data-sort="pagesFirst" class="button outline small">{{ trans('entities.chapters_sort_pages_first') }}</a>
+        <a href="#" data-sort="pagesLast" class="button outline small">{{ trans('entities.chapters_sort_pages_last') }}</a>
```
ssddanbrown, line 11: As per other comments, Don't think there's a purpose for these.

ssddanbrown (changes requested):
Thanks for this @Shubhamt04. Looks like it's coming together.
Got a few changes to request, Most of them relatively quick and minor.

Shubhamt04:
> Thanks for this @Shubhamt04. Looks like it's coming together. Got a few changes to request, Most of them relatively quick and minor.

Sure @ssddanbrown , Will work on the suggested changes.

ssddanbrown:
Since there's not been much recent active progress on this I'm going to close this off.
Feel free to re-open though if this could be implemented with the above comments addressed. 
In addition, I think there's since been some further changes in the core codebase which would need incorporating for this work if proceeding (A fair amount of extra permission checks).
Either way, thanks for your efforts so far @Shubhamt04 and for your offer of this code.

PR #3068, opened 2021-11-21
Set taborder for TOTP Verification

raccettura:
Adding tabindex=0 means when pressing tab the focus goes right to the TOTP input field.  When using a Password Manager this makes it easier than having to hit tab 3X to get the right focus.

### DIFF
--- a/resources/views/mfa/parts/verify-totp.blade.php
+++ b/resources/views/mfa/parts/verify-totp.blade.php
@@ -6,6 +6,7 @@
     {{ csrf_field() }}
     <input type="text"
            name="code"
+           tabindex="0"
            placeholder="{{ trans('auth.mfa_gen_totp_provide_code_here') }}"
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for the PR @raccettura.

> Adding tabindex=0 means when pressing tab the focus goes right to the TOTP input field.

Have you tested this? From my understanding, and testing of this PR, this is not the case. 

Even if this PR did change the tabbing order I try not to alter the tab order to ensure we're causing an odd accessibility experience. I think we've used an `autofocus` attribute elsewhere. Might be more apt here, assuming no accessibility concerns.

raccettura:
Seemed to work for me in Chrome.  But you make some good points to prefer `autofocus` so updated the PR to reflect that.

ssddanbrown:
Thanks @raccettura. Merged for next patch release.

Issue #2951, opened 2021-10-03
Grouping audit log by user's IP address


johnroyer:
**Describe the feature you'd like**
BookStack is now logging user's IP address in audit log. There should be functionality to access audit log by IP address.

**Describe the benefits this feature would bring to BookStack users**
It is easier to know the relationship between user and IP address.

**Additional context**
This is an enhancement from issue #2747. Please assign this issue to @johnroyer. Thanks.

ssddanbrown:
Thanks for the suggestion @johnroyer.

Rather than a specific visual grouping view I'd suggest that we just provide the ability to search via IP address. 

> Please assign this issue to @johnroyer

Just to confirm, does that mean you're looking to work on an implementation for this yourself?

johnroyer:
I think searching via IP address us not implement yet. And this is more important then grouping.

I am  not starting, just keep it in mind.

johnroyer:
I have some times now, I think I will do the "Search by IP" functionality.

 It some thing I missed ?

ssddanbrown:
@johnroyer Don't think you've missed anything. Feel free to work on and create a PR for IP search.

johnroyer:
Audit log did not have a field for searching IP address. There are some UI mocks as candidates .

----

Put a field in the right of form :

![ip-address-field](https://user-images.githubusercontent.com/114491/140860976-8fb1e3b7-4ff6-4505-a40f-f2849833e8cf.png)

----

Maybe IP address is not a part of date range form. Put it under Event Filder:

![event-filter](https://user-images.githubusercontent.com/114491/140861630-c0e3c255-7757-4641-8beb-97f517c99eb8.png)

----

Or IP address is should be isolated from the others. I do not know which is better. Any suggestions?

ssddanbrown:
@johnroyer On the right of the form is fine if there's room on a desktop view :+1:

ssddanbrown:
Since searching by IP has been added in #3081 I'll close this off.

PR #3081, opened 2021-11-26
Search by User IP Adress in Audit Log

johnroyer:
Features:

- add input field for searching IP address in audit log view.
    - add a hidden submit button in view. Form submits automatically after user press `Enter`
- add search type in audit log controller. IP address search in full string match. (use `=` not `like`)
- create index for user ip address (DB migration)

UI preview :

擷取選取區域_003

----

Note: related with issue #2951

### DIFF
--- a/app/Http/Controllers/AuditLogController.php
+++ b/app/Http/Controllers/AuditLogController.php
@@ -46,7 +46,7 @@ class AuditLogController extends Controller
             $query->where('created_at', '<=', $listDetails['date_to']);
         }
         if ($listDetails['ip']) {
-            $query->where('ip', 'like', $listDetails['ip'] . "%");
+            $query->where('ip', 'like', $listDetails['ip'] . '%');
         }
 
         $activities = $query->paginate(100);
--- a/app/Http/Controllers/AuditLogController.php
+++ b/app/Http/Controllers/AuditLogController.php
@@ -46,7 +46,7 @@ class AuditLogController extends Controller
             $query->where('created_at', '<=', $listDetails['date_to']);
         }
         if ($listDetails['ip']) {
-            $query->where('ip', '=', $listDetails['ip']);
+            $query->where('ip', 'like', $listDetails['ip'] . "%");
         }
 
         $activities = $query->paginate(100);
--- a/app/Http/Controllers/AuditLogController.php
+++ b/app/Http/Controllers/AuditLogController.php
@@ -20,6 +20,7 @@ class AuditLogController extends Controller
             'date_from' => $request->get('date_from', ''),
             'date_to'   => $request->get('date_to', ''),
             'user'      => $request->get('user', ''),
+            'ip'        => $request->get('ip', ''),
         ];
 
         $query = Activity::query()
@@ -43,6 +44,9 @@ class AuditLogController extends Controller
         }
         if ($listDetails['date_to']) {
             $query->where('created_at', '<=', $listDetails['date_to']);
+        }
+        if ($listDetails['ip']) {
+            $query->where('ip', '=', $listDetails['ip']);
         }
 
         $activities = $query->paginate(100);
--- a/database/migrations/2021_11_26_070438_add_index_for_user_ip.php
+++ b/database/migrations/2021_11_26_070438_add_index_for_user_ip.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddIndexForUserIp extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('activities', function (Blueprint $table) {
+            $table->index('ip', 'ip_address_index');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/AuditLogController.php
```diff
@@ -44,6 +45,9 @@ public function index(Request $request)
         if ($listDetails['date_to']) {
             $query->where('created_at', '<=', $listDetails['date_to']);
         }
+        if ($listDetails['ip']) {
+            $query->where('ip', '=', $listDetails['ip']);
```
ssddanbrown, line 49: I think it would prove useful if this was a prefix-based search so, for example, you could search based on the first 3 parts of an IP (`127.0.0.`) in the event that someone was searching for an IP range.

database/migrations/2021_11_26_070438_add_index_for_user_ip.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('activities', function (Blueprint $table) {
+            $table->index('ip', 'ip_address_index');
```
ssddanbrown, line 17: Could the index instead be called `activities_ip_index`? Would align better with other indexes and the column name.

resources/lang/en/settings.php
```diff
@@ -123,6 +123,7 @@
     'audit_table_date' => 'Activity Date',
     'audit_date_from' => 'Date Range From',
     'audit_date_to' => 'Date Range To',
+    'ip_address' => 'IP',
```
ssddanbrown, line 126: We already have the `audit_table_ip` string above. Could we just use that translation instead for now? I think showing "IP Address" as the label is better suited while aligning with the table, plus it'll reduce required translations.

resources/views/settings/audit.blade.php
```diff
@@ -41,12 +41,19 @@
...
                     <label for="owner">{{ trans('settings.audit_table_user') }}</label>
                     @include('form.user-select', ['user' => $listDetails['user'] ? \BookStack\Auth\User::query()->find($listDetails['user']) : null, 'name' => 'user', 'compact' =>  true])
                 </div>
+
+
+                <div class="form-group ml-auto">
+                    <label for="ip">{{ trans('settings.ip_address') }}</label>
+                    @include('form.text', ['name' => 'ip', 'model' => $listDetails['ip'] ? \BookStack\Actions\Activity::query()->where(['ip' => $listDetails['ip']])->first() : null])
+                    <input type="submit" style="display: none">
+                </div>
```
ssddanbrown, line 56: Currently there's a bug in the input field:

- If you perform an IP-filtered search which has no results, the input value will disappear. 

This is likely due to this line:

```php
'model' => $listDetails['ip'] ? \BookStack\Actions\Activity::query()->where(['ip' => $listDetails['ip']])->first()
```

Ideally I wouldn't want view-side database queries for information we already have (The searched IP address). I understand that it's akward though due to how the `text.blade.php` file expects a `$model`. 
For a non-database hack I think you could instead do the following:

```php
'model' => (object) $listDetails
```

ssddanbrown (changes requested):
Thanks for this @johnroyer!

I've requested some minor changes and a bugfix as part of this review. Just let me know if you have any questions on these.

In addition, it would be good to ensure we have a test added to cover this functionality. Are you familiar with PHPUnit-based testing? If not, I'm happy to add the testing before merging.

app/Http/Controllers/AuditLogController.php
johnroyer: IP address is save as string in database. It is good to go searching with a prefix.

If we want to provide searching with IPv4 / IPv4 and CIDR, it is better to save data in integer.

database/migrations/2021_11_26_070438_add_index_for_user_ip.php
johnroyer: Sure, I will update the index name.

resources/lang/en/settings.php
johnroyer: Thanks for reminding me. I did not notice there is a prefix.

resources/views/settings/audit.blade.php
johnroyer: I will use `(object) $listDetails` instead. I was ran out of ideas.

And I think I also need to change the name to `audit_table_ip_address` for translation.

johnroyer:
I can add unit tests on my own. But I can not understand how Github knows to build different environment (PHP 7.4 / 8.0 / 8.1 ... etc). The environment description did not show in `phpunit.xml`. Could you give me some directions ?

Thanks.

ssddanbrown:
> I can add unit tests on my own. But I can not understand how Github knows to build different environment (PHP 7.4 / 8.0 / 8.1 ... etc). The environment description did not show in phpunit.xml. Could you give me some directions ?

I'm not totally sure what you need here. The PHPUnit action that GitHub runs is defined here:
https://github.com/BookStackApp/BookStack/blob/master/.github/workflows/phpunit.yml

I wouldn't worry about testing on different PHP versions though for your sake, Just get it passing on one version and it'll likely pass on the others.

johnroyer:
Update:
- issues in discussion has been fixed.
- create unit test for ip searching

ssddanbrown:
Thanks for the changes @johnroyer. All looked great! Now merged for next feature release.

The only tweak I made was to the test, to check that the search would filter out non-matching IPs: 04d59763c3aaaf72a0a12760e5e9405f213cabbe

PR #3098, opened 2021-12-07
Added clarifying text to OIDC environment variables.

Northguy:
Clarifying text is now in sync with code snippets that are also available on bookstack documentation website.

I realize that there are now two areas to maintain the same information, which in future might lead to two different versions. In my opinion, the .env.example.complete should always be leading above the code snippets on the website (as this is delivered with a Bookstack install and is first point of reference when configuring).

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -134,7 +134,7 @@ STORAGE_S3_ENDPOINT=https://my-custom-s3-compatible.service.com:8001
 STORAGE_URL=false
 
 # Authentication method to use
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this @Northguy but I really don't want to pad out the `.env.example.complete` file with auth information since we do already document these on the website.

I've instead added a link to the website near these options, as we do for the other authentication methods. This has been added in 867cbe15eab76215e1e84ae285b5c443ab16959f.

Issue #2738, opened 2021-05-10
Editor Review - Initial Scoping


ssddanbrown:
We need to upgrade and/or replace the existing WYSIWYG editor. We can use this as an opportunity to find something that will better suit BookStack's users requirements now that we have almost 6 years of experience and feedback on the matter. There are currently two editors within BookStack that need to be considered:

- The WYSIWYG Editor (TinyMCE4)
- The Markdown Editor (Codemirror 5 with custom preview pane).

## Main Goals

Here are some high-level goals it would be great to achieve with a new editor. These are not hard requirements but things that would improve upon the existing setup:

- Align the functionality available within both editors.
- Allow easy switching between the editors.
- Retain all functionality, in a way that produces a compatible output, from the existing editors.
- Allow the future potential of collaboration on a page. 

## Requirements

This is primarily a list of current functionality we'll ideally need to retain to ensure forward compatibility.

- Retain a similar ease of use to the existing WYSIWYG editor.
- Sane API abilities for adding custom functionality.
   - Inserting content (At end, at start, at cursor).
   - Selection management.
   - UI button/controls support.
   - Allow external extensibility (To BookStack Users)
- Block formats: div, p, h2...h6, blockquote, codeblock (Currently codemirror embed on WYSIWYG), callout (div with set class), hr. 
- Inline formats: code, bold (strong), italic (em), underline, strike (span with inline style), Superscript (sup), subscript (sub).
- Background and text color (Span with inline style).
- Text left, center, right, justify alignment (Class `align-{center,right,left}` or inline style).
- Indenting (Inline padding-left at block element).
- Lists (ol/ul) with x-level nesting support.
- Table insert & management (With header row support), ??TinyMCE provides a lot of sub-menus here, Need to support still??
- Images
    - Managed via Custom BookStack UI.
    - Custom dimensions with UI-handle resizing.
    - Alignment (Inc Floating for left/right)  via alignment buttons.
    - Linked images.
    - Alt text.
- Links
    - Custom BookStack UI for linking to content.
    - `[target]` control.
    - Title text control.
    - Assistance for linking to in-content headings.
- Text direction for RTL language content.
- Drawings, As per images but custom functionality on double click.
- Media (iframe) embeds.
    - Dimension controls.
- Clear formatting.
- Source code viewing.
- Full screen editor viewing.
- Content preview (Markdown).
- Editor shortcuts.
- Compatible license.

## Possible Options

[...]

PR #3110, opened 2021-12-15
User can create markdown page even system is in wysiwyg mode

tangramor:
On current BookStack App, if you chose to use WYSIWYG, you cannot create Markdown page any more.

However, in a lot of companies there are both developers and other users, which developers may prefer Markdown and others use WYSIWYG.

I add a "New Markdown" link below "New Page" so user can choose which kind of page to create.

### DIFF
--- a/app/Entities/Repos/PageRepo.php
+++ b/app/Entities/Repos/PageRepo.php
@@ -154,6 +154,33 @@ class PageRepo
     }
 
     /**
+     * Get a new draft markdown page belonging to the given parent entity.
+     */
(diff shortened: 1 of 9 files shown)


ssddanbrown:
Thanks for offering this PR @tangramor but this is not the approach I'd like to take. There are currently some conversion consequences to flipping between formats, and I feel that the editor type changing between pages could cause confusion for users.

The overall goal is to allow switching instantly within editor as per work ongoing within #2738.

Apologies if this was not the response you desired.

PR #3113, opened 2021-12-15
Make building of search results work for multi-byte encoded characters

Kristian-Krastev:
When a search is made, the information that is shown in every result snippet is taken from database tables 'bookshelves', 'books', 'chapters', 'pages', but the string data in them may be encoded in format different than one-to-one byte format. For example for every cyrillic character are used up to four bytes for representation.

Operations on strings in this context are not multi-byte safe and the returned snippets contain 'broken' information.
I think the multibyte string methods are a good alternative for the solution of that problem.

### DIFF
--- a/app/Entities/Tools/PageContent.php
+++ b/app/Entities/Tools/PageContent.php
@@ -80,7 +80,7 @@ class PageContent
      */
     protected function extractBase64ImagesFromHtml(string $htmlText): string
     {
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks @Kristian-Krastev for offering this PR. Could you provide a minimal example of content and search term that causes breakage? Would help so I can add a test case to prevent regression.

ssddanbrown:
Thanks for confirming the content! PR now merged for next feature release.

PR #3140, opened 2022-01-03
Added in controller bookshelfChildren

Abe-Telo:
This will allow the use of trees  bookshelfChildren. with entities.


ssddanbrown:
Thanks for offering this @Abe-Telo but I'm not really willing to accept and maintain added data, that's used in customizations, into the core project repo. Doing so is not really feasible for long-term application performance or maintenance. 
Let me know though if I've misunderstood the idea behind this pull request.

ssddanbrown:
Will therefore close this off

PR #3176, opened 2022-01-18
fix overscroll not working on Chrome

Emmeran:
the previous fix appears not to work on Chrome - this one should work better

### DIFF
--- a/resources/sass/_tinymce.scss
+++ b/resources/sass/_tinymce.scss
@@ -69,8 +69,10 @@
   display: block;
 }
 
-.page-content.mce-content-body > :last-child {
-  margin-bottom: 3rem;
+.page-content.mce-content-body::after {
+    content: "";
+    display: block;
+    height: 80vh;
 }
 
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this @Emmeran but I can't accept as-is due to the following reasons:

- This drastically changes the size of the overscroll area, not only fixes the reported issue.
- This causes a large amount of interaction "dead zone" on Firefox. On chrome the click within the pseudo element will focus on the last block but that's not the case for Firefox. Have not tested Safari.

From a bit of testing, the current method should work on Chrome but it specifically seems the `height: 100%` applied to the html tag breaks the margin being taken into account. Or alternative the scrolling could be delegated to the body instead of the HTML. Either of these changes would affect other things though and need fair amounts of testing.

ssddanbrown:
Will therefore close this off due to the above but happy for an altered PR to be opened (with the comments addressed where possible). Should still be able to comment here if you have any questions/queries (Otherwise I generally check the discord daily)

Issue #3174, opened 2022-01-17
[Bug Report]: User creation done but reported as failed to the user


Julesdevops:
### Describe the Bug

When creating a user and selecting the option to send an invitation by email, I got an error like 'An unknown error occured'.
Checking the logs, I have this : 
```
[2022-01-17 16:54:45] production.ERROR: Expected response code 220 but got an empty response {"userId":5,"exception":"[object] (Swift_TransportException(code: 0): Expected response code 220 but got an empty response at /srv/bookstack.santepros.com/BookStack/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/AbstractSmtpTransport.php:447)
[stacktrace]
#0 .........
```

So I think my email configuration is broken right ?
But the problem is that the user was created anyway, without the assigned role/roles and without the 'user_create' log.
Maybe we could wrap this in a transaction, wdyt ?

### Steps to Reproduce

- Break your bookstack email sending config
- Create a user with email invitation
- Normally you will get an error but the user is created if you look at the user list

### Expected Behaviour

I think the user creation should fail entirely and maybe report a log entry with a high severity like : 'email sending failed' ?

### Exact BookStack Version

21.12

### PHP Version

7.4

### Hosting Environment

Ubuntu 20.04

Julesdevops:
@ssddanbrown Idk if you have started working on this, if not I can try and provide a PR. Let me know

ssddanbrown:
@Julesdevops I have not started work on this. Feel free to look to address, although I have assigned to the next patch release. If I start to wrap up the next patch release, and there's no PR, I may take this on myself.

ssddanbrown:
Addressed within #3179, To be part of the next patch release.

PR #3179, opened 2022-01-19
When creating a user, do not persist the user on invitation sending failure

Julesdevops:
See https://github.com/BookStackApp/BookStack/issues/3174

What this PR do:
- Wrap the user creation process in a transaction
- Add test


ssddanbrown:
Thanks for this @Julesdevops and extra thanks for covering with some tests!
All looks great and seems to work wonderfully when manually testing. Will merge for next patch release.

Issue #3205, opened 2022-01-27
[Bug Report]: "Be back soon" page rendering fails due to usage of `$cspNonce` which is undefined


ssddanbrown:
Probably best to massively simplify this view to not be relying on so much app logic. 

Discovered by @Julesdevops on Discord.

Julesdevops:
@ssddanbrown I started on this, will submit a PR this week end

ssddanbrown:
Fixed in #3210.

PR #3210, opened 2022-01-29
Massively simplify the 503 error view

bourdaisj:
As suggested by @ssddanbrown this PR simplifies the 503 error view.
See https://github.com/BookStackApp/BookStack/issues/3205 for informations.


ssddanbrown:
All looked good, Thanks again @Julesdevops!

Issue #2397, opened 2020-12-09
Integration with Nextcloud


Baptistou:
**Describe the feature you'd like**
We could provide an integration with Nextcloud in 3 steps :
- [ ] Provide a SSO authentication with Nextcloud with Oauth2 support for example (#1390)
- [ ] Create a BookStack app on Nextcloud to provide an easy access to a wiki. For now, I use [External Sites](https://apps.nextcloud.com/apps/external) as a convenient way to link BookStack, but it would be awesome to have further integration with for example [Dashboard](https://apps.nextcloud.com/categories/dashboard) or inside the File interface.
- [ ] Add ability to browse from/save into Nextcloud when inserting images and file attachements to a wiki page

**Describe the benefits this feature would bring to BookStack users**
BookStack is an amazing open-source wiki software and Nextcloud is a powerful open-source file storage software.
They are both installed on my server and it would bring huge benefits if they could interact with each other.

ssddanbrown:
Thanks for the request @Baptistou.

To be totally honest, I'd personally consider this out of scope, apart from the auth which I'd only consider if not NextCloud specific, for which an issue is already open as you have linked. 

The effort to do the second and third points well would likely not be worth the effort, support and maintenance required for the small portion of the user base that may happen to use both platforms to benefit from such additions. 

Don't wanna be too close minded on this though as I may be missing something so I'll leave this open for a while so feedback can be given but I'll likely close this off after a few months.

Baptistou:
Yeah, I know this requires a huge effort while BookStack is still in beta version. Still, I am just amazed what a good project it is and a Nextcloud integration could greatly increase its visibility. Nextcloud lacks a wiki software, contrary to Office 365, and I think BookStack will greatly improve the open-source ecosystem.

See the discussion :  https://help.nextcloud.com/t/intranet-wiki-app/6416/11

> Ideally there would be an easy integration with Bookstack

> I would definitely love to see Bookstack as an app in nextcloud +1000

If you need help, I would be very interested in getting involved there. :smile:

ssddanbrown:
Sure, I can totally see how this could be useful to existing users of Owncloud.

> Nextcloud integration could greatly increase its visibility. 

I'm sure it would but I'm wary not to have visibility/popularity become a driving force behind decisions for the platform as that's a very slippery slope. Benefit to existing users & maintainability are much higher prioritised for this project at the current time.

tiredofit:
[...]

PR #3215, opened 2022-01-30
Adding nextcloud for third party auth

sebclemens:
**Request:**

We want to use our existing nextcloud instance for authentification.

**Solution:**

Add nextcloud to social providers by using the already implemented socialite providers library.

### DIFF
--- a/app/Auth/Access/SocialAuthService.php
+++ b/app/Auth/Access/SocialAuthService.php
@@ -46,6 +46,7 @@ class SocialAuthService
         'gitlab',
         'twitch',
         'discord',
+        'nextcloud',
     ];
 
     /**
--- a/app/Config/services.php
+++ b/app/Config/services.php
@@ -116,6 +116,16 @@ return [
         'auto_confirm'  => env('DISCORD_AUTO_CONFIRM_EMAIL', false),
     ],
 
+    'nextcloud' => [
+        'client_id'     => env('NEXTCLOUD_CLIENT_ID'),
+        'client_secret' => env('NEXTCLOUD_CLIENT_SECRET'),
+        'redirect'      => env('APP_URL') . '/login/service/nextcloud/callback',
+        'instance_uri'  => env('NEXTCLOUD_BASE_URI'),
+        'name'          => env('NEXTCLOUD_NAME', 'Nextcloud'),
+        'auto_register' => env('NEXTCLOUD_AUTO_REGISTER', false),
+        'auto_confirm'  => env('NEXTCLOUD_AUTO_CONFIRM_EMAIL', false),
+    ],
+
     'ldap' => [
         'server'                 => env('LDAP_SERVER', false),
         'dump_user_details'      => env('LDAP_DUMP_USER_DETAILS', false),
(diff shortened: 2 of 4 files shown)


ssddanbrown:
Thanks for offering this @sebclemens, but did you come across #2607 or #2397?

My comments made here still apply: https://github.com/BookStackApp/BookStack/pull/2607#issuecomment-819027156
Upon that, we now have an OIDC option which would replace the default logic when active, although I'm not sure if nextcloud offers itself as an OIDC identity provider.

sebclemens:
I spend a few hours on this. Nextcloud does not support OIDC. The plugin which adds this functionality did not work in my case.

But https://github.com/BookStackApp/BookStack/blob/release/dev/docs/logical-theme-system.md looks nice and could actually make my PR obsolete.

sebclemens:
I solved this with the logical theme system docs. Please update the docs at bookstackapp.com, I spend a lot of time and didn't find it.

putt1ck:
@sebclemens any chance you could share your solution for getting Bookstack to auth with NC via the logical theme system?

MarioOsswald:
I tried yesterday the available NextCloud OIDC App from https://github.com/H2CK/oidc
The Purpose of this app is to use NC as the identity provider. I came close to get this working, but could not solve an issue with the Issuer definition, in particular, I guess, it's a rewrite problem or something related to it.

The details are:
I specified in BS .env file
OIDC_ISSUER=https://mynextclouddomain.tld
When I try a browser request with "https://mynextclouddomain.tld/.well-known/openid-configuration", the NC nginx redirects correctly to "https://mynextclouddomain.tld/index.php/.well-known/openid-configuration" and displays all the desired details, among them at first line 'issuer: "https://mynextclouddomain.tld/"'.
The problem is that BS doesn't get the results from the redirected URL. Instead, I receive an error 
"OIDC Discovery Error: Error discovering provider settings from issuer at URL https://mynextclouddomain.tld/.well-known/openid-configuration". The request from BS is not redirected the way the browser request was.

My approach was to change the .env variable to
OIDC_ISSUER=https://mynextclouddomain.tld/index.php
This solves the discovery problem, but then leads to an error message like
"OIDC Discovery Error: Unexpected issuer value found on discovery response". There is then an obvious mismatch, since on BS's side the issuer variable includes the index.php (OIDC_ISSUER=https://mynextclouddomain.tld/index.php), while the discovery response delivers only 'issuer: "https://mynextclouddomain.tld/"'.

A "curl https://mynextclouddomain.tld/.well-known/openid-configuration" delivers
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>

while a  "curl https://mynextclouddomain.tld/index.php/.well-known/openid-configuration" delivers the desired results.

I'm not experienced enough with NC or nginx redirect / rewrite features. Since the browser redirects correctly (without any required change from my side), I thought that's enough. I tried a few configurations, with no success. But probably someone more experienced has an immediate idea what to do so that a curl request is also redirected correctly. If so, I guess, BS will also auto discover correctly.

Issue #1893, opened 2020-02-13
Artisan command for "refetching" user avatars for existing users.


ch0wm3in:
I've recently made my own web app, where i can serve pictures from Active Directory. It works, but only for new users because the function downloadAndAssignUserAvatar(); seems to be run on user creation.

For now the only way is to delete the user and then create the user again, which removes revisions from different books created by the user in question. This would be a much nicer way of doing it, when you already have a running instance of the application in production.

ssddanbrown:
Thanks to @MarcHagen via PR #4560, this has now been added and will be part of the next feature release.

PR #3232, opened 2022-02-03
Artisan command for "refetching" user avatars for existing users

shelmz:
Added an artisan command to search for new (gr)avatar for given user but deletes previous avatar, even if no new one is found.
Solves #1893.

### DIFF
--- a/app/Console/Commands/ResetAvatar.php
+++ b/app/Console/Commands/ResetAvatar.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace BookStack\Console\Commands;
+
+use BookStack\Auth\User;
+use Illuminate\Console\Command;
+use BookStack\Auth\UserRepo;
+
+class ResetAvatar extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     */
+    protected $signature = 'bookstack:reset-avatar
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Hi @shelmz, Thanks for offering this. Looks functionally good.

A few notes from my side:

- [ ] I think the command should be named using "Refetch" instead of "Reset" since people may instead expect it to be reset to the default empty state. Fetching enforces the idea we'd be "fetching" the image from elsewhere.
- [ ] The `$signature` needs to be updated since it still refers to `reset MFA`. 
- [ ] The `User` is being injected in via the constructor but not used, can be removed (Including the comment on line 27).
- [ ] Ideally we'd add a test to cover the functionality ([example](https://github.com/BookStackApp/BookStack/blob/development/tests/Commands/ResetMfaCommandTest.php)). Just say if not comfortable though and I can add this afterwards.
- [ ] I think we should have an optional `--users-without-avatars` flag which will run this for all users without avatars assigned. This would considerably help the scenario given in  #1893 where they presumably have many existing users without avatars, rather than them needing to run this command for each id/email desired.

ssddanbrown:
Since there's been no follow-up I'm going to close this but would be happy to re-open or accept a new PR if the above notes can be taken into account.

PR #3246, opened 2022-02-05
Wysiwyg: preserves line feeds in code block mode

### LINKED ISSUE #3200: [Bug Report]: selecting multiple lines to make them code block
### Describe the Bug

When I have multiple lines and I select them and click on 'code block', all the paragraphs on the page get smushed to one line. 
Strangely, if I select the same paragraphs and select block quote, it works perfectly.
Unrelated, but requires same bug fix: If I shift+enter some lines and then select code block, the entire paragraph gets smushed to a single line inside the code block. Again, block quote works as expected.

### Steps to Reproduce

New document
add a few lines of text, some with SHIFT+ENTER
select 2 lines from either the enter delimited and shift+enter delimited text and click code block.

### Expected Behaviour

It has to be like the block quote, retain line endings, so if you have 2 lines with an ENTER between them (2 paragraphs), the code block needs to be "line1, empty line, line2" , and if you have a paragraph with multiple lines, it needs to be "line1,line2".

_No response_

### Exact BookStack Version

v21.12.2

### PHP Version

7.4

### Hosting Environment

Ubuntu 20.4 LTS

bourdaisj:
Fix https://github.com/BookStackApp/BookStack/issues/3200

### DIFF
--- a/resources/js/components/wysiwyg-editor.js
+++ b/resources/js/components/wysiwyg-editor.js
@@ -136,7 +136,7 @@ function codePlugin() {
         const selectedNode = editor.selection.getNode();
 
         if (!elemIsCodeBlock(selectedNode)) {
-            const providedCode = editor.selection.getNode().textContent;
+            const providedCode = editor.selection.getNode().innerText;
             window.components.first('code-editor').open(providedCode, '', (code, lang) => {
                 const wrap = document.createElement('div');
                 wrap.innerHTML = `<pre><code class="language-${lang}"></code></pre>`;


ssddanbrown:
Thank you once again @Julesdevops!

This change was definitely an improvement to preserve the newlines.
Unfortunately it alone would not fully address the issue in #3200.

When the selection spanned across multiple top-level nodes, inserting a code block would attempt to (and breakingly) convert the whole page into a code block. You would also see all text within the code content editor popup. This was due to my usages of `getNode()` within the below lines:

https://github.com/BookStackApp/BookStack/blob/815f8d79ed2d4dcb30be3a7b17b257c1a0208eb6/resources/js/components/wysiwyg-editor.js#L139-L148

When multiple top-level nodes are selected this would resolve to the shared parent body, hence we would fetch all text from those and attempt to replace the whole body with code on code-block-save.

I found a handy `editor.selection.getContent({format: 'text'})` method which is smart and provides just the selection, even if terminating part-way through a block.

I followed up your changes with b2f863e1f18ab9997dc43178ef53bc04b849e10c which uses this method and simplifies the strange stuff I was doing to set the code block, to make the selection replacement more accurate.

bourdaisj:
Indeed a far better solution! thanks for following up on this

PR #3315, opened 2022-03-08
copy composer file from composer official image

souhaiebtar:
a little optimization on how composer is put inside the php container


ssddanbrown:
Thanks for offering this @souhaiebtar but can you explain why this method is any better than the existing method?
From my perspective, downloading composer via direct official means seems more efficient and follows official usage paths compared to needing an extra container image involved, just to extract a single file.

souhaiebtar:
@ssddanbrown the docker image from which composer is pulled is official, also as it's less code which i think make it more comprehensive, but you are right it's the same thing, if you think it's not necessary i can close the request, also thank you for this great project

ssddanbrown:
> the docker image from which composer is pulled is official

Sure, but it's provided within the intent of using the container, not specifically for providing a single file.

> also as it's less code which i think make it more comprehensive

Only slightly less code in our docker file but it adds reliance to another container image and would likely require downloading a lot more data from a user perspective.

> if you think it's not necessary i can close the request

I will go ahead and close this, since I do think it adds complexity rather than reducing it.
Thanks though for looking to contribute.

> also thank you for this great project

No worries, glad to provide something others find useful!

Issue #2886, opened 2021-08-16
Error uploading files more then 150mb


dividerbeam:
![image](https://user-images.githubusercontent.com/52495131/129556184-1837354e-fcd7-4556-b08b-5bd5c5c9286f.png)

files uploading run with connection to Minio s3

all server settings added:
php:
post_max_size
upload_max_filesize
memory_limit
nginx:
client_max_body_size
client_body_timeout
js:
window.uploadTimeout
window.uploadLimit

dividerbeam:
![image](https://user-images.githubusercontent.com/52495131/129560543-9357305a-cdee-4d57-a068-ddee3163089b.png)

ssddanbrown:
Hi @dividerbeam,
Would you be able to find the log file described within this page: https://www.bookstackapp.com/docs/admin/debugging/?

Once found, could you find the latest error logged upon causing that error? That should provide more detail as to what is failing here.

dividerbeam:
![image](https://user-images.githubusercontent.com/52495131/129604825-36f5276e-29b8-4392-8e8d-8d82f0d31668.png)

but i cant change memory_limit, my .ini file in volumes being overwritten on container start

`version: "2"
services:
  bookstack:
    image: ghcr.io/linuxserver/bookstack
    container_name: bookstack
    volumes:
      - /docs:/config
      - /home/artel/mydocs:/var/www/html
      - ./php-local.ini:/etc/php7/conf.d/php-local.ini
    environment:
        ....`

php-local.ini not applied after "docker-compose up", moreover, the file from the container replaces the local file

ssddanbrown:
@dividerbeam Based upon the following lines:
https://github.com/linuxserver/docker-bookstack/blob/3aec0b9e4974eed807f2f363c250b8e62abadf01/root/etc/cont-init.d/50-config#L84-L85

I would assume that the linuxserver/bookstack image is already providing a `php.ini` at `config/php/php-local.ini` to provide customizations, hence you might be fighting that. Based on your volumes, do you have a `/docs/php/php-local.ini` file? and if so have you tried adding your customizations there?

dividerbeam:
> @dividerbeam На основе следующих строк:
> https://github.com/linuxserver/docker-bookstack/blob/3aec0b9e4974eed807f2f363c250b8e62abadf01/root/etc/cont-init.d/50-config#L84-L85
> 
[...]

PR #3365, opened 2022-04-02
Add data streaming where beneficial to reduce memory usage

ssddanbrown:
### Areas to address

- [x] - Attachment Downloads
- [x] - Attachment Uploads
- [x] - API Attachment Downloads
- [x] - API Attachment Uploads
- [ ] - ~~Image Uploads~~ Upload path already looks to be at reasonable minimum. Resizing is loaded into memory either way. 18MB image on 32MB memory uploads okay.
- [x] - Image Downloads? Was already streaming.

### Todo

- [x] Test when using s3-like storage for each area
    - Tested on minio, Works very nicely. 

### Related 

Related to #2886


ssddanbrown:
Ready for merge when appropriate.

PR #3373, opened 2022-04-05
Tiny header

evandroamaro:
Had the same translation as the small header. Corrected the translation.


ssddanbrown:
Thank you for this @evandroamaro!

Issue #3372, opened 2022-04-04
Recycle bin endpoints


bourdaisj:
### API Endpoint or Feature

Ability to read from the recycle bin, as well as restoring items.

### Use-Case

Context: currently working on a side project for fun using the rest API.
After item deletion,  I want to allow the end-user to cancel the deletion afterward.

bourdaisj:
Can get started on this today

ssddanbrown:
@Julesdevops Awesome, Would be happy to accept a PR for such a feature.
Just follow convention for the other API endpoints. 
Will need to ensure the endpoints are covered by tests.
Will also need to ensure any request/response body examples are added to `dev/api`: https://github.com/BookStackApp/BookStack/tree/development/dev/api

Just shout if you need any guidance!

ssddanbrown:
Code now merged as detailed in #3377, Will be part of the next feature release. Thanks again @Julesdevops!

PR #3377, opened 2022-04-07
[API] Recycle bin endpoints

bourdaisj:
Related to: https://github.com/BookStackApp/BookStack/issues/3372
This is a draft PR to discuss and make sure I'm on the right track.

### DIFF
--- a/app/Entities/Repos/DeletionRepo.php
+++ b/app/Entities/Repos/DeletionRepo.php
@@ -21,6 +21,7 @@ class DeletionRepo
         /** @var Deletion $deletion */
         $deletion = Deletion::query()->findOrFail($id);
         Activity::add(ActivityType::RECYCLE_BIN_RESTORE, $deletion);
+
         return $this->trashCan->restoreFromDeletion($deletion);
     }
 
@@ -29,6 +30,7 @@ class DeletionRepo
         /** @var Deletion $deletion */
         $deletion = Deletion::query()->findOrFail($id);
         Activity::add(ActivityType::RECYCLE_BIN_DESTROY, $deletion);
+
         return $this->trashCan->destroyFromDeletion($deletion);
     }
 }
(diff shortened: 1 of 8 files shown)


ssddanbrown:
Hi @Julesdevops, 
This is looking great so far, can't see any major things to change at this time.
Just let me know when all is implemented & ready from your side and I'll jump in for a full review.

bourdaisj:
I think the three endpoints I already implemented are fine as-is.
I'm thinking about adding a fourth one, 'read', which would returns the full chain of deleted items (children & parents) with their ids as well as some (limited) informations about them.
Wdyt? @ssddanbrown

ssddanbrown:
@Julesdevops I would prefer the most minimal approach that covers your use-case, which sounds like it may be covered by the existing three?

If (and only if) the extra detail was really needed I'd be tempted to say an additional parameter on the existing book/chapter/page/shelf  "read" endpoints might be better suited to avoid needing a more complex implementation for this specific use-case.

ssddanbrown:
@Julesdevops Let me know your intention either way. No rush on the actual work but I'll be wrapping up the next feature release over the next week so, if you're happy with the current state, I can plan to sneak this in `v22.04` otherwise we can plan for a latter release.

bourdaisj:
@ssddanbrown It already covers my use-case indeed. If someone is facing a more complex use-case someday I'd be happy to follow-up on this though.
I'll do a little pass on the code, doc and then I will mark this as ready for review

ssddanbrown:
Thanks again for this @Julesdevops, Now merged for next release.

During review, I did make a series of changes detailed in commit ff8dadefee9e400bb46a54b301a8b701c4a0b61b. Everything was functionally great and high-quality, most of the changes I made were just alignment with existing codebase conventions. The main significant other change was around how entity children and parent details were provided back on the list response. If you have any queries on my changes feel free to ask.

I wasn't a fan of exposing internal namespacing for API use (`deletable_type`) but I saw the value in having this for identification and filtering, so I also followed this up with 0930e8519c508d46a848db93c76f258da5ee84a9 which simplifies all internal morph types within the database to simple `page/book/chapter/bookshelf` names instead of namespaces.

bourdaisj:
@ssddanbrown Fair enough. Thanks for following up on this so quickly!

Issue #2175, opened 2020-07-07
Enable Auto Login for SAML2


tiredofit:
**Describe the feature you'd like**
I feel it would be useful to have full single sign on capabilities for SAML2 without having to click the login button.

**Describe the benefits this feature would bring to BookStack users**
This would avoid a click for users to login, and be automatically sent to the destination page.

sswany:
This is relatively crude, but I was looking to do the same thing. I added the following to the Custom HTML head content and seems to work great. Basically just clicks on the SSO login button if its visible on the page after the DOM loads:
```
<script>
window.addEventListener("load", function () {
   document.getElementById("saml-login").click();
});
</script>
```

ssddanbrown:
With the merge of #3406 this functionality is now available as a built-in option.
This will be part of the next feature release.

PR #3406, opened 2022-05-02
Skip intermediate login page with single provider

### LINKED ISSUE #3216: [Feature Request]: Skip intermediate login page with single provider

When a single authentication provider is configured, skip the intermediate user interaction on the `/login`-page and redirect directly to the auth provider's actual login page.

### Describe the benefits this feature would bring to BookStack users

50%-reduction in clicks-to-login :P

### Additional context

With OIDC configured, users currently have to go click on "Login" at top left, then land on `/login`, where they select the only available OIDC provider (in my setup) before the are redirected to for the actual authentication page.

rjmidau:
Resolves #3216, #2175

When a single authentication provider is configured, it is now possible to skip the intermediate user interaction on the /login page and redirect directly to the auth provider's actual login page, saving an extra click for users.
This new behaviour is disabled by default, and can be enabled by setting the AUTH_AUTO_REDIRECT environment variable to true.

This new behaviour only works for saml2 and oidc providers, and will not work if Social Auth providers are enabled.
This new behaviour will not be honoured for /login if argument ?logout=1 is passed (which is where users are redirected to after logout when AUTH_AUTO_REDIRECT is set to true).

This also introduces an overridden logout() function so post-logout redirection can be honoured (it appears redirectAfterLogout has not been honoured in Laravel since ~5.3 - see laravel/framework#15525)

### DIFF
--- a/app/Config/auth.php
+++ b/app/Config/auth.php
@@ -13,6 +13,11 @@ return [
     // Options: standard, ldap, saml2, oidc
     'method' => env('AUTH_METHOD', 'standard'),
 
+    // Automatically redirect to external login provider if only one provider is being used
+    // instead of displaying a single-button login page and requiring users to click through
+    // Supported methods: saml2, oidc
+    'auto_redirect' => env('AUTH_AUTO_REDIRECT', false),
+
     // Authentication Defaults
     // This option controls the default authentication "guard" and password
     // reset options for your application.
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -25,14 +25,14 @@ class LoginController extends Controller
     |
     */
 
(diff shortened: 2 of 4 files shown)


isurinda:
I have applied these changes and can confirm that its working in version 22.02.3

mvicomoya:
> I have applied these changes and can confirm that its working in version 22.02.3

+1

I tried these on top of 22.04.2 and work as expected.

ssddanbrown:
Thank you very much for this @rjmidau. All looked good.
I have now merged this for next feature release.

I did follow this up with a set of changes in 8b211ed4618299a22bb2a3dda9bb317d8cde65b6. These were primarily naming changes in addition to expanding tests and making the new intermediate view an actual view with content & loading animation, with a manual-non-js option to trigger the authentication form submission as a backup.

Issue #3166, opened 2022-01-14
[Feature Request]: Allow disable SMTP certificate verification


azhinu:
### Describe the feature you'd like

I didn't find any way to use SMTP server with self signed certificate.

It would be great to have an option like 'MAIL_TLS_VERIFICATION=true/false'

To allow insecure certs we need only add folowing to [app/Config/mail.php](https://github.com/BookStackApp/BookStack/blob/master/app/Config/mail.php)
```
'stream' => [
   'ssl' => [
       'allow_self_signed' => true,
       'verify_peer' => false,
       'verify_peer_name' => false,
   ],
],
```

### Describe the benefits this feature would bring to BookStack users

Sometimes users need to use self signed certificates and usually services allow to disable TLS verification, but not bookstack.

lue30499:
I would like this as well. I have a local server just for email notifs that I just have a self-assigned cert on.

ssddanbrown:
An alternative solution is to add your certificate CA to the host's trusted certificate store, that way I don't think you'd need to then disable additional security options.

azhinu:
@ssddanbrown it's too complicated for quick local testing purposes.

MDKAOD:
+1 on this request. I can confirm that the merged code works. I modified mail.php and I was able to use smtp. Sendmail was not working and threw an empty request received 220 error. Any chance this will be pushed?

rockenren:
Upvote for adding this feature as option.

If I remember correctly, smtp without certificate validation worked some versions ago (I have test email from March, 2022 in my mailbox from our bookstack instance, mail server always have self-signed certificate).

a-carvallo:
Had this issue after updating bookstack, though the option for ssl didn't work for me since its location seems to have moved with Laravel 9 (https://github.com/laravel/framework/discussions/41385).

I also get @ssddanbrown's concerns in https://github.com/BookStackApp/BookStack/pull/3439 but in my case it's a local SMTP server with no security at all basically.

talesrodarte:
That functionality would be great. It is always necessary to correct in other ways and that takes time. It is very common for email servers to have self-signed certificates.

ssddanbrown:
This has now been added as per of #4126. Will be part of the next feature release.
The setting to add to your `.env` fill be as follows:

```bash
MAIL_VERIFY_SSL=false
```

PR #3439, opened 2022-05-17
#3166 Allow disable SMTP certificate verification

MKaterbarg:
While I don't recommend using this, I can see some users needing this for testing. This relates to request # 3166


ssddanbrown:
Thanks for offering this @MKaterbarg.

What worries me about introducing this is that it will be used to fix SSL/TLS issues that should be addressed properly via other means. I feel people will enable this, see it fixes their issue, then presume they require to have this option enabled.

To me, If they've specifically enabled an encryption option, they should want a secured setup instead of opting to disable security controls. 

---

Upon the above, if there is found to be a valid case to include this, we'd also need to consider the following:

- We'd need to know when these actions take affect. Only on certain `MAIL_ENCRYPTION` options? Which ones?
- We'll need to move to Symfony mailer soon. Can all the required settings here be used with that library?
- Might be tricky but would be great to have a test cover this as such functionality an option can easily be forgotten and become broken, especially with mailer changes on the horizon.

MKaterbarg:
@ssddanbrown  I do share your concerns. Maybe it should be an undocumented feature?

From my initial findings, I do believe the same approach should work for Symfony mailer

ssddanbrown:
>  I do share your concerns. Maybe it should be an undocumented feature?

But then we'd just be supporting an undiscoverable feature. I'm still not fully sure what the use-case is to be honest, outside of being a work-around of security. 

> From my initial findings, I do believe the same approach should work for Symfony mailer.

Some documentation (Or links to docs) of how this would work would be ideal. I'd have to be sure to ensure we're not adding functionality we cannot support. 

Out of interest, Do you specifically have a need for this option? If so, clarity on your use-case may help me understand the need for this.

ssddanbrown:
Since there's been no follow-up on this I'll close this off but feel free to still comment and this can always be re-opened if needed.

Issue #1170, opened 2018-12-10
Draw.io using SVG to support links within diagram


mark-james:
**Describe the feature you'd like**
If the draw.io integration saved the diagram as SVG rather than png then links created within the diagram could also be followed directly from the bookstack page

**Describe the benefits this feature would bring to BookStack users**
Diagrams elements can be used as shortcuts to access other pages within bookstack.

Often more visual representations are easier to understand. So a diagram could replace a table of contents. 

This provides another form of interlinking between pages within bookstack.

**Additional context**
With E11 support for bookstack disappearing anyway, is there a reason not to use SVG instead?

ssddanbrown:
Thanks for the suggestion @mark-james, I had not thought about links within drawings.

Would somewhat rely on #1103 in addition to many other changes.

> is there a reason not to use SVG instead?

Yeah, SVG's are harder to deal with, There's a lot more to them (From an editor/display point of view) whereas with a PNG everything is effectively encapsulated within a single unit.

mark-james:
Ok. Thanks @ssddanbrown . 

I'm happy to contribute a pull request if you could point me in the right direction as far as changes that need to be made? (Or at least things I need to consider)

This is a fairly important change to me as I'm going to start moving a knowledge base onto bookstack soon. The existing solution while inferior in many ways unfortunately does allow for links within diagrams. So it's going to cause me some conflict if I can't provide users this feature.

ssddanbrown:
### Considerations

* The back-end image upload endpoints would need to be made compatible with SVG.
* The front-end image manager would need to be updated to work with SVG.
* The draw.io front-end code would need to be updated to handle draw.io SVGs while ensuring full feature compatibility when re-importing (So no layer loss for example, Would need to check if SVG's retain all data themselves or if extra metadata needs to be saved).
* All of the above would need to be made backwards compatible with the existing png-based system.
* Both the markdown editor and WYSWIYG editor drawing implementations would need to be updated to handle SVG's, Might be a challenge to cleanly a large drawing as SVG content in markdown due to the amount of content that will be embedded. Might tie into next point:
* Need to consider how to display SVG drawings without issue, Without limited possibility of external page styles affecting them. This will be tricky as I think draw.io drawings can contain standard elements such as `<h1>` or `<p>` tags. Maybe you could iframe them into the page? Would mean cleaner markdown code but sizing/layout gets trickier. Perhaps shadow DOM could work here? Needs investigation.

[...]

Issue #1103, opened 2018-11-01
 Feature Request: Allow svg images


TBK:
**Describe the feature you'd like**
Make it possible to upload svg images.

**Describe the benefits this feature would bring to BookStack users**
Scalable Vector Graphics WTF 👍 

**Additional context**
This is what happens at the moment:
![image](https://user-images.githubusercontent.com/858296/47861734-bea3d700-ddf3-11e8-83d6-cce173c4db2e.png)

wouterloedeman:
I agree. Would be nice if they could be used as application logo as well!

xblitz:
Yes!  Just tried using SVG logo since the current requirement of 43px Height looks bad on HiDPI Screens (4k, phones)

bharatrajagopalan:
Hi

EDIT: a slight tweak that lets me get rid of the **div** element by using the beforeInject hook of SVGInject

Not perfect but managed to get interactive SVG working with the following

I am using plantuml to generate SVG for UML diagrams

I upload svgs as attachments and then am using styles to add a nice grey background. i am using the excellent SVGinject javascript to find img tags with class plantsvg and replace it with the SVG content. Consequently as SVG is basically html text, i am using an enclosing div to center it

<details><summary>Details</summary>
<p>

Paste this into Settings > Custom HTML Head

<img width="818" alt="image" src="https://user-images.githubusercontent.com/15735432/50048646-d0062180-00c9-11e9-946d-e0e3c1864f67.png">


```javascript
<style> 
 .plantsvg  {
   height:auto;
   background-color:#FAFAFA;
   padding:20px;
   max-width:100%;
 }

 .plantdiv {
   text-align:center;

 } 
</style>

<script>


!function(o,l){var r,a,s="createElement",y="getElementsByTagName",g="length",E="style",d="title",b="undefined",h="setAttribute",k="getAttribute",w=null,x="__svgInject",A="--inject-",C=new RegExp(A+"\\d+","g"),S="LOAD_FAIL",t="SVG_NOT_SUPPORTED",I="SVG_INVALID",v=["src","alt","onload","onerror"],L=l[s]("a"),j=typeof SVGRect!=b,f={useCache:!0,copyAttributes:!0,makeIdsUnique:!0},G={clipPath:["clip-path"],"color-profile":w,cursor:w,filter:w,linearGradient:["fill","stroke"],marker:["marker",
[...]

PR #3452, opened 2022-05-22
SVG Image and Drawing Support

ssddanbrown:
The main intent of this PR is to support SVG images and use SVG-based drawings.
This PR will explore integrating drawings through the use of `<embed>` tags to allow interactivity.

Related to #1103 and #1170

### TODO

- [x] Support SVG for the standard image gallery.
- [x] Support SVG format for drawings and make it the default.
- [ ] Update drawing editor logic to use embed tags.
    - [x] WYSIWYG editor support.
    - [x] Markdown editor support.
    - [ ] Handle embeds for contained HTML exports.
    - [ ] Handle embeds for  PDF exports.
- [ ] Testing
    - [ ] Ensure old PNG drawings work for all editor functionality.
    - [ ] Ensure SVG images are handled/display in exports.

### Docs Update

Potentially breaking change: The `displayEl` of the `editor-markdown::setup` editor event will now no longer be an iframe, but instead the direct parent wrapper of markdown preview content.


ssddanbrown:
Kind of stuck with this right now, as per my comment here:
https://github.com/BookStackApp/BookStack/issues/1170#issuecomment-1136182504

PR #3455, opened 2022-05-24
Fix typos and add translations in FR

ericgaspar:
- Add missing translations and fix typos.


ericgaspar:
Damn, just saw that you are using Crowdin for translations...

PR #3473, opened 2022-06-01
feat: mattermost socialite provider

LouisHrg:
Hello, 
Is it possible to add this provider to bookstack ? 

Thank you :)

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -201,6 +201,12 @@ TWITTER_APP_SECRET=false
 TWITTER_AUTO_REGISTER=false
 TWITTER_AUTO_CONFIRM_EMAIL=false
 
+MATTERMOST_APP_ID=false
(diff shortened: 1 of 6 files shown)


ssddanbrown:
Hi @LouisHrg,

Thanks for offering this but I've generally stopped extending our range of built-in social providers due to maintenance impact.
Instead, for cases where an additional provider is needed, we do provide an option to extend your BookStack instance to add social drivers using our logical theme system:
https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md#custom-socialite-service-example

This would allow you to add the provider without editing any core app files.
Let me know if you need any guidance at all or if you get stuck.
Note, you'd be able to set an icon by adding an `icons/auth/mattermost.svg` file within your theme folder.

LouisHrg:
Ok thank you very much for your response ! Cheers

PR #3483, opened 2022-06-08
Feature/favourites api

### LINKED ISSUE #3471: Favourites accessible via API
### API Endpoint or Feature

At the moment the favourites of a user are not visible in the API. The status should be visible in the following calls as `isFavourite: boolean`:
- GET /api/[shelves|books|chapters|pages]
- GET /api/[shelves|books|chapters|pages]/{id}

In addition a user should be able to set/unset them
- PUT /api/[shelves|books|chapters|pages]/{id}/favourite
```
{
    isFavourite: boolean
}
```

### Use-Case

Feature completeness of the API. Make all the information stored in Bookstack available over the API

### Additional context

_No response_

bytepoets-ssi:
Resolves #3471

That would be a simple implementation of favourites in the API. I haven't tested the performance impact for the list API call. There is as well a slightly different situation as at the other API calls which we can discuss, because every user can change its favourite status regardless of the permission.

### DIFF
--- a/app/Entities/Models/Page.php
+++ b/app/Entities/Models/Page.php
@@ -39,7 +39,7 @@ class Page extends BookChild
 
     public $textField = 'text';
 
-    protected $hidden = ['html', 'markdown', 'text', 'restricted', 'pivot', 'deleted_at'];
+    protected $hidden = ['html', 'markdown', 'text', 'restricted', 'pivot', 'deleted_at', 'favourites'];
 
     protected $casts = [
         'draft'    => 'boolean',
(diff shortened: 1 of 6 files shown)


ssddanbrown:
Thanks for offering this @bytepoets-ssi, but it would be ideal if the conversation in https://github.com/BookStackApp/BookStack/issues/3471#issuecomment-1143321310 can be continued first to provide perspective on use-case so we're not focused on a specific implementation. For example, it may be better to that we add a range entity-generic `/favorites` endpoints instead of requiring new endpoints on each content type. Validating the use-case requirement would be the first step to this assessment though.

ssddanbrown:
Since there's been no further discussion in #3471 as per my message above I'll close this off. 
This can always be re-opened if required.

PR #3490, opened 2022-06-09
Improve 'LanguageTest::test_locales_config_key_set_properly'

andrii-bodnar:
Hey everyone!

In this PR I've updated the `LanguageTest::test_locales_config_key_set_properly` Unit test to better handle such cases when a new locale is added in Crowdin. 

As we can see [here](https://github.com/BookStackApp/BookStack/actions/runs/2468769080), a new locale was added in Crowdin and the test fails. The strict comparison is unnecessary. It should be enough to check if there are corresponding language folders to the config locales. Nothing happens if the lang folder will contain extra language.

In addition, I've made a minor fix of Crowdin name in the issue template 🙂

### DIFF
--- a/.github/ISSUE_TEMPLATE/language_request.yml
+++ b/.github/ISSUE_TEMPLATE/language_request.yml
@@ -1,5 +1,5 @@
 name: Language Request
-description: Request a new language to be added to CrowdIn for you to translate
+description: Request a new language to be added to Crowdin for you to translate
 labels: [":earth_africa: Translations"]
 assignees:
   - ssddanbrown
@@ -23,7 +23,7 @@ body:
         This issue template is to request a new language be added to our [Crowdin translation management project](https://crowdin.com/project/bookstack).
         Please don't use this template to request a new language that you are not prepared to provide translations for.
       options:
-        - label: I confirm I'm offering to help translate for this new language via CrowdIn.
+        - label: I confirm I'm offering to help translate for this new language via Crowdin.
           required: true
   - type: markdown
     attributes:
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this PR @andrii-bodnar but I'm not really sure I understand the point of this change.

> As we can see [here](https://github.com/BookStackApp/BookStack/actions/runs/2468769080), a new locale was added in Crowdin and the test fails. The strict comparison is unnecessary. It should be enough to check if there are corresponding language folders to the config locales. Nothing happens if the lang folder will contain extra language.

I would expect this test to fail, since the locales config no longer matches the available languages. This test is designed to check when there's a difference between the language files and the locales config. With the provided change, the test would no longer fail on mis-match, instead only failing when the config contains more options which is a rarer scenario.

Let me know if I've mis-understood this change though.

andrii-bodnar:
@ssddanbrown yes, you understood me correctly. The test will fail if the config has some language that does not exist in the langs directory, but if the lang directory contains some language missing in the config - it will pass. 

This is ok in my opinion. Languages could be merged and the config can be updated later (for example, if we don't want to add some language to the config until it will not be fully translated) - in this case, the application will work correctly, all tests will pass, and will not distract other developers.

ssddanbrown:
> This is ok in my opinion. Languages could be merged and the config can be updated later (for example, if we don't want to add some language to the config until it will not be fully translated)

The trouble is that this test is supposed to fail in the scenario of new languages being added, that's it's entire purpose, as a reminder that a new language has been added and that the config needs to be updated to align, hence this PR invalidates the reason I wrote the test.

It shouldn't really distract other developers since the I10n branch won't be under their responsibility.

andrii-bodnar:
@ssddanbrown okay, as a reminder - it makes sense.

PR #3503, opened 2022-06-16
Fix Crowdin name in the language_request issue template

andrii-bodnar:


ssddanbrown:
Thanks for this @andrii-bodnar!

Issue #3239, opened 2022-02-03
[Feature Request]: Migrate Self Referential Links on Copy


Arcynic49:
### Describe the feature you'd like

I document a software product for work, and initially skipped over considering migrating to BookStack (from Confluence) because of the lack of copy functionality. One workflow I needed to be able to use was copying the documentation for one version into a new version, and then changing the sections that were updated in the new version. With the recent release of copy book/chapter functionality, I've started getting ready to use BookStack (it is very well made) but I've noticed that the copy functionality isn't working exactly how I hoped.

While the copy does in fact create a new fully copy of everything, any self-referential links do not get updated in the new copy. It would be desirable for these links to point to the new copy rather than the original. By way of example (meant to be illustrative, not how I'm structuring stuff currently):

- Book- Introduction
  - Page- Landing Page
  - Page- Products
  - Page- FAQs
- Book- Software Version 1.0
  - Page- Release Notes, has link to Book-Introduction/Page-Products
  - Page- Installation, has link to Book-Software Version 1.0/Page-Release Notes
  - Page- User Guide, has links to Book-Software Version 1.0/Page-Admin Guide
  - Page- Admin Guide

If you perform a copy of Book- Software Version 1.0 to version 2.0, the section below shows the results (actual links in bold, desired in italics):

- Book- Software Version 2.0
  - Page- Release Notes, has link to **Book-Introduction/Page-Products** _(this is desired)_
  - Page- Installation, has link to **Book-Software Version 1.0/Page-Release Notes** _(should be Book-Software Version 2.0/Page-Release Notes)_
  - Page- User Guide, has links to **Book-Software Version 1.0/Page-Admin Guide** _(should be Book-Software Version 2.0/Page-Admin Guide)_
  - Page- Admin Guide

This example shows it for a book, but the logic conceptually applies to chapter copies as well as page copies (if there happen to be links on a page referring to itself). Anything linked via permalink should naturally be exempt from these rules.

### Describe the benefits this feature would bring to BookStack users

Ideally this would enable faster updates of content following a copy operation, as you would not need to manually update links that the copy has that are meant to refer to itself-as-a-copy and not the original-prior-to-copy. Since this logic only applies to copy operations and not moves, it is thus hopefully more contained and useful.

weiyuemin:
I have exactly the same condition, hope this feature..

bourdaisj:
Hi, I started taking a look at this

bourdaisj:
[...]

PR #3504, opened 2022-06-16
Migrate self referential links on copy

bourdaisj:
Early draft PR for https://github.com/BookStackApp/BookStack/issues/3239

- [x] Figure out which tests we need to add to cover every use-case/path
- [ ] Write the actual tests
- [ ] Do the actual implementation
- [ ] Test it on entities with very large content to make sure we're not harming performances too much


ssddanbrown:
Thanks for making a start on this @Julesdevops but, since it's been over a year, I'm going to go ahead and close this off.
I would offer to continue this, or offer help to progress it, but things have changed a fair bit and there's probably now better/easier ways to go about this since we added specific reference tracking after this PR was started.

bourdaisj:
make sense, and apologies for the idle PR @ssddanbrown

PR #3556, opened 2022-07-06
Fixed comment count update error

GongMingCai:


ssddanbrown:
Thanks for providing these translation fixes @GongMingCai.
Note: Our main way of managing translations is by using Crowdin. While we can change translation content through GitHub, sometimes such method of providing translations will conflict with our Crowdin system which can lead to lost translation content provided in this manner.

GongMingCai:
Okay，I didn't know that. I have submitted zh_CN translation on Crowdin

PR #3570, opened 2022-07-12
Fixed some typos and corrected grammar.

smartshogu:
+ punctuation marks are made similar between de and de_informal


ssddanbrown:
Thanks for providing these translation fixes @smartshogu, Now merged for next feature release.

Note: Our main way of managing translations is by using Crowdin. While we can change translation content through GitHub, sometimes such method of providing translations will conflict with our Crowdin system which can lead to lost translation content provided in this manner.

PR #3621, opened 2022-08-06
Add Colorpicker for table background and border

Spacefish:
Enable the colorpicker plugin in tinyMCE (used in table cell background color and such)
image


Spacefish:
ah damn just ignore it, the newest version already has a working color picker

PR #3639, opened 2022-08-10
Completion of Persian translation 2022-08-10

samadha56:
Completion of Persian translation at 2022 August 10


ssddanbrown:
Thanks for these @samadha56, Now merged for next patch release.

PR #3674, opened 2022-08-24
API - Search pages by slug

n0nag0n:
I have a specific need where I need to take a wiki url and make it work "backwards" by searching for the slug and then determining the book. We have hundreds of articles and pulling back hundreds of articles every time to start comparing slugs and books would make it cumbersome and resource intensive.

I've never used Laravel before, but I think I got the code added in correctly and I think I got the API docs page updated as well. Couldn't get docker working, but I did live testing on the repo I'm hosting on my server and the changes worked as I hoped they would. Tweak whatever you need to tweak.

### DIFF
--- a/app/Http/Controllers/Api/PageApiController.php
+++ b/app/Http/Controllers/Api/PageApiController.php
@@ -15,6 +15,9 @@ class PageApiController extends ApiController
     protected PageRepo $pageRepo;
 
     protected $rules = [
+        'list' => [
+            'slug' => [ 'string' ]
+        ],
         'create' => [
             'book_id'    => ['required_without:chapter_id', 'integer'],
             'chapter_id' => ['required_without:book_id', 'integer'],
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Hi @n0nag0n,

Thank you for offering this, but this is already possible by adding a filter to the request.
For example:

```
https://demo.bookstackapp.com/api/pages?filter[slug]=mixed-content-example-page
```

You can see the result of the above by [logging into our demo](https://demo.bookstackapp.com/login?email=admin@example.com&password=password) instance then visiting that URL.

See the "Listing Endpoints" section of the API documentation ([Section on demo site](https://demo.bookstackapp.com/api/docs#listing-endpoints)) for full details on the `filter` parameter, you can do some fairly advanced filtering with this param.

n0nag0n:
Genius. Well I learned something about Laravel :) I saw that section, but it didn't sink in that it applied to any endpoint. Makes my life easier :)

ssddanbrown:
Yeah, it applies to any listing endpoint (Except for the search endpoint I think).
BTW, it's not a Laravel specific thing, it's some [custom handling we have for listing responses](https://github.com/BookStackApp/BookStack/blob/development/app/Api/ListingResponseBuilder.php).

n0nag0n:
Ahh I see. Thank you! And my Laravel comment was I had to figure out your Laravel app in order to even try to make the change, hence why I said I learned something about Laravel.

I'm functioning on 2 hours of sleep, just ignore what I say :)

PR #3742, opened 2022-09-21
Multiple LDAP servers

Khazhinov:
Hello!
I want to express my gratitude, because Bookstack has become an indispensable tool for me in my work!
But in order to get a failover option for Bookstack service on a domain network, I needed to be able to specify multiple domain controller addresses.
I propose a simple implementation of this feature.

Just specify the ```LDAP_SERVER``` variable in the .env file in the format: ```dc1.domain.local:389;dc2.domain.local:389```


.env.example.complete
```diff
@@ -207,6 +207,7 @@ TWITTER_AUTO_CONFIRM_EMAIL=false
 
 # LDAP authentication configuration
 # Refer to https://www.bookstackapp.com/docs/admin/ldap-auth/
+# If you need to use multiple addresses, separate them with a semicolon. Ex: dc1.domain.local:389;dc2.domain.local:389
```
ssddanbrown: Please can you remove this comment, this would be something I'd document on our docs website link referenced above.

app/Auth/Access/LdapService.php
```diff
@@ -217,22 +217,54 @@ protected function getConnection()
         }
 
         $serverDetails = $this->parseServerString($this->config['server']);
-        $ldapConnection = $this->ldap->connect($serverDetails['host'], $serverDetails['port']);
+        if (array_key_exists('hosts', $serverDetails)) {
```
ssddanbrown: Please could this logic be aligned so we don't have seperate logic (via this `if` statement) for single vs mulitple hosts.
See my comment for the `parseServerString` function for more detail.

app/Auth/Access/LdapService.php
```diff
@@ -217,22 +217,54 @@ protected function getConnection()
...
+                } catch (\Throwable $exception) {
+                    $fail_counter++;
+                }
+            }
 
-        if ($ldapConnection === false) {
-            throw new LdapException(trans('errors.ldap_cannot_connect'));
-        }
+            if ($fail_counter == count($serverDetails['hosts'])) {
+                throw new LdapException(trans('errors.ldap_cannot_connect'));
```
ssddanbrown: This error handling would hide indication of TLS issues (as thrown on line 239).
At the very least, it would be good to emit the last received exception message.

app/Auth/Access/LdapService.php
```diff
@@ -244,21 +276,45 @@ protected function getConnection()
...
      * Parse a LDAP server string and return the host and port for a connection.
      * Is flexible to formats such as 'ldap.example.com:8069' or 'ldaps://ldap.example.com'.
+     * If you need to use multiple addresses, separate them with a semicolon.
+     * Ex: dc1.domain.local:389;dc2.domain.local:389
      */
     protected function parseServerString(string $serverString): array
     {
-        $serverNameParts = explode(':', $serverString);
+        $explodedServerString = explode(';', $serverString);
+        if (count($explodedServerString) > 1) {
```
ssddanbrown: This change currently puts the old logic into the `else` of this condition, while adding all new logic within an `if` condition, but much of this logic is exactly the same. This also means there's two different return structures to consider.

Please could this logic be aligned so we don't have separate logic in an outer `if` statement like this?
You simplify the return type to always by an array of `['host' => $hostName, 'port' => $ldapPort]` pairs.

app/Auth/Access/LdapService.php
```diff
@@ -244,21 +276,45 @@ protected function getConnection()
...
-        }
+            foreach ($explodedServerString as $serverString) {
+                $serverNameParts = explode(':', $serverString);
 
-        // Otherwise, extract the port out
-        $hostName = $serverNameParts[0];
-        $ldapPort = (count($serverNameParts) > 1) ? intval($serverNameParts[1]) : 389;
+                // If we have a protocol just return the full string since PHP will ignore a separate port.
+                if ($serverNameParts[0] === 'ldaps' || $serverNameParts[0] === 'ldap') {
+                    return ['host' => $serverString, 'port' => 389];
```
ssddanbrown: The return here means that multi-hosts won't be handled if they include a protocol.

ssddanbrown (changes requested):
Thank you very much for offering this PR @Khazhinov!

I've performed a quick review of the submitted updates, and have picked up some changes that would be good to action, in the interest of ensuring this all works as expected for different options, and to ensure we are not duplicating code to keep things maintainable.

It would also be good to add some testing to cover this logic.

Don't hesitate to let me know if you have any issues and questions in regards to my comments, if needed I could pick this up and address the comments if you're unable to progress.

Khazhinov:
Thanks for the detailed review! Posted an update

ssddanbrown:
Thanks @Khazhinov, 
From a quick can this is looking real good now. 
Have assigned to be reviewed and merged as part of the next release.

BTW, Don't worry about keeping this branch up-to-date with `development`, I can merge it in once required.

ssddanbrown:
Just been reviewing this, [adding tests and little refactoring](https://github.com/BookStackApp/BookStack/compare/ldap_host_failover), but while doing that I found this implementation only works if `LDAP_START_TLS=true` since otherwise a connection is not attempted until later. Although a recommended option, I'm wouldn't be happy with having that caveat for this addition. 

We could do some form of connection test request, but think that would be a bit hacky. 
We always bind after connection creation, but just in a few different ways, so the failover logic could wrap `(connect + bind)` instead of just `(connect)`. I'll need to spend a little time to find a clean approach without affecting/risking existing logic.

Going to unassign this from next feature release though as I don't want it delaying anything.

ssddanbrown:
Continued in #3784, this will remain open until #3784 is merged.

ssddanbrown:
Sorry to say I'm giving up on this for now and will hence close this off, as I have done for #3784.
My attempts to make this fit have just used a lot of time and ended up with an unreasonable amount of code surface area growth relative to the benefit, which would be a feature that really has had quite minimal demand. 
Feel like it needs a new approach but I'm out of patience for this.

To be honest, I would have advised against creating a PR if asked up-front, due to the little demand for this, but the changes in the original PR looked minimal so I thought it would be a quick addition, but I've probably spent over 6 hours on this now and don't want to spend any more. Can always come back to it in the future.

Thanks @Khazhinov for offering the original PR though.

Khazhinov:
I don't understand how domain authentication can be without fallback address features.
Thank you for taking the time for this.
Hope you can solve this problem in the future.

Issue #185, opened 2016-09-10
Integration with spatie/laravel-backup


unnikked:
I've been using Bookstack for a couple of months and I really look forward for this project. 

Since it is developed with Laravel it would be nice to integrate `spatie/laravel-backup` in order to automate the backup process so we can all feel secure about our data. 

**Why I think `spatie/laravel-backup` is a good choice ?**
- It's developed by an active webagency, so the package should not be abandoned. 
- It can perform and manage backups of both database and files. 
- You can schedule backups within your Laravel app. 
- You can monitor your backups. 
- You can get notifications on slack, email, pushover and custom channels. 

**How we should integrate this package ?**
I would like to hear the opinion of the author since this package can be integrated in the UI so a user can set his own logic for the backup. 

For reference there is also a [good documentation](https://docs.spatie.be/laravel-backup/v3/introduction) about this package.

ssddanbrown:
Hi @unnikked,
Thanks for putting forward this idea.

I think building in backup to BookStack is a great idea and the linked package looks fantastic. Here are my thoughts on the idea and the `spatie/laravel-backup` package:

1. If backup is built in to BookStack then we'd have to ensure the implementation is solid as responsibility will be effectively be largely shifted from users to BookStack developers. This is not an issue, Just something to note.
2. The `spatie/laravel-backup` package looks like it does backups very well but, After a quick read of their documentation, I did not see anything about restore. I think we'd ideally implement both backup and restore.
3. I'm not sure how the spatie/laravel-backup [license requirements](https://github.com/spatie/laravel-backup#postcardware) will work for something like BookStack so I'd have to double check that. I don't want everyone who enables backup on their instance be required to send a postcard but I will respect that's their license.

unnikked:
1. Agree on that. Also we could provide a simple backup solution that just works. Shifting the maintenance to the user. We can just provide the facility since Laravel provides it. 

2. Why don't make a feature request for the package in the future? 

3. No idea on that too. But I think the end user needs to send it if he is using this project on a production server (other than for a personal use). Maybe @freekmurze can reply in regards.

freekmurze:
Hi,

1. I am dedicated to maintaining our package. You should not however that v4 will be released soon that ups the requirements to PHP 7 and Laravel 5.3. v3 should be pretty stable though

[...]

PR #3783, opened 2022-10-14
Install spatie/laravel-backup for automated & easy backups

### LINKED ISSUE #723: Feature Request:  Backup/Restore from settings page
### For Feature Requests

Desired Feature:  Backup/Restore from Settings Page

It would be a nice feature to, without configuring nightly backups, be able to take an ad hoc backup of bookstack from the interface to download, and be able to upload and restore a backup.

Backups would include SQL data and images, etc.

### LINKED ISSUE #2405: Backup and Restore functionality
**User story**
As the owner of a bookstack wiki that contains a lot of information that I have spent huge amounts of time in curating. In addition this information is very important to me. Bookstack is so great and I use it for business and private purposes.
Because of this heavy reliance, I want to be able to know that I can ensure that I have a way to back up the entire bookstack instance. I also need to be able to restore this backup regardless of the runtime version of the Bookstack version.

A backup feature where I can back up all the information and images and structure that I have entered as an end user of Bookstack.

A restore feature where I can restore all the information and images and structure that I have entered as an end user of Bookstack, which my backup container contains. It is of course critical that restore is possible without failing, regardless if a backup package was done on an older version than what is currently installed. But this is of course hard to achieve.

Benefits for a user:
The backup and restore feature will make it very easy to secure information which I as a user really care about.

As a user I am able to quickly be able to restore Bookstack instances, and I can do it without interacting **without touching the infrastructure / container layer***

Benefits for the product/project: Bookstack will be perceived as more reliable and a secure viable solution, for use in different scenarios where information criticality is high.

**Additional context**
This is needed to make the product more mature. 
It is a highly usable feature which will make the product more attractive.

I am aware of a feature request which has been closed, regarding backup of Bookstack data. However it focused on backing up singular pages. This feature request is regarding the entire bookstack instance covering the following data objects:

** Data **
- Books
- Books structure
- Pages
- Page structure
- Page Metadata

**Configuration**
- Bookstack instance user configuration
- Users in the solution
- Groups in the solution

axelrindle:
### Description

[spatie/laravel-backup](https://github.com/spatie/laravel-backup) is a package for automated backup creation.

The default configuration includes the following data in a backup:

- Complete database dump
- `public/uploads`
- `storage/uploads`
- `.env`

I incorporated the data [mentioned in the docs](https://www.bookstackapp.com/docs/admin/backup-restore/).

This does not add any possibility to create backups from the UI. Backups can be created by running the following Artisan command:

```shell
php artisan backup:run
```

If the user wishes he may create a database-only backup:

```shell
php artisan backup:run --only-db
```

I like this package because it lets me create backups with ease and without having to deal with custom scripts. And I'm lazy `¯\_(ツ)_/¯`

#### Configuration

The package is highly configurable and I already changed some of the defaults, for example the following:

https://github.com/axelrindle/BookStack/blob/336bd2204614de3f20016e4eb4dc12e1a113ca26/app/Config/backup.php#L119-L122

In my eyes the biggest point to discuss is: should there be anything else to configure via environment variables? I'd say **no** to keep the complexity of the config file small.

#### Restore

The package does not provide the possibility to restore backups which still leaves that up to the user.

> I don't think it'd be worth maintaining such logic in the core project [...]
>
> -- <cite>https://github.com/BookStackApp/BookStack/issues/723#issuecomment-1110220775</cite>

I agree with that.

#### Resources

The packages published a lot of language resources. I'm currently questioning whether they should reside in the application's resources as I don't think there will be any changes to them.

### Relations

I read through the comments in #185, #723 and #2405.

Closes #723, closes #2405.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,3 +42,13 @@ MAIL_PORT=1025
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
+
+# Backup settings
+# Specifies all the disks to which the backup should be persisted.
+# May be a comma-separated list
+BACKUP_DISKS=backup
+
+# In case of using the "backup" disk, this is the base directory
+# where backup will be stored. Should be outside of the
+# application directory
+BACKUP_ROOT=/var/backups/BookStack
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -139,6 +139,9 @@ STORAGE_S3_ENDPOINT=https://my-custom-s3-compatible.service.com:8001
 # An s3-format URL will be generated if not set.
 STORAGE_URL=false
 
+# The base directory to store backups in
+BACKUP_ROOT=/var/backups/BookStack
+
 # Authentication method to use
 # Can be 'standard', 'ldap', 'saml2' or 'oidc'
 AUTH_METHOD=standard
--- a/app/Config/backup.php
+++ b/app/Config/backup.php
@@ -1,0 +1,268 @@
+<?php
+
+use Illuminate\Support\Str;
+
+return [
+
+    'backup' => [
+
+        /*
+         * The name of this application. You can use this name to monitor
+         * the backups.
+         */
+        'name' => env('APP_NAME', 'laravel-backup'),
+
+        'source' => [
+
+            'files' => [
+
+                /*
+                 * The list of directories and files that will be included in the backup.
+                 */
+                'include' => [
+                    base_path('.env'),
+                    base_path('public/uploads'),
+                    base_path('storage/uploads'),
+                ],
+
+                /*
+                 * These directories and files will be excluded from the backup.
+                 *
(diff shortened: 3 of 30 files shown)


ssddanbrown:
Thanks for offering this @axelrindle!

To be honest though, I was hoping for some discussion in regards to the options presented in my [#723 comment](https://github.com/BookStackApp/BookStack/issues/723#issuecomment-1110220775), before diving into an implementation. 
Even taking this approach of using the Spatie package, I reflect some concerns, such as the compatibility of support pattern. Additionally, the surface area of the library is quite large in terms of what it does, while it also introduces some system requirements. We could lock down any exposed options to reduce supported surface area, but then it becomes more viable to support our own logic instead of introducing additional dependencies.

ssddanbrown:
I'm going to go ahead and close this off in favour of exploring options as per my message above.

PR #3784, opened 2022-10-17
Ldap host failover

ssddanbrown:
Continuation of #3742

### TODO 

- [x] Connection is no longer cached per-request. Either need to tweak test expectations or add caching layer in. 
- [ ] Manual Testing
- [ ] Added automated failover/error coverage


ssddanbrown:
Closing this off since I can't get it to a point I'm happy with after spending another few hours today. 
Is technically working but now has multiple layers of additional complexity and worse testing capabilities and code just to expose details for testing.
Would need a fresh approach that keeps the low-level function abstraction, but not in the way done here.

The value/demand is minimal for this feature, and I don't want to waste any more time for something of little benefit, especially as this would still need additional real-world testing which requires even more time. Closing off for now.

Issue #3792, opened 2022-10-20
Automate creation of new users from a completely fresh instance, without manual UI access


Benouare:
### Describe the feature you'd like

Hi, 

I would like to create a user by scripting (CLI or API). 
To do this, i would like to do it directly by CLI or to use the API.
The API endpoint is available to create a user, but it needs an token/secret. And to get this, if i am not wrong, i must connect to the web interface and create the token. In this case i would like to create by cli. 
I want to do this automaticaly, so i dont want to go in the configuration and click every where. I want to script it. 

### Describe the benefits this would bring to existing BookStack users

In my case, i want to be able to give an "editor/basic" account and not the admin one. I dont want to allow my users (/client) to have access to the configuration of bookstack, just to have a final/basic account.

 

 

### Can the goal of this request already be achieved via other means?

Yes. 
I can use the API, but i need a token to use it, and to have it, i must connect to the app and generate it manually. 
In my case, i want to script everything, so i want to have something like "php artisan bookstack:generate-token --user admin --enddate 12/01/2060..." OR "php artisan bookstack:create-user --name timmy --email timmy@southpark.com --password jimmy --role viewer" 

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundemental request

### How long have you been using BookStack?

6 months to 1 year

### Additional context

Thx for your feedback ;)

ssddanbrown:
Hi @Benouare,
Could you explain your use-case a bit further?
It sounds like your end goal is to create users automatically, I'm just trying to understand why it's a problem to initially access the system to create setup an account with API access to be able to do that.

Benouare:
Hi @ssddanbrown.
First of all, thx for your reply.  

>  It sounds like your end goal is to create users automatically, 

That's exactly my goal, but WITHOUT the web interface (so by CLI or API). 

> I'm just trying to understand why it's a problem to initially access the system to create setup an account with API access to be able to do that.

Part of the job is already done, as you say it's already possible to create a user with the API. 
But, if i am not wrong, to do that you need the Token/Secret of the admin user.
And, if i am not wrong, this token can be generated only by the web interface. So "we" need to connect to the web interface to generate the token/secret before to use it in the api. 


[...]

PR #3806, opened 2022-10-24
feat: add token generation command (#3792)

Benouare:


Benouare:
@ssddanbrown let me know what do you think about that. 
Cheers

ssddanbrown:
Thanks for offering this @Benouare, but as I said in #3792:

> I don't want to add new CLI commands to the core app for one-off/limited needs, nor do I want to replicate all admin tasks as API commands.

Sure this _could potentially_ be useful for others, but I don't think I've seen any past use-cases or requests that would be helped by this addition yet. As also mentioned in my #3792 comment, you could use the logical theme system to add this command to individual instances where required, without editing core app files.

ssddanbrown:
With the reasoning provided in my comment above, I'm going to go ahead and close this off.

PR #3879, opened 2022-11-29
Add popular PHP templating languages to code editor

jhit:
Smarty and Twig are two very popular PHP templating engines and might be useful to some Bookstack users too.

### DIFF
--- a/resources/js/code.mjs
+++ b/resources/js/code.mjs
@@ -26,10 +26,12 @@ import 'codemirror/mode/python/python';
 import 'codemirror/mode/ruby/ruby';
 import 'codemirror/mode/rust/rust';
 import 'codemirror/mode/shell/shell';
+import 'codemirror/mode/smarty/smarty';
 import 'codemirror/mode/sql/sql';
 import 'codemirror/mode/stex/stex';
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this @jhit. Just to confirm, do you personally have a current desire/need to use both of these code formats within BookStack?

jhit:
@ssddanbrown Hi.,
Yes, [we](https://www.sfxonline.de) develop e-commerce solutions based on [JTL-Shop](https://www.jtl-software.de/online-shopsystem) and [Shopware](https://github.com/shopware).
We recently started to use BookStack for internal documentation and as developers a significant part of this contains code samples with both of this templating engines.
At least Twig, as part of the Symfony framework should have a large user base today.

ssddanbrown:
Thanks @jhit for confirming. Just a bit too late for the next feature release, since I've wrapped that up to go out tomorrow, but I've assigned this to be in the next patch release thereafter.

jhit:
No problem @ssddanbrown. I can add the change manually for the next release. In the end, it is not a huge effort :-)
Thanks for accepting the PR

PR #3901, opened 2022-12-06
fix: upload file when create a new book

mahdieight:
fix permission error while upload image cover when to create a new book


ssddanbrown:
Thanks for offering this @mahdieight but, from a quick scan at the changes, I'd need to understand why the `chown` line, before your added line, does not provide the permissions needed since this should give the expected user full ownership of the required directories.

ssddanbrown:
Since there's been no further follow up I'll go ahead and close this off.

Issue #1747, opened 2019-10-25
Add collaborators


Luffyyy:
**Describe the feature you'd like**
Add a collaborators option for pages and what not so a user who created a page/etc will be able to give permissions to other users to also edit it.

**Describe the benefits this feature would bring to BookStack users**
This will allow users to have shared pages/etc without needing to create special roles for each group of people.

ssddanbrown:
Hi @Luffyyy,
Thank you for your request.

We already have a role permission to allow users to manage permissions on their own content.
Am I correct in thinking that, to support your request, you'd just need the ability to assign permissions per user instead of per role in the page permissions view?

Luffyyy:
Yes

ssddanbrown:
Thanks for confirming @Luffyyy. I have updated the title to reflect the core feature requirement. "Permission System Review" is on the roadmap, It will likely be at that point which this feature will be considered for inclusion.

Struki84:
Hi guys, 

I came to this thread looking I think for the same thing, if it's ok I'll just describe shortly the feture I'm interested. 
 
As an Admin user I can assign individual users specific permissions per shelf or per individual book. 

So if I have 2 users i can select which of the users will have access to  which of the books/shelfs

ssddanbrown:
Thanks for the input @Struki84, That use-case aligns with the updated title.

Just as an FYI, in the event that you're new to the project, you could currently achieve this but you'd need to create a custom role for those two users, to then use that role in the book/shelf level permissions pages. You may have already discovered this though.

Struki84:
Wow, tnx, I wasn't aware you can create custom roles, I think that would work for me.

modem7:
> Thanks for the input @Struki84, That use-case aligns with the updated title.
> 
> Just as an FYI, in the event that you're new to the project, you could currently achieve this but you'd need to create a custom role for those two users, to then use that role in the book/shelf level permissions pages. You may have already discovered this though.

Unfortunately, that does preclude the permission of "all" in public, so they'd still be able to see it unless you override the book/shelf permission, at which point it also overrides the "own" permission..........if that makes sense! Unless I'm totally missing an ability here!

cb3inco:
I would love to see something like this at the book level too. A 'private' book that people can add and remove users and adjust permissions as needed.

ramazanpolat:
This is exactly what we are looking for.
We'd like to create a book and add view/alter permissions for specific user(s).

[...]

PR #3908, opened 2022-12-07
User permissions

ssddanbrown:
As start of user permissions work. Related to #1747.

### Todo

- [x] Update entity permissions table
- [x] Update "Other users" permission handling to be denoted by entity permissions with `user_id = null AND role_id = null` instead of `role_id = 0`. 
- [x] Update joint_permissions table to support user_id?
- [ ] Update joint_permission handling. 
  - [x] Fix misalignment between joint and non-joint handling, Currently seen with 2-role user, granted view on parent book on Role A, prevented view on chapter on Role B. 
    - [x] Check scenario against release. 
      - Release has this same scenario. Works if inherit permissions is inactive, otherwise chapter (and content) is visible but not accessible, even if both roles are set to not allow view while inherit is active. 
  - [x] Chapter permission fallback "Other users" option not taking account.
  - [x] Check scenario of "Role A" granting permission via role permissions to sub-item that has blocked "Role B" permission at entity-level.
    - [x] Check on release - Can view (query), can't access (userCan).
    - [x] Check view via joint permission behaviour
    - [x] Check edit via `userCan` behaviour.
- [x] Add user interface for user permissions. 
- [ ] Delete entity permissions and joint permissions on user delete.
- [ ] Wrap gen/delete in transactions to avoid visibility change during generation. 
- [ ] Check for todos in changes.
- [ ] Testing.
   - [x] Create dev-doc for permission scenario testing, with test case IDs that link to PHPUnit test cases.
   - [ ] Test admin system role retains full item-level permission.
   - [ ] Manually & thoroughly scenario test permission application.
   - [x] Ensure view (joint permission) logic aligns.
   - [ ] Check prevention (Lack of permission) aligns with last release, and pre-v22.10 release (Where applicable), in that role permissions will prevent access where defined (Unless another role for the user specifically allows).
   - [ ] Check copying of shelf permissions.
   - [ ] Check copying of items (If permissions come across.)
- [ ] Extract text to language files.

### Docs update

- Update permissions user doc page.
  - Provide an "advanced" overview of deeper combination logic
- Add upgrade notice, linking to above added section for more details.


ssddanbrown:
### Note of approach failure

Written at 55642a33ee7b3221b760ff6ebf122d5f7e8a49c8

As part of this work I tried to bring the "cached" permissions, stored in the database, up a level so they're not per-role, per-item.
This was a `entity_permissions_collapsed` that would store the collapsed/flattened/inherited role & user permission overrides only where item-level permission overrides were active. Permissions were then joined into queries so the fallback, role & user status would be worked out at query time. 

As part of the permission checking we need to check owner user status.
Upon this, we'd also want to check page draft status (and its
creator/owner).
These, for cross-entity/relation queries would need up to another 4 joins (Totalling 7)
The performance/index usage was already questionable here.

This approach now looks like the wrong path, since we're implementing large changes, adding significant risk, without proof of overall benefit. The improved cache-time performance and reduced joint permission table would be risking query-time performance, which may not work out long-term. 
Think we should hold back on on changing the permission system in this way until we can zoom out and look at more fundamental potential changes to support (Aligning entities to single table). 

For now, might need to look to tack-on user permissions to existing system, but we should be able to keep changes in non-query work.

ssddanbrown:
Superseded by #3986, although in a different scope (No user permissions, mainly testing and alignment work)

Issue #3794, opened 2022-10-20
Link to Template Draft


kenswampy:
### Describe the feature you'd like

So I'd like a feature where I can have a link (say in the footer for example) where if a user clicks on it, then it brings them to a new page prepopulated with one of the templates I created and in a designated book. I understand users can simply click New Page in a book, and then select a template, but I was hoping to be able to have the option to create a new page that is already prepopulated with the selected template. 

### Describe the benefits this would bring to existing BookStack users

This would allow editors control over what type of page users can create. Or at least present authors with a template that they would otherwise have to delete the contents of to avoid using.

### Can the goal of this request already be achieved via other means?

If so, I'd love to know how. 

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundemental request

### How long have you been using BookStack?

0 to 6 months

### Additional context

I'm a tech writer currently using bookstack for a client as an intranet wiki and I'm hoping to get more coworkers to author content, so I may have some more ideas over time that would add value to end users.

ssddanbrown:
Thanks for the suggestion @kenswampy.
I've just merged the code to enable #1803, which will be part of the next feature release.
That issue has some overlap here, in that a template could be set as a default for a Book, so you could then achieve what you need just via sharing the page-create link for book with template assigned.

I appreciate it's not exactly the same though, but it may cover the use-case depending on the exact scenario.

ssddanbrown:
BookStack is now [managed on Codeberg](https://codeberg.org/bookstack/bookstack).

As part of this change, we're closing all existing GitHub issues and pull requests.
You'll instead find this issue on Codeberg here: 
https://codeberg.org/bookstack/bookstack/issues/3794
  
Any further discussion for this will now be done on Codeberg.
Therefore this GitHub thread will be locked.

Issue #1803, opened 2019-11-27
Default page templates


NikolayPolyagoshko:
**Describe the feature you'd like**
Page templates is great, and i think it would very useful if a template could be set as default one for new pages in a book (book-level setting).

**Describe the benefits this feature would bring to BookStack users**
Simplify creation of pages with predifined structure. E.g. problem-solution articles, etc.

cod3monk:
I would love this feature to be added, an intermediate solution (for my problem at hand) would be a create-page link that will auto insert a specified template.

**Describe the benefits this feature would bring to BookStack users**
I want to pass a link to a large crowd (of mostly non-techs) that will allow them to quickly create a page with a predefined structure. Having to explain how to insert a template significantly reduces the experience for them. Especially because it is not easy to explain in words the "Template Icon".

ssddanbrown:
Thanks to @lennertdaniels in #3918, followed by some further changes in #4721, this functionality has now been added and will be part of the next feature release.

Thanks @NikolayPolyagoshko for the original request, and thanks @cod3monk for providing an intermediate solution, even if we didn't get round to implementing in the mean time.

PR #3918, opened 2022-12-15
WIP: Default page templates

lennertdaniels:
I took a stab at implementing default page templates as indicated in #1803.

### What I have implemented:
- There's a book-level setting to pick a default template
- A new page automatically copies the template's html on creation
- A new guests page also has the template applied to it
- Deleted pages don't work as a template anymore
- When deleting a page that is used as a default template somewhere, there's a warning on the delete confirmation screen

image

image

### What I haven't fully considered or implemented yet:
- Any permission/scope problems
- If there's a difference in handling HTML vs. markdown templates
- A better selector to pick the default template (pagination, AJAX search) like the one in the book.edit sidebar (template-manager)
- Should other entities also support a default template? _(User, Shelf, Chapter)_ If so, which one has priority when multiple default templates are chosen on different levels?
- Applying a template from a URL parameter (@cod3monk and #3794)

### Random small fixes I came across:
- Autofocus on the name field on guest-create.blade
- Template Manager: fixed cancel icon placement

If anyone with more thorough knowledge of the codebase would like to shine some light onto this PR, that would be great.

### DIFF
--- a/app/Entities/Repos/PageRepo.php
+++ b/app/Entities/Repos/PageRepo.php
@@ -148,6 +148,12 @@ class PageRepo
             $page->book_id = $parent->id;
         }
 
+        if ($page->book->defaultTemplate) {
+            $page->forceFill([
+                'html'  => $page->book->defaultTemplate->html,
+            ]);
+        }
+
         $page->save();
         $page->refresh()->rebuildPermissions();
 
--- a/app/Http/Controllers/BookController.php
+++ b/app/Http/Controllers/BookController.php
@@ -6,6 +6,7 @@ use BookStack\Actions\ActivityQueries;
 use BookStack\Actions\ActivityType;
 use BookStack\Actions\View;
 use BookStack\Entities\Models\Bookshelf;
+use BookStack\Entities\Models\Page;
 use BookStack\Entities\Repos\BookRepo;
 use BookStack\Entities\Tools\BookContents;
 use BookStack\Entities\Tools\Cloner;
@@ -79,8 +80,14 @@ class BookController extends Controller
 
         $this->setPageTitle(trans('entities.books_create'));
 
+        $templates = Page::visible()
+            ->where('template', '=', true)
+            ->orderBy('name', 'asc')
+            ->get();
+
         return view('books.create', [
             'bookshelf' => $bookshelf,
+            'templates' => $templates,
         ]);
     }
 
@@ -98,6 +105,7 @@ class BookController extends Controller
             'description' => ['string', 'max:1000'],
             'image'       => array_merge(['nullable'], $this->getImageValidationRules()),
             'tags'        => ['array'],
+            'default_template'  => ['nullable', 'exists:pages,id'], 
         ]);
 
         $bookshelf = null;
@@ -151,7 +159,12 @@ class BookController extends Controller
         $this->checkOwnablePermission('book-update', $book);
         $this->setPageTitle(trans('entities.books_edit_named', ['bookName' => $book->getShortName()]));
 
-        return view('books.edit', ['book' => $book, 'current' => $book]);
+        $templates = Page::visible()
+            ->where('template', '=', true)
+            ->orderBy('name', 'asc')
+            ->get();
+
+        return view('books.edit', ['book' => $book, 'current' => $book, 'templates' => $templates]);
     }
 
     /**
@@ -171,6 +184,7 @@ class BookController extends Controller
             'description' => ['string', 'max:1000'],
             'image'       => array_merge(['nullable'], $this->getImageValidationRules()),
             'tags'        => ['array'],
+            'default_template'  => ['nullable', 'exists:pages,id'], 
         ]);
 
         if ($request->has('image_reset')) {
--- a/app/Http/Controllers/PageController.php
+++ b/app/Http/Controllers/PageController.php
@@ -74,7 +74,6 @@ class PageController extends Controller
         $page = $this->pageRepo->getNewDraftPage($parent);
         $this->pageRepo->publishDraft($page, [
             'name' => $request->get('name'),
-            'html' => '',
         ]);
 
         return redirect($page->getUrl('/edit'));
--- a/app/Http/Controllers/PageController.php
+++ b/app/Http/Controllers/PageController.php
@@ -3,6 +3,7 @@
 namespace BookStack\Http\Controllers;
 
 use BookStack\Actions\View;
+use BookStack\Entities\Models\Book;
 use BookStack\Entities\Models\Page;
 use BookStack\Entities\Repos\PageRepo;
 use BookStack\Entities\Tools\BookContents;
@@ -265,11 +266,13 @@ class PageController extends Controller
         $page = $this->pageRepo->getBySlug($bookSlug, $pageSlug);
         $this->checkOwnablePermission('page-delete', $page);
         $this->setPageTitle(trans('entities.pages_delete_named', ['pageName' => $page->getShortName()]));
+        $times_used_as_template = Book::where('default_template', '=', $page->id)->count();
 
         return view('pages.delete', [
             'book'    => $page->book,
             'page'    => $page,
             'current' => $page,
+            'times_used_as_template' => $times_used_as_template,
         ]);
     }
 
--- a/resources/lang/en/entities.php
+++ b/resources/lang/en/entities.php
@@ -328,6 +328,8 @@ return [
     'templates_replace_content' => 'Replace page content',
     'templates_append_content' => 'Append to page content',
     'templates_prepend_content' => 'Prepend to page content',
+    'default_template' => 'Default Page Template',
+    'default_template_explain' => "Assign a default template that will be used for all new pages in this book.",
 
     // Profile View
     'profile_user_for_x' => 'User for :time',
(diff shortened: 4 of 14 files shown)


lennertdaniels:
@ssddanbrown Would you mind taking a look if you find the time? Thanks.

ssddanbrown:
Hi @lennertdaniels,
Sorry for my lack of response so far, I plan to review pending PRs after finishing up the next release.
I remember doing a quick scan across the PR code when opened, and from what I can remember it was all looking really good, with maybe just needing some thought into ensuring the selection is scalable/efficient. Of course I need to look into this to properly review your code and considerations.
The one bit of feedback I will provide now is to not worry about any optional sub-features at this time (query param, other entities). I prefer to keep implementation/scope narrow and only expand on proven requirement/demand, to keep the platform surface area reasonable. 

Should review towards the end of the month unless things go further wrong (Current release cycle is being a challenge).

lennertdaniels:
@ssddanbrown absolutely fine, take your time. I just wasn't sure wether I had to add some tag/reviewer/assignee or something  to this PR. :)

I'll take your quick feedback into consideration and work on it a little more over the coming weeks.

app/Entities/Repos/PageRepo.php
```diff
@@ -148,6 +148,12 @@ public function getNewDraftPage(Entity $parent)
             $page->book_id = $parent->id;
         }
 
+        if ($page->book->defaultTemplate) {
+            $page->forceFill([
+                'html'  => $page->book->defaultTemplate->html,
+            ]);
+        }
+
```
ssddanbrown, line 156: This currently ignores permissions. I think we'll need to respect page permissions, so view access is required for the template to be used, otherwise other side-effects will be introduced (Image access permissions for example).

ssddanbrown, line 156: We do need to support markdown also here, and probably the original used editor.
I suspect it should just be a case of also copying the `markdown` and `editor` values from the original page model.

app/Http/Controllers/BookController.php
```diff
@@ -79,8 +80,14 @@ public function create(string $shelfSlug = null)
 
         $this->setPageTitle(trans('entities.books_create'));
 
+        $templates = Page::visible()
+            ->where('template', '=', true)
+            ->orderBy('name', 'asc')
+            ->get();
```
ssddanbrown, line 86: Right now, this'll be loading in all template pages, with all their properties (including html, markdown & text) into memory for use in a list. Probably fine for the majority of use-cases but I'm uncomfortable with the scalability of this.

It's a pain but ideally template selection would be done via front-end querying, much like what's shown when using the "Move" or "Copy" interfaces, to prevent requiring load of all options on form load.

app/Http/Controllers/BookController.php
```diff
@@ -98,6 +105,7 @@ public function store(Request $request, string $shelfSlug = null)
             'description' => ['string', 'max:1000'],
             'image'       => array_merge(['nullable'], $this->getImageValidationRules()),
             'tags'        => ['array'],
+            'default_template'  => ['nullable', 'exists:pages,id'],
```
ssddanbrown, line 108: Right now there's nothing stopping a user from setting a page template they don't have access to. 
We probably should have some light validation to check the user can view and that the template page is marked as a template. 

This exists validation check could also be abused to check the existence of a certain page id what the user may not have access to. Not sure how well we currently defend against that in other areas of the app, but probably would be something I'd attempt to avoid where possible.

app/Http/Controllers/BookController.php
```diff
@@ -151,7 +159,12 @@ public function edit(string $slug)
         $this->checkOwnablePermission('book-update', $book);
         $this->setPageTitle(trans('entities.books_edit_named', ['bookName' => $book->getShortName()]));
 
-        return view('books.edit', ['book' => $book, 'current' => $book]);
+        $templates = Page::visible()
+            ->where('template', '=', true)
+            ->orderBy('name', 'asc')
+            ->get();
```
ssddanbrown, line 165: As per create method comment.

app/Http/Controllers/BookController.php
```diff
@@ -171,6 +184,7 @@ public function update(Request $request, string $slug)
             'description' => ['string', 'max:1000'],
             'image'       => array_merge(['nullable'], $this->getImageValidationRules()),
             'tags'        => ['array'],
+            'default_template'  => ['nullable', 'exists:pages,id'],
```
ssddanbrown, line 187: As per create method comment.

ssddanbrown, line 187: Currently, if the template option is removed when editing a book the column value will end up as `0`. Since nulls are supported in this column, i think the value should end up `null` instead.

app/Http/Controllers/PageController.php
```diff
@@ -266,11 +266,13 @@ public function showDelete(string $bookSlug, string $pageSlug)
         $page = $this->pageRepo->getBySlug($bookSlug, $pageSlug);
         $this->checkOwnablePermission('page-delete', $page);
         $this->setPageTitle(trans('entities.pages_delete_named', ['pageName' => $page->getShortName()]));
+        $times_used_as_template = Book::where('default_template', '=', $page->id)->count();
```
ssddanbrown, line 269: Variable name does not follow project conventions, but use may be redundant (lang file comments).

resources/lang/en/entities.php
```diff
@@ -192,6 +192,7 @@
     'pages_delete_draft' => 'Delete Draft Page',
     'pages_delete_success' => 'Page deleted',
     'pages_delete_draft_success' => 'Draft page deleted',
+    'pages_delete_warning_template' => '{0}|{1}Be careful: this page is used as a template for :count book.|[2,*]Be careful: this page is used as a template for :count books.',
```
ssddanbrown, line 195: I don't think we need to show the count here (Which again can complicate matters when permissions are in play) but I just think we need the user to be aware of the consequences, something like:

> This page is in active use as a book default page template. Such books will no longer have a page default template assigned after this page is deleted.

resources/lang/en/entities.php
```diff
@@ -328,6 +329,8 @@
     'templates_replace_content' => 'Replace page content',
     'templates_append_content' => 'Append to page content',
     'templates_prepend_content' => 'Prepend to page content',
+    'default_template' => 'Default Page Template',
+    'default_template_explain' => "Assign a default template that will be used for all new pages in this book.",
```
ssddanbrown, line 333: This should probably include a note on permissions, advising that the page template will only be used if the page creator has view permission for the page template.

resources/views/entities/template-manager.blade.php
```diff
@@ -0,0 +1,10 @@
+<p class="text-muted small">
+    {!! nl2br(e(trans('entities.default_template_explain'))) !!}
```
ssddanbrown, line 2: There's no newlines in the source string so this can be a simple non-escaped template tag. If there are multiple clear lines I try to split the translation itself these days.

resources/views/pages/parts/template-manager.blade.php
```diff
@@ -14,7 +14,7 @@
     <div class="search-box flexible mb-m" style="display: {{ count($templates) > 0 ? 'block' : 'none' }}">
         <input refs="template-manager@searchInput" type="text" name="template-search" placeholder="{{ trans('common.search') }}">
         <button refs="template-manager@searchButton" tabindex="-1" type="button">@icon('search')</button>
-        <button refs="template-manager@searchCancel" class="search-box-cancel text-neg" type="button" style="display: none">@icon('close')</button>
+        <button refs="template-manager@searchCancel" class="search-box-cancel text-neg" tabindex="-1" type="button" style="display: none">@icon('close')</button>
```
ssddanbrown, line 17: I'd generally advise slipping small fixes into larger works otherwise they get blocked by the larger feature and/or conflict if the fix gets applied in the meantime.

ssddanbrown (changes requested):
Thanks for this @lennertdaniels, All was clear and seemed to function well. 
Just raised my thoughts/findings from a deeper dive into the code.
Feel free to query or ask for clarification/assistance/guidance for any of my notes.

Ideally we'd have testing to cover the functionality additions here also, at least to cover the main functionality. Happy to help or advise on that inf you're unfamiliar with testing.

ssddanbrown:
Hi @lennertdaniels,
Would you be looking to address the review points above yourself? Or do you need a certain level of support/advise? Or would you prefer for me to carry this on from this point?

lennertdaniels:
Hi @ssddanbrown,

I'm slowly chipping away at your remarks whenever I find some spare time :) Will update the PR in due time.
The feedback was very useful, thanks!

melat0nin:
Hi @lennertdaniels, have you managed to make any progress on this PR? It's a feature I'd really love to have in BookStack :)

ssddanbrown:
Hi @lennertdaniels,
Would you be okay for me to take this on and finish things off from this point?

lennertdaniels:
@ssddanbrown please do! 
I wasn't able to dedicate a lot of time to this PR, sorry. I'm excited to see this feature come alive :)

ssddanbrown:
Alright, I'll assign this to take on for the next release.
Please keep your fork & branch, and this PR up and active in the meantime until this is merged.

> I wasn't able to dedicate a lot of time to this PR, sorry.

No reason to be sorry, thanks for contributing and getting this started!

ssddanbrown:
Okay, now all merged with follow-up work continued in #4721.

Thanks again for getting this started @lennertdaniels!

Issue #3946, opened 2023-01-05
Remove use of "admin.com" default admin user email


ghost:
### Describe the feature you'd like

Instead of using `admin.com` as default domain better us `example.com` or `example.org`. The domain `admin.com` resolves to a privately owned website whereas `example.com` and `example.org` are special domains, which are valid and provided by ICANN for testing purposes.

When a user logs in the first time on a fresh install of BookStack with the default admin credentials and tests the email settings one will successfully send an email but also get a bounce from `admin@admin.com`. This is bad for the reputation of the email server (either self hosted or third party service).

To solve this issue it's a better solution to use `admin@example.com` because when sending emails to this address the email will get successfully send and delivered.

### Describe the benefits this would bring to existing BookStack users

Testing email delivery on a fresh install will not result in a bounced email.

### Can the goal of this request already be achieved via other means?

Changing the default mail address of the admin user to a personal mail address.

This feature request will prevent bounced emails even if a user forgets to change the mail address before testing email settings.

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundemental request

### How long have you been using BookStack?

1 to 5 years

ghost:
Issue #602 requests similar change but does not mention the disadvantages of using `admin.com`.

timptner:
Pull-Request #3947

ssddanbrown:
Thanks for raising @timptner.
I'm going to close this in favour of centralising things in #4575, where a plan is put forward and tracked against for moving away from the existing default credential handling.

PR #3947, opened 2023-01-05
Update default user table migration

ghost:
Change default admin email to a valid domain for testing purposes provided by ICANN.


ghost:
Solves issue #3946.

ssddanbrown:
Thanks for offering this change @timptner but I'm going to close this off without merge.
Change of the default credentials will cause a lot of existing documentation & processes to become incorrect/outdated & in need of update, It would effectively be a breaking change in my view that requires some co-ordination and warning.

I do agree it would be good to move away from the existing credentials, but I don't think the reasons in #3946 are worth it alone. If we are to make changes to the default credential process, I'd like to do so as part of more substantial changes (An initial setup process or randomly generated default creds or something similar, needs some thought).

Issue #1663, opened 2019-09-13
Help buttons/links in markdown editor


ethanmdavidson:
**Describe the feature you'd like**
On the markdown editor, I would like to have a link to some markdown help. It would also be nice to have some buttons which insert the markdown syntax for you.

Two examples of what I mean:

Gitlab markdown editor
![gitlabMarkdownButtons](https://user-images.githubusercontent.com/1311998/64886843-268e9500-d635-11e9-9594-5298703933c5.PNG)

Github markdown editor
![githubEditor](https://user-images.githubusercontent.com/1311998/64886902-3dcd8280-d635-11e9-9cc5-46d9a1ac0610.png)

**Describe the benefits this feature would bring to BookStack users**
For new users, the buttons would greatly ease the learning curve because the buttons provide a familiar UI, and the documentation would be right at their fingertips. It would also help more advanced users because it can be hard to remember all the different markdown syntax. The [existing keyboard shortcuts](https://www.bookstackapp.com/docs/user/markdown-editor/) are nice, but without a link to the documentation most people will never know they exist.

**Additional context**
This is related to #1661

berot3:
Yes please! For me bookstack is not an option until the WYSIWYG-editor stores the pages as markdown and not as html.
It would be a "work-around" to use the markdown-editor. But than the markdown-way is not usable until at least there are some helper-buttons for editors.

Technosoft2000:
👍 +1 to see this feature in future versions :)

Logician724:
:+1: Really need this feature. Cannot adopt markdown as the default editor for all users if only those proficient in it can use it. Or at least provide the option for different users to use different editors.

TheNomad11:
Yes, please, or more shortcuts at least

ssddanbrown:
> Yes, please, or more shortcuts at least

@TheNomad11 The markdown editor does support a range of shortcuts as listed here: https://www.bookstackapp.com/docs/user/markdown-editor/

Zazou49:
A toolbar will be very usefull for beginner !

EmberLightVFX:
Big +1 on having a toolbar for markdown!

I write markdown way faster than using the WYSIWYG editor and it's very easy to use markdown written somewhere else and just paste it into the editor and bam, you're done.

But I sometimes don't remember specific markdown commands like how to create tables or strike-throughs. I then have to go to https://www.markdownguide.org/ and search there until I find what i was looking for.
[...]

PR #3952, opened 2023-01-07
Feature: Add styling options to Markdown toolbar

phsteffen:
We have the use-case that we are using the Markdown editor with technical as well as non-technical users.  
The non-technical users sometimes find it a bit difficult to use the Markdown markup.

This pull request adds buttons for the following markups in the Markdown editor:

- Headline
- Bold
- Italic
- Strikethrough
- Bulleted List
- Numbered List

Additionally it adds a kind of snippet function to insert the following things with it:

- Table
- Collapsible Sections
- Callouts (Info, Success, Warning, Danger)

**The whole thing looks like this:**

markdown_toolbar_1

**Dropdown with snippets:**

markdown_toolbar_2

**In small resolutions, however, the buttons are hidden:**

markdown_toolbar_3

_What do you think of the solution?_  
_Should I adjust/remove something?_


ssddanbrown:
Thanks for offering this PR @phsteffen.
To be honest though, without diving specifically into the technical implementation at all, I'm not totally sure about the idea of including this as a feature.

Our primary focus for non-technical users has always been the WYSIWYG editor, with the markdown editor available to those technical enough to prefer it. I originally built BookStack as a platform to avoid non-technical users having to understand/learn specific syntax. This changes the scope an intention of the markdown editor from being technical-user focused to also, along with WYSIWYG editor, being all-technical-skill focused. I'm just not sure on that re-focusing being a good idea and/or worthwhile.  While the surface area of the PR isn't that wide as-is, It still has an impact to maintenance and the expectations of the issues and features we'd further support. In addition, if we go down this path I feel there'd be an expectation that the controls and usability matches that provided in the WYSIWYG editor, in which case quite a few more controls would need to be added to align. 

To help my understand the requirement better: for your use-case, what requires the need for the markdown editor rather than just using the WYSIWYG editor?

---

This PR related to #1663 but the state of that issue is somewhat different nowadays, since we added the ability to jump between editors.

EmberLightVFX:
I would love to have this toolbar added to the markdown editor.
I write markdown way faster than using the WYSIWYG editor and it's very easy to use markdown written somewhere else and just paste it into the editor and bam, you're done.

But I sometimes don't remember specific markdown commands like how to create tables or strike-throughs. I then have to go to https://www.markdownguide.org and search there until I find what i was looking for.
Having these added to the toolbar as shortcuts would help a lot for things the user don't remember.

gauthiersiri:
I'd like to see a toolbar like this :)
With my little brain i sometimes forget the syntax too :)
I agree that i could use the WYSIWYG editor, but more and more people use markdown, even beginner, so it would help.

And on page with different author, it would/may avoid, switching from MD to WYSIWYG to MD to WYSIWYG etc...

My 2 cents :)

ssddanbrown:
I'm going to go ahead and close this off as per my last message.

@EmberLightVFX @gauthiersiri Thanks for your input. It'd be worth copying your thoughts to #1663, to help indicate desire for such a feature and to provide context as we can think about various solutions to suit needs. Sounds like just having a simple built-in syntax reference guide might meet most of your needs while having a relatively small maintenance impact on the project.

EmberLightVFX:
Too bad it got closed and not merged. Added my thoughts to the issue you linked and hoping it will get added later down the line.

PR #3986, opened 2023-01-21
Permission Testing & Alignment

ssddanbrown:
This is a merge of the permission work done in #3908 , at commit 3083979855c6395fd7a35fb92790e0efb37b44f6, but only of the testing work/changes since the fundamental user permissions work came to a barrier. 

### Docs update

- Update permissions user doc page.
  - Provide an "advanced" overview of deeper combination logic
- **Upgrade notice**, linking to above added section for more details, advising slight change in logic handling.
- **Upgrade notice**, potential longer db migration time on upgrade due to permission regeneration.


ssddanbrown:
Merging in now so things can be passively tested during other developments.
Will continue manual testing here and there on `development` up to release.

PR #4032, opened 2023-02-08
Generate favicon.ico file

ssddanbrown:
Followed wikipedia-defined ICO file format info, and used with Intervention's good bmp support, to create a working proof-of-concept. Just need to finish off and integrate.

#### Todo

- [x] Todos in code comments.
- [ ] Would be ideal to confirm the incoming bitmaps will always be 32bit-per-pixel.


ssddanbrown:
Looks like phpgd only supports 8 or 24bit bmp images, not 32bit with alpha: https://github.com/libgd/libgd/blob/d11db230d3e8c48630e0abe24e34f0f12f45d989/src/gd_bmp.c#L210

Next step is to trial png-in-ico support.

ssddanbrown:
Changed to png based ICO files, test on Firefox, Chrome, Gnome Web, seems to work nicely.
Now there's a challenge of how to handle the `favicon.ico` file in regards to git. Ideally want it to exist as a default but don't want it to show as an untracked change upon BookStack icon change.
Probably need to remove from repo by default then copy across default icon on access when no icon exists already.

Issue #3771, opened 2022-10-08
Support updating of user details upon OIDC/SAML2/LDAP Login


danicc097:
### Describe the Bug

When user info is changed in the identity provider, changes are not reflected in the bookstack database. Only groups are synced at the moment AFAICT:

https://github.com/BookStackApp/BookStack/blob/a090720241ed3a0471a4dc9860b9f7cf8837f562/app/Auth/Access/Oidc/OidcService.php#L250-L265

### Steps to Reproduce

1. Register a new user in the IDP.
2. Login in Bookstack
3. Update the user in the IDP
4. Logout of Bookstack, clear cache and log and back in
5. Changes not reflected

### Expected Behaviour

I'd expect changes from the IDP to be reflected automatically

### Exact BookStack Version

22.09

### Hosting Environment

Ubuntu 20.04, docker image

ssddanbrown:
Yeah, this is the case with all external auth options. I'm gonna re-categorise this as a feature request since it's not specifically a bug in existing supported functionality.

Misaka-L:
@ssddanbrown Sorry to disturb, but this issue has been hanging on for 2 years and I want to ask if there is any progress on it?

ssddanbrown:
@Misaka-L No, as above I see this as a feature request rather than an issue, so I'm not really aiming to address but more awaiting to validate as a need to justify the added desired complexity, for which there has been very little demand so far.

ssddanbrown:
BookStack is now [managed on Codeberg](https://codeberg.org/bookstack/bookstack).

As part of this change, we're closing all existing GitHub issues and pull requests.
You'll instead find this issue on Codeberg here: 
https://codeberg.org/bookstack/bookstack/issues/3771
  
Any further discussion for this will now be done on Codeberg.
Therefore this GitHub thread will be locked.

PR #4060, opened 2023-02-22
feature: update user on login

rolandspannagl:
To simplify user-management this pull request implements a feature, which updates the user's display name and email on log-ins using SAML2.

It is enabled using a new config-parameter: _SAML2_UPDATE_USER_ON_LOGIN_.


app/Auth/Access/RegistrationService.php
```diff
@@ -72,6 +73,13 @@ public function findOrRegister(string $name, string $email, string $externalId):
             ];
 
             $user = $this->registerUser($userData, null, false);
+        } else if ($updateUser && ($user->name !== $name || $user->email !== $email)) {
+            $user->update([
+                'name'  => $name,
+                'email' => $email,
+            ]);
+
```
ssddanbrown, line 81: Could this logic be moved to it's own method within this class instead? 
Passing a flag argument to trigger extra logic starts to make this method a little muddy, would prefer a focused method that's optionally called (depending on setting) from the Saml2service side.
If you want to specifically only update new instances I think there might be something within the Eloquent model for that, or you could resort to checking the created date.

app/Auth/Access/RegistrationService.php
```diff
@@ -72,6 +73,13 @@ public function findOrRegister(string $name, string $email, string $externalId):
             ];
 
             $user = $this->registerUser($userData, null, false);
+        } else if ($updateUser && ($user->name !== $name || $user->email !== $email)) {
+            $user->update([
+                'name'  => $name,
+                'email' => $email,
```
ssddanbrown, line 79: The email should be user-unique so it might be wise to have a check before updating, but there becomes a question of what to do on error.

There's also a question of how this interacts with email confirmation and domain restriction. 
Don't want to complicate things thought so it might just be that those are left alone but we ensure there are warnings against the setting documentation about email confirmation/domain restriction.

app/Auth/Access/RegistrationService.php
```diff
@@ -72,6 +73,13 @@ public function findOrRegister(string $name, string $email, string $externalId):
...
+            ]);
+
+            Activity::add(ActivityType::USER_UPDATE, $user);
```
ssddanbrown, line 82: Please could this be removed?
Activities generally reflect user actions or high-level processes, rather than raw changes to data. 
While the user data is technically being updated, no one is (intentionally) actioning a user update themselves.

ssddanbrown (changes requested):
Thank you for offering this PR.
I've reviewed the changes and added some comments.
If you have any questions, or need any help, just shout.

In addition to those, it would also be ideal to have this added functionality covered by testing. Would you be willing to add testing to cover your changes?

ssddanbrown:
This is related to #3771. Ideally we'd align functionality across these methods, but SAML2 can be a starting point.

app/Auth/Access/RegistrationService.php
rolandspannagl: Is there really the need for email-verification in registrations from SSO-Services? In my oppinion this information must be verified by the identity-prodiver. I would suggest that in such cases a verified address is assumed.

rolandspannagl: What do you think about removing email confirmation/domain restriction from SSO-Logins in general?

ssddanbrown: > What do you think about removing email confirmation/domain restriction from SSO-Logins in general?

I am not willing to remove a feature that could be in use, just in the interest of making a PR easier to merge.
We can ignore these specific elements though and, as mentioned above, we'd just need to ensure these edge cases are in the documentation for the setting.

rolandspannagl: This was not meant for making the PR easier to merge. I thought this would be a solution for preventing an inconsistent email-confirmation-feature in SAML-auth rather just "ignoring" it.

Now the question is how to resolve this, while maintaining consistency in function. How about a new config-parameter for email-domain-restrictions in SAML (and another parameter for every other auth-method)? So you can use domain-restriction on local-auth and disable it on saml-auth.

ssddanbrown: As said above, I'd advise we ignore these specific elements in functionality and defer to documenting the implications instead, rather than increasing the surface area and scope of the PR.

ssddanbrown:
Since there's been recent follow up or activity I'll close this off.
Feel free to still comment or update though if there is progress and this can be re-opened if needed.

PR #4084, opened 2023-03-03
Admin CLI build

ssddanbrown:
Playing around with building a BookStack CLI for system-level tasks (Backup, Restore, Install, Update, etc..).
Building as a separate component to the BookStack codebase/laravel, compilable to a single phar, so that it function fully independently, even if the main app is in a broken state.


ssddanbrown:
Development moved to https://github.com/BookStackApp/system-cli

PR #4123, opened 2023-03-22
Clone book to multiple shelves feature

tanvir-simformsolutions:
Added feature to clone/copy book to shelves with following
 - Copy book to multiple shelves 
 - Added translation for all the available languages
 - Create new blade component for re-usability purpose
 - Create new function for re-usability purpose

### DIFF
--- a/app/Entities/Tools/Cloner.php
+++ b/app/Entities/Tools/Cloner.php
@@ -68,11 +68,46 @@ class Cloner
      * Clone the given book.
      * Clones all child chapters & pages.
      */
-    public function cloneBook(Book $original, string $newName): Book
+    public function cloneBook(Book $original, string $newName, string $newShelves = null): Book
     {
         $bookDetails = $this->entityToInputData($original);
         $bookDetails['name'] = $newName;
 
+        // Clone bookshelf relationships
+        /** @var Bookshelf $shelf */
+        if ($newShelves == null) {
+
+            // Clone book content
+            $copyBook = $this->cloneBookContent($original, $bookDetails);
+
+            foreach ($original->shelves as $shelf) {
+                if (userCan('bookshelf-update', $shelf)) {
+                    $shelf->appendBook($copyBook);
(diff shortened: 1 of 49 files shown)


ssddanbrown:
Thanks for offering this feature @tanvir-simformsolutions, but I'm not sure there's a proven need for this? Just to decide if the value of addition, to the existing user-base, is worth the complexity added. I'm just not sure there's a frequent use-case to _copy_ a book into multiple new parent shelves. 

In addition to that, I'm also unsure of how this presents the shelf to book relationship to a user.  I'd worry this adds confusion, potentially leading users to think this is the primary way of getting a book in multiple shelves, enforcing the (incorrect) assumption that a book belongs to a single shelf.

tanvir-simformsolutions:
Thanks for noticing the feature PR @ssddanbrown, it's up to you to add this feature or not.

**The reason why i have added this feature,**

1) When creating a new shelf we have an option to add the existing book to it but it does not create new copy of that book for the new shelf but it gives reference of that book to this shelf

2) When we do COPY/CLONE book, we can only CLONE/COPY the book to same shelf. I have added this so that we can COPY/CLONE the book from one shelf to other shelf as well. (To single shelf of multiple shelves)

3) Please check this screenshot of what exactly i have added - https://monosnap.com/direct/pMN4sWgrbs9KFKN2Evxh4GjIOh8HkG

ssddanbrown:
@tanvir-simformsolutions Sure, I understand what this does, It's just that I'm not sure this specific scenario is a common enough scenario to be worth the added complexity, especially since this isn't achieving anything that wasn't possible before, but instead just speeding up the process. 

Therefore I'm going to close this off. Thanks though for offering the PR. 
Just in case you specifically required this to be part of the core project, I am fairly strict in regards to what gets added to the platform, to keep maintenance feasible. If you're worried about wasting time on a PR, please start a discussion via an issue first.

PR #4126, opened 2023-03-24
Allow a user to disable peer check when using TLS/STARTTLS

### LINKED ISSUE #3166: [Feature Request]: Allow disable SMTP certificate verification

vincentbernat:
This is useful when developing and on Docker setups. Despite setting encryption to null, if a server supports STARTTLS with a self-signed certificate, the mailer try to upgrade the connection with STARTTLS.

Fix #3166

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -79,6 +79,7 @@ MAIL_PORT=1025
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
 MAIL_ENCRYPTION=null
+MAIL_VERIFY_PEER=true
 
 # Command to use when email is sent via sendmail
 MAIL_SENDMAIL_COMMAND="/usr/sbin/sendmail -bs"
--- a/app/Config/mail.php
+++ b/app/Config/mail.php
@@ -32,6 +32,7 @@ return [
             'encryption' => env('MAIL_ENCRYPTION', 'tls'),
             'username' => env('MAIL_USERNAME'),
             'password' => env('MAIL_PASSWORD'),
+            'verify_peer' => env('MAIL_VERIFY_PEER', true),
             'timeout' => null,
             'local_domain' => env('MAIL_EHLO_DOMAIN'),
         ],


ssddanbrown:
Thanks for offering this @vincentbernat,

Could you confirm this has change has been tested in a real actual use here, where the environment would error without this change? I just want to be sure we're adding what's needed here (Since there are many options in PHP relating to possible mail/TLS connection verification) and that this is not being added without being sure to cover the scenario.

I'll also need to test the scenario myself so just want to confirm this has been tested before I spend a lot of time to replicate a similar scenario. 

Additionally, it would be ideal if we could have some level of testing in place within the codebase, if possible, to ensure the mailer at least receives these options, so that our testing would alert if the config API changes like it did for Laravel 9.

vincentbernat:
I have tested it myself. Our internal mail server exposes STARTTLS and use a self-signed certificate. So, it's working. I am using the Docker setup (so I suppose with Laravel 9).

ssddanbrown:
Thanks again @vincentbernat, now merged.

I followed this up with 1de72d09ca22878d02496b40f2be516adba5e736, just to tweak the name to be a bit more generic so we can bundle additional ssl/tls-verification-disabling options if required in future, and to add some testing to ensure the option affects the underlying mailer.

PR #4145, opened 2023-03-30
Add WKHTMLTOPDF to .env.example.complete

7nohe:
The WKHTMLTOPDF environment variable was missing from the .env.example.complete file.
This pull request adds the variable to the file with a commented example path to the wkhtmltopdf binary.


ssddanbrown:
Thanks @7nohe!

PR #4148, opened 2023-03-31
Allow roles to be an array

the-voidl:
Our oidc authentication endpoint (ZITADEL) does not return roles as simple array but single arrays per given role.

This pull request contains a simple fix to use the so given role names (keys of subarrays) as expected by BookStack.

An example payload of from ZITADEL looks like this...:

```
{
  "iss": "https://oicd.example.com",
  "aud": [
    "823417009781275377@oicd",  
    "593371374829733854@oicd",
    "298754342231354326@oicd",
    "207625234567516721@oicd",
    "111111111111111111"
  ],
  "azp": "298754342231354326@oicd",
  "at_hash": "h4ivntmqlr3v43_svT",
  "c_hash": "iv43lw34n4312A7af$_Vzc",

  "amr": [
    "password",
    "pwd",
    "mfa",
    "user"
  ],
  "exp": 1680264155,
  "iat": 1680260555,
  "auth_time": 1680247938,
  "email": "my.email@example.com",
  "email_verified": true,
  "family_name": "Name",
  "given_name": "GivenName",
  "name": "GivenName Name",
  "nickname": "GivenName Name",
  "preferred_username": "my.email@example.com",
  "sub": "111111111111111111",
  "updated_at": 1680255197,
  "urn:zitadel:iam:org:project:roles": {  <--- groups scope
    "admin": {                            <--- subarray for role admin: instance=12345, issuer=oicd.example.com
      "12345": "oicd.example.com"
    },
    "user": {
      "12345": "oicd.example.com"
    } 
  } 
}
```

### DIFF
--- a/app/Auth/Access/Oidc/OidcService.php
+++ b/app/Auth/Access/Oidc/OidcService.php
@@ -186,6 +186,11 @@ class OidcService
             return [];
         }
 
+        // Use keys of nested arrays as values here
+        if (gettype(reset($groupsList)) === "array") {
+            $groupsList = array_keys($groupsList);
+        }
+
         return array_values(array_filter($groupsList, function ($val) {
             return is_string($val);
         }));


ssddanbrown:
Thanks for offering this @the-voidl. Even though this is a relatively small change, I'm very much hesitant to increase the scope of what we support in terms of auth, especially where it's to suit one-off providers/scenarios so, before looking at implementation, I'd just want to be sure this is the required and the correct direction to go.

From my quick search about [I came across this Zitadel issue](https://github.com/zitadel/zitadel/issues/3997) which looks like is covers providing groups in the common string-array format? and has recently been closed so may now be possible in recent Zitadel versions? But I'm not familiar with zitadel at all so can't advise for sure.

If this isn't possible from the Zitadel end, I'm wondering if it may be better to instead create a new [logical theme system event](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md) to allow customization of the incoming OIDC token data as a way of allowing any possible mapping/handling that may arise in the future, instead of us needing to support specific edge-cases in the core project.

the-voidl:
I understand your point and I support you in the thinking that, especially in the manner of authorization, we have to be careful.

Maybe the logical theme system would be the right place for things like that that to not confuse things in the core project :+1: 

Thanks for finding that Zitadel issue. In our case it solves exactly that problem.

PR #4157, opened 2023-04-04
Add support to import docx file for page creation

shankar-d-simform:
- Added a feature to import DOCX files for page creation. We will count the number of heading tags (H1) to determine the page titles, and the content within each tag's child elements will become the respective page's content. For example, if there are 7 heading tags, we will create 7 pages for the corresponding book.

- When editing a book, if we upload a new DOCX file, we have two options: create new pages or append to the existing ones. If we select 'create new pages', we will delete all the existing pages of the respective book and create new ones. If we choose the second option, we will simply append the new pages to the existing ones.

image

image

### DIFF
--- a/app/Actions/ActivityType.php
+++ b/app/Actions/ActivityType.php
@@ -62,6 +62,8 @@ class ActivityType
     const WEBHOOK_CREATE = 'webhook_create';
     const WEBHOOK_UPDATE = 'webhook_update';
     const WEBHOOK_DELETE = 'webhook_delete';
+    const BOOK_PAGES_DELETE = 'book_pages_delete';
+
 
     /**
      * Get all the possible values.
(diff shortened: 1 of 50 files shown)


ssddanbrown:
Thanks for offering this @shankar-d-simform but, while I'm sure this may be useful to some folks, I'm not looking to extend the scope of what we maintain/support in the core project to third-party complex formats.

Eikir:
In my opinion this is a really nice feature. Is any plugin system planned, so these features can be part of a third party plugin?

ssddanbrown:
@Eikir No plugin system yet (maybe in long term future) but we have various methods of hackery that something like this could be built using: https://www.bookstackapp.com/docs/admin/hacking-bookstack/

Alternatively something could be built externally to use the BookStack API [like this script here](https://github.com/BookStackApp/api-scripts/tree/main/node-docx-to-page).

As per my thoughts in my previous comment, and since there's been no further follow-up on that, I'm going to close this off.

PR #4191, opened 2023-04-24
Fix timestamp in API docs example response

tigsikram:
Hello! The `created_at` timestamp in the API docs seems to be malformed. I inserted the value from `chapters-read.json`.


ssddanbrown:
Thanks @tigsikram!

Issue #2845, opened 2021-07-14
Support nested levels/use of content include tags


sid3windr:
**Describe the bug**
Multiple layers of page inclusion are not being evaluated correctly.

**Steps To Reproduce**
1. Create page A with some content.
2. Create page B with some content, and include page A in it. This will display correctly.
3. Create page C with some content, and include page B in it.
4. Page C will have the page C content from step 3, and the page B content from step 2, followed by the {{@123}} inclusion tag for page A, instead of page A's content.

**Expected behavior**
All pages to be included should be displayed correctly.

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): 21.04
 - PHP Version: 7.3.29
 - Hosting Method (Nginx/Apache/Docker):  Apache

ssddanbrown:
Related to #1601

robertfshort:
+1 for this feature

JoeIzzard:
+1 for this

jasonF1000:
+1 for this feature

deleyva:
+1 for this feature

jasonF1000:
Hi,

i think a depth of 3 layers for the page inclusion should be enoght for the most cases.
Is there any implementation Plan in the near Future for this Feature?

Regards,
jason

ssddanbrown:
Up to three levels of nested has now been added within the work of PR #4192, thanks to @jasonF1000.
Will be part of the next feature release.

PR #4192, opened 2023-04-24
changed PageContent.php to accept nested includes

jasonF1000:
tested and it works. 
We have implemented levels depths that can be configured as needed. ( via includeDepth variable)
This solves issue #2845

### DIFF
--- a/app/Entities/Tools/PageContent.php
+++ b/app/Entities/Tools/PageContent.php
@@ -301,13 +301,13 @@ class PageContent
             $content = HtmlContentFilter::removeScripts($content);
         }
 
-        if ($blankIncludes) {
+        if ($blankIncludes) {
             $content = $this->blankPageIncludes($content);
             } else {
-                for ($includeDepth = 0; $includeDepth <= 3; $includeDepth++){
-                    $content = $this->parsePageIncludes($content);
+                for ($includeDepth = 0; $includeDepth <= 3; $includeDepth++) {
+                    $content = $this->parsePageIncludes($content);
                 }
-            }
+        }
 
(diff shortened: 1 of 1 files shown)


jasonF1000:
hi @ssddanbrown, could you please review this pull request?

Thanks in Advance.

ssddanbrown:
Thanks @jasonF1000 for offering this PR. A few of things:

- The code provided seems to have whitespace issues, mainly in that it seems to be using non-breaking spaces instead of normal spaces within the indentation, which breaks this as code.
- This loop only needs to be around `parsePageIncludes`, not ` blankPageIncludes` too. 
- Ideally I'd want to have testing added to cover this scenario (With some depth checking to ensure this sticks to three levels max).

There are existing tests to cover page includes in this file:
https://github.com/BookStackApp/BookStack/blob/development/tests/Entity/PageContentTest.php
Would you be comfortable in adding the requested testing?

### Note-to-self

If merged, this will need to be documented as a breaking change.

jasonF1000:
Hi @ssddanbrown i followed your recommendation and fixed the space issue and change the loop that it is only around the parsePageIncludes.

But i feel not so comfortable to add the testings, therefore i have to low experience with PHP. Sorry :(

ssddanbrown:
Thanks @jasonF1000, now merged for next release. 
I followed with up with 56f234d1ee046a84da163aaaa5048a2463dbabb1 to adjust formatting and add testing.

jasonF1000:
@ssddanbrown thank you for taking over the testing part and for the implementation of the feature :)

PR #4197, opened 2023-04-25
.vscode folder is added in .gitignore file

esakkiraja100116:
- `.vscode` folder added in `.gitignore` file.


ssddanbrown:
@esakkiraja100116 Happy to accept the change to the `.gitignore` file since vscode is popular.

I would not look to accept the `.env.example` change though, since that file is used for production/release workflows as a base for the user env. Development mode should be opt-in, not a default.

esakkiraja100116:
okay understand. Now I removed that line from `.env.example` file.

esakkiraja100116 (commented):
`.vscode ` folder is added here!

ssddanbrown:
Thanks!

PR #4204, opened 2023-04-27
Added max-width to videos to prevent them escaping the page content area

snowc0de:
This is to prevent that videos included in pages don't exceed the page border. This is to avoid the following issue: 
image

Which now becomes this:

image

I didn't build the project and only added the CSS code into my instance's "head config" but as far as I know this fix should work.

### DIFF
--- a/resources/sass/_pages.scss
+++ b/resources/sass/_pages.scss
@@ -182,6 +182,10 @@ body.tox-fullscreen, body.markdown-fullscreen {
   .cm-editor {
     margin-bottom: 1.375em;
   }
+
+  video {
+    max-width: 100%;
+  }
 }
 
 // Page content pointers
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering. Setting a `width` could cause further issues though (videos with pre-defined sizes, videos smaller than the page width). Instead, a `max-width` may be a better approach, within our [page content](https://github.com/BookStackApp/BookStack/blob/8cd6c797e8a1b80de01bf0ddc0645e3c4765ff5e/resources/sass/_pages.scss#L79) style block so its clear this rule relates to video in page content (Rather than in the CSS reset).

snowc0de:
Thanks for the feedback @ssddanbrown, I just made the change :)

ssddanbrown:
Thanks for updating @chopin2712. Will merge for next release.

---

_Weakly related to #1760_

JNR8:
Did this get implemented?  My videos till break page width on mobile devices.

PR #4259, opened 2023-05-23
Update docker-compose.yml

AlFcl:
add 
 APP_TIMEZONE=UTC   for time problem


ssddanbrown:
Thanks for offering this but it's not clear what this is attempting to fix, and this compose setup is intended for development only. If you need to add extra vars you should be able to via your `.env` file.

Issue #4290, opened 2023-06-08
API returns HTTP 500 instead of 404 on not found


devdot:
### Describe the Bug

When requesting a non-existing page, the API returns HTTP 500 instead of 404. The JSON looks like this:

```json
{"error":{"message":"Page not found","code":500}}
```

When I read the documentation (https://demo.bookstackapp.com/api/docs#error-handling) I assumed it meant that the API returns matching HTTP statuses when possible. 

After clarifying on discord (thanks @ssddanbrown !), this behavior is not expected.

### Steps to Reproduce

1. Make an authorized API call to a non-existing page (example: https://demo.bookstackapp.com/api/pages/5000)
2. Receive HTTP 500 instead of 404

### Expected Behaviour

HTTP 404 Not Found

```json
{"error":{"message":"Page not found","code":404}}
```

### Exact BookStack Version

v23.05.2

### Hosting Environment

ubuntu-22.04

devdot:
I've implemented a test to verify the issue: https://github.com/BookStackApp/BookStack/commit/ecf99fa0edbddc5c47eb94b9cb3b5f22c65890ee

Looking into the issue, there are two relevant places:

[Exceptions\NotFoundException](https://github.com/BookStackApp/BookStack/blob/development/app/Exceptions/NotFoundException.php)

- this exception is triggered correctly
- code 404 is set correctly
- the method `getStatus` does not exist (would be called by the exception handler, see below)
- possible solution 1: implement `getStatus`

[Exceptions\Handler::renderApiException](https://github.com/BookStackApp/BookStack/blob/development/app/Exceptions/Handler.php#L80)

- the default `$code` is 500 and Exception->getCode() only called on specific exceptions (not Exceptions\NotFoundException)
- Exception->getStatus is called when it exists, overwriting `$code`
- `$code` used to be set by Exception->getCode, but was changed: https://github.com/BookStackApp/BookStack/commit/b4fa82e3298a15443ca40bff205b7a16a1031d92
- possible solution 2: set `$code` from Exception->getCode by default

Solution 2 seems to be more generic and elegant, therefore I'll submit a PR

ssddanbrown:
Thank you for much for the PR @devdot, it looks great.

The only thing I'd be nervous about is essentially always using the exception code as a status code by default.
We can't be sure what exceptions might go through this, and I don't think we can always assure they'd be using http aligned codes.

Could I propose a tweak on your solution 1 above? 
[...]

PR #4291, opened 2023-06-08
Fix API returns HTTP 500 instead of 404 on not found

devdot:
Fix issue #4290 as described here: https://github.com/BookStackApp/BookStack/issues/4290#issuecomment-1582151669


ssddanbrown:
Thanks @devdot, now merged for next release.
I did follow this up with e72cf61f7eb2245b2af59c0862e39ad55bb7a459, to add some types and to simplify classes by removing anything that I didn't believe to be specifically used. If I've misunderstood and removed anything actively functional just let me know.

devdot:
Awesome, thanks! No, I've added the header properties and their setters only for potential future use, currently they are not functional.

PR #4309, opened 2023-06-15
Refactor Exception handling

devdot:
While working on exceptions for #4290, I found exception handling to be quite fuzzy. Trying to untangle it all, I started working on this refactoring, but I'm not sure if this is really a helpful direction at all.

First of all, there are three major kinds of exceptions:

- [PrettyException](https://github.com/BookStackApp/BookStack/blob/development/app/Exceptions/PrettyException.php), used for showing a good-looking exception to the user for those exceptions that cannot be reconciled from.
- [NotifyException](https://github.com/BookStackApp/BookStack/blob/development/app/Exceptions/NotifyException.php), used for showing a notification (session flash) after a HTTP redirect, it is also JSON compatible (without a redirect in that case).
- `\Exception`, the base exception mostly used for internal exceptions that ought to be caught in all cases (usually in controllers), therefore never exposed to user or API.

Now, I noticed the most common handling is with notifications of some kind:

- by throwing exceptions that are never caught in controllers, like `ConfirmationEmailException`: https://github.com/devdot/BookStack/blob/development/app/Access/EmailConfirmationService.php#L23
- by throwing internal exceptions and catching them in the controller, and then using `Controller:showErrorNotification`, like `MoveOperationException` https://github.com/devdot/BookStack/blob/development/app/Entities/Controllers/ChapterController.php#L188-L192
- re-throwing exceptions because they should not show the technical message to the user, like here: https://github.com/devdot/BookStack/blob/development/app/Uploads/UserAvatars.php#L116-L118
- re-throwing exceptions because they may need different messages for different controllers, like here: https://github.com/devdot/BookStack/blob/development/app/Access/Controllers/UserInviteController.php#L87-L100

So with all this observed, here is my question @ssddanbrown: Where are error notifications supposed to be handled? As I see it, there are these options:

1. Let exceptions bubble up from services, models, guards etc. into the exception handler. This would remove the exception handling from controllers entirely, making them cleaner and handling of these same types of exceptions would be the same (the `NotifyException` already creates proper HTTP responses/redirects, even for JSON). However, this would mean that services, models etc. determine how an exception should be handled, where it should be redirected to and what the message should be. That is a problem i.e. for the `UserTokenService` which currently has different error messages depending on the controller that accesses it.
2. Remove `NotifyException` altogether, make `PrettyException` the base class for all exceptions (perhaps rename to `BaseException` and let any notifying be handled by the controller, not the exception (and error handler). While this would potentially bloat controllers, the controllers get to decide how to present any error to the user. When the controllers do not catch exceptions, they would still be presentable (as extensions of `PrettyException`). However, this would eliminate the ease of throwing an error that will force a redirect with notification, which `NotifyException` currently does.
3. Leave it as it is.

To me, either 2 or 3 make to most sense.


ssddanbrown:
Hi @devdot,

> Where are error notifications supposed to be handled?

Wherever is appropriate to be honest.  If the error is specifically being handled with knowledge at the controller level, I've tended to prefer using the explicit, non-expection, methods to return a response to follow the natural job of a controller. 
But where the error is not specifically known/handled (or needed to be known) by the controller, I've left this up to the exception handler.

Out of the options presented, I'd prefer option 3. There for sure is some messiness that could be cleaned up, but while those other options advertise a cleaner or more standardised approach, there will likely still be edge cases that will impact their aims, and the current approach has worked to this point and on balance I don't think more significant changes would be worthwhile.

If you're continuing to aim for a clean-up/refactoring PR upon option 3, then I'd ask to please keep the scope of such a PR narrow. As is already, it'd take a fair amount of time to go through and properly assess the potential impact of the changes  proposed. Chasing "clean"/"optimized" code, while often necessary and important, can also be a significant distraction and vacuum of time in it's never-ending quest.

devdot:
> Chasing "clean"/"optimized" code, while often necessary and important, can also be a significant distraction and vacuum of time in it's never-ending quest.

Seeing that this quest turned out to have a much broader scope than I thought, I agree. So thanks for your reply!

I'll move some of the clean ups into separate PRs, especially those that would improve logging or API handling.

Issue #4298, opened 2023-06-11
API: Add ability to sort / order content within a book


riton:
### API Endpoint or Feature

I was unable to find an API call / endpoint that lets me adjust the **order** of content within a book.

I was only able to find an _HTML_ view that let me do that and seem to call [BookSortController](https://github.com/BookStackApp/BookStack/blob/af0b4fa851d8acda52d6176d6e9b2aef2448b499/app/Entities/Controllers/BookSortController.php#L52) for me.

### Use-Case

At [IN2P3-CC](https://cc.in2p3.fr/en/), we're planning to manage part of our documentation from Gitlab. CI/CD jobs would then update `Bookstack` according to the _Gitlab repository content_ which acts as our _source of truth_.

If new _pages_ or _chapters_ are inserted right in the middle of _other pages or chapters_ within the Gitlab repository, we require programmatic possibility to modify the _book chapters and pages_ order to match the one represented in our Gitlab repository.

### Additional context

Thanks in advance for your time and consideration

ssddanbrown:
Thanks to @rouet in #4313 this functionality is now merged to be part of the next feature release.
A `priority` integer property can be provided on chapters and pages to control the ordering (smallest to largest priority).

PR #4313, opened 2023-06-15
[API] add priority

rouet:
[API] add priority in book read
[API] add priority in chapter create and update
[API] add priority in page create and update

I propose a pull request in order to add the possibility of modifying the priority field of pages and chapters by the api, this in order to allow sorting a book using the api.
This proposal solves the problem posed by @riton in ticket #4298.
I also added in addition to display the priorities when a book is requested by the rest api.
We have tested and validated it internally.
I am not a Laravel specialist, but if it is necessary to add the corresponding tests, I can do it, but it will need help on the scenarios to be tested.

Regards.


ssddanbrown:
Thanks for this @rouet.

> We have tested and validated it internally.

Is it currently working on page create? Not tested it myself but looking at the existing code I'd be surprised if priority was filled here.

> I am not a Laravel specialist, but if it is necessary to add the corresponding tests, I can do it, but it will need help on the scenarios to be tested.

Tests would be ideal. For the update endpoints, you can update the existing `test_update_endpoint` tests to include priority. [Example of such test](https://github.com/BookStackApp/BookStack/blob/0323ebccd3747cc757819e265ccad87889ec0acf/tests/Api/ChaptersApiTest.php#L127). 
For the create endpoints, it'd be best to have new tests for those, so a `test_create_applies_correct_priority` test cases for both chapters and pages. You can follow the `test_create_endpoint` test cases as an example in the relevant `PagesApiTest` and `ChaptersApiTest` classes.

Additionally, could you update the relevant example API requests/responses within this directory?:
https://github.com/BookStackApp/BookStack/tree/1e220c473fa71a412918a8e55b63949019b40744/dev/api

rouet:
> Thanks for this @rouet.
> 
> > We have tested and validated it internally.
> 
> Is it currently working on page create? Not tested it myself but looking at the existing code I'd be surprised if priority was filled here.
> 

my tests were not complete, indeed, it does not work at creation, should we make it work? because I have the impression that the modification of the code will be much more consequent.

> > I am not a Laravel specialist, but if it is necessary to add the corresponding tests, I can do it, but it will need help on the scenarios to be tested.
> 
> Tests would be ideal. For the update endpoints, you can update the existing `test_update_endpoint` tests to include priority. [Example of such test](https://github.com/BookStackApp/BookStack/blob/0323ebccd3747cc757819e265ccad87889ec0acf/tests/Api/ChaptersApiTest.php#L127). For the create endpoints, it'd be best to have new tests for those, so a `test_create_applies_correct_priority` test cases for both chapters and pages. You can follow the `test_create_endpoint` test cases as an example in the relevant `PagesApiTest` and `ChaptersApiTest` classes.
> 
> Additionally, could you update the relevant example API requests/responses within this directory?: https://github.com/BookStackApp/BookStack/tree/1e220c473fa71a412918a8e55b63949019b40744/dev/api

ok i'll take care of that

ssddanbrown:
> it does not work at creation, should we make it work?

Ideally yes, but if you're uncomfortable making further changes you can ignore this but be sure to still add a test, so that I have a failing test case to address before merging.

rouet:
Hi

I've push a last version (if it's ok) with priority set correctly on insert.
I wrote or modify some tests
I modify the Json examples

regards

ssddanbrown:
Thanks for this @rouet,
Now merged in to development to be part of the next feature release.
Made a few last minor changes in 9ca1139ab0d0f5dd8d8eda4e9cd96f6c33530344.

PR #4314, opened 2023-06-15
Add resource id

rouet:
I propose a pull request in order to display the internal id on the "book" page, the "chapter" page and the "page" page.
For what ? When we use the rest api intensively, it is difficult to know the id of an element by browsing the web interface, this id that we use to request objects by the rest api.
This id will be displayed in the page of each entity in the metadata block.
A configuration option can be added to enable this display by the administrator.
I am not a Laravel specialist, but if it is necessary to add the corresponding tests, I can do it, but it will need help on the scenarios to be tested.

Regards


rouet:
Capture d’écran 2023-06-15 à 16 46 28

ssddanbrown:
Thanks for offering this @rouet, but I've never been keen on exposing metadata without there being specific value to the main audience of the platform, especially where the are alternative options like there are here since:

- You can get the ID via the [content/permalink popup](https://www.bookstackapp.com/docs/user/reusing-page-content/#easily-grabbing-the-syntax). 
- You could expose this detail via the [visual theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/visual-theme-system.md).
- You could expose this detail via custom custom code in the "HTML Head Content" customization setting.

rouet:
> Thanks for offering this @rouet, but I've never been keen on exposing metadata without there being specific value to the main audience of the platform, especially where the are alternative options like there are here since:
> 
> * You can get the ID via the [content/permalink popup](https://www.bookstackapp.com/docs/user/reusing-page-content/#easily-grabbing-the-syntax).

it's really not an easy fix for our use of BookStack. in addition we can not have the id of a chapter or a book

> * You could expose this detail via the [visual theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/visual-theme-system.md).

we will use this system for our needs

> * You could expose this detail via custom custom code in the "HTML Head Content" customization setting.

the pull request may be closed

PR #4317, opened 2023-06-16
Modify HttpFetchException flow to log the exception

devdot:
Currently, `HttpFetchException` is caught and logged, but without the exception (that would contain the actual curl error):

https://github.com/BookStackApp/BookStack/blob/ec775aec02c0887d5cf2dc23c938a75b7eaf67d2/app/Uploads/UserAvatars.php#L46-L53

In that helper method, the exception is caught and re-thrown, but without the original exception as previous exception:

https://github.com/BookStackApp/BookStack/blob/ec775aec02c0887d5cf2dc23c938a75b7eaf67d2/app/Uploads/UserAvatars.php#L112-L121

The proposed change is to add the exception to the logged message for improved debugging, and to make the exception a `PrettyException` for when it's shown to the user. By using `PrettyException`, the details property is available to show the original (technical) error message, although that may be removed if it should rather not be shown to the user.

### DIFF
--- a/app/Exceptions/HttpFetchException.php
+++ b/app/Exceptions/HttpFetchException.php
@@ -2,22 +2,8 @@
 
 namespace BookStack\Exceptions;
 
-class HttpFetchException extends PrettyException
+use Exception;
+
+class HttpFetchException extends Exception
 {
-    /**
-     * Construct exception within BookStack\Uploads\HttpFetcher.
-     */
-    public function __construct(string $message = '', int $code = 0, ?\Throwable $previous = null)
-    {
-        parent::__construct($message, $code, $previous);
-
-        if ($previous) {
-            $this->setDetails($previous->getMessage());
-        }
-    }
-
-    public function getStatusCode(): int
-    {
-        return 500;
-    }
 }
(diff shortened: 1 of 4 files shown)


ssddanbrown:
Thanks for this @devdot!

I agree the actual underlying error should be logged, to provide something for users to go on, and I like the test added to ensure that happens, but I'm not too sure on the exception changes.

Previously `HttpFetchException` was a plain exception, indicating that it's for internal, non-user-facing, use. 
With these changes, it's running double-duty, by being a `PrettyException` while also directly carrying internal error messages, with juggling/handling between the two. 
I'm not sure this change, along with the complexity it adds, is worthwhile here.

I appreciate you may be attempting to align my messy variation of exception handling (Plain inner exceptions vs pretty/notify exception usage) but I don't want to add complexity to get there.

Additionally, the type changes within the `UserAvatars.php` `catch ()` block does change what may be caught there. A lot more can go wrong in the image save process, beyond the HTTP fetch. One of those altered blocks doesn't even make a http request (in standard file driver usage).

devdot:
Alright, thanks for the feedback. I reduced the changes to logging and the test

ssddanbrown:
Wonderful, thank you very much for this @devdot!

PR #4318, opened 2023-06-16
Change JsonDebugException to Responsable interface

devdot:
In all other exceptions, when a Response is supposed to be returned, the Responsable interface is used instead of render.

It's a simple small issue that only regards code quality.


ssddanbrown:
Thanks @devdot!

PR #4320, opened 2023-06-16
Improve ApiAuthException control flow

devdot:
First of all, the `UnauthorizedException` seems to be a strange base class for `ApiAuthException`, as it implies HTTP 401 but the exceptions are sometimes thrown with 403 instead:

https://github.com/BookStackApp/BookStack/blob/ec775aec02c0887d5cf2dc23c938a75b7eaf67d2/app/Api/ApiTokenGuard.php#L134-L152

Further, the `ApiAuthenticate` middleware catches all exceptions that are thrown by the `ApiTokenGuard`, only to return a JSON response that is exactly the same as the application exception handler:

https://github.com/BookStackApp/BookStack/blob/ec775aec02c0887d5cf2dc23c938a75b7eaf67d2/app/Http/Middleware/ApiAuthenticate.php#L15-L25

[unauthorizedResponse](https://github.com/BookStackApp/BookStack/blob/ec775aec02c0887d5cf2dc23c938a75b7eaf67d2/app/Http/Middleware/ApiAuthenticate.php#L65-L73) will return the same JSON response as [the exception handler](https://github.com/BookStackApp/BookStack/blob/ec775aec02c0887d5cf2dc23c938a75b7eaf67d2/app/Exceptions/Handler.php#L80-L109).

My changes would further simplify the exception return messages for API errors into one place, rather than multiple places that create the standardized JSON response pattern.

The only downside I see is that this change would make the exceptions get logged through the exception logger, which may not be desired (altough that might be an improvement as well, right now there is no logging of failed API access attempts). However, if this exception should not be logged, it could be added here:

https://github.com/BookStackApp/BookStack/blob/ec775aec02c0887d5cf2dc23c938a75b7eaf67d2/app/Exceptions/Handler.php#L22-L25


ssddanbrown:
Thanks for this @devdot. I like a lot of the cleanup here.

One thing that I'm not keen on are the changes to the `ApiTokenGuard`.  They significantly change the signature of the `user()` method. 
As far as I can see, these changes are fine in context of the current app state, looking specific at how this class is used.
But it departs from the current expected `?User` signature of the method, becoming `User || throws` (Which is what `authenticate()` is specifically there for) which also misaligned this with BookStack's and Laravel's guards.
While this may have no logical negative impact right now, I feel this brings potential to cause trouble in the future where assumptions may be made about Guards being aligned.

devdot:
I agree, changing the signature of `user()` may cause trouble in some future cases. As I understand it, changing a return value from nullable to never-null should not be an issue, since null is never the anticipated value in a well designed system. But throwing an exception might be unexpected, so it may be best to keep it as-is.

The changes in `ApiTokenGuard` are unnecessary when `user()` cannot throw an exception. They are all in https://github.com/BookStackApp/BookStack/pull/4320/commits/706efb9497f2f8185e908b7a7c2ddddaa21e07f2 . If the changes in https://github.com/BookStackApp/BookStack/pull/4320/commits/74097bd47c050cd3ddd1c3ccf7eef3a20bc463a3 are alright, I'll reduce the PR to only that commit

ssddanbrown:
> If the changes in https://github.com/BookStackApp/BookStack/commit/74097bd47c050cd3ddd1c3ccf7eef3a20bc463a3 are alright, I'll reduce the PR to only that commit

Yeah, they're all good as far as I'm concerned.

ssddanbrown:
Thanks again @devdot!

PR #4329, opened 2023-06-20
added toggleable feature to force books to exist in a single shelf

scwheele:
My goal was to implement this in a way that was not disruptive to the m:m relationship between books and shelves. This change caters to larger organizations that want to be more restrictive in how users (or teams) can interact with books and shelves within their own domain. 

- Books will only exist on a single shelf
- Newly created books will automatically inherit the permissions of the shelf they were created on

If the feature is enabled:
- Create Book link is hidden from the books/index view (I wanted this to allow the user to select a shelf, but thought that might create more disruption than value)
- Permissions link is hidden from the books/show view
- Books can only be created from the "Shelf" view
- Users still require 'book-create-all' to be able to create books

### DIFF
--- a/app/Entities/Controllers/BookController.php
+++ b/app/Entities/Controllers/BookController.php
@@ -78,6 +78,12 @@ class BookController extends Controller
             $this->checkOwnablePermission('bookshelf-update', $bookshelf);
         }
 
+        if(!$bookshelf && setting()->get('app-force-books-to-shelf'))
+        {
+            $this->showErrorNotification(trans('errors.force_book_to_shelf'));
+            return redirect('/');
+        }
+
         $this->setPageTitle(trans('entities.books_create'));
 
         return view('books.create', [
@@ -112,6 +118,15 @@ class BookController extends Controller
         if ($bookshelf) {
             $bookshelf->appendBook($book);
             Activity::add(ActivityType::BOOKSHELF_UPDATE, $bookshelf);
+        }
+
+        if(setting()->get('app-force-books-to-shelf'))
+        {
+            $shelfPermissions = $bookshelf->permissions()->get(['role_id', 'view', 'create', 'update', 'delete'])->toArray();
+
+            $book->permissions()->delete();
+            $book->permissions()->createMany($shelfPermissions);
+            $book->rebuildPermissions();
         }
 
         return redirect($book->getUrl());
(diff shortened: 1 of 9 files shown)


ssddanbrown:
Thanks for offering this @scwheele but, while I understand the desire for such a feature, I'm not sure it'd be something I'd want to include. This would introduce variance, or branching logic, for existing feature-set via settings, which is something I've always tried to avoid where possible, especially where it relates to the content structure and user flow. While things don't look overly complex to maintain when focused on this one request, going down the path of supporting logical variations in the system ends up having multiplied complexity impacts.

Even if I ignored my above concerns, the scope and complexity of this request soon increases when you start think about the edge cases or all user paths, and/or additional requirements for inclusion. For example:

- Books can be added to other shelves via shelf edit view.
- Shelves can be deleted and leave books stranded.
- The API would also need to handle things differently, or be misaligned with the UI behaviour.
- Testing would be required to cover this alternative behaviour.
- Users could switch this but already not be following the expected state of the system.

This tries to move closer to wanting shelves => books to act the same as books => chapters, but the system isn't fundamentally built to support that so there's always going to be awkwardness and misalignment, affecting user expectations, without fundamental change to how shelves are built as part of the hierarchy. 
I don't want to try to move closer to that, adding complexity to support two different ux/structure flows, especially if the end result will remain awkward.

scwheele:
> This tries to move closer to wanting shelves => books to act the same as books => chapters, but the system isn't fundamentally built to support that so there's always going to be awkwardness and misalignment, affecting user expectations, without fundamental change to how shelves are built as part of the hierarchy.
I don't want to try to move closer to that, adding complexity to support two different ux/structure flows, especially if the end result will remain awkward.

Thanks for the feedback! I agree with the functionality awkwardness this could create, however, I disagree with how you are interpreting it. I don't want shelves to act as books but rather be able to have granular control over how users can interact with shelves. This proposal was more guided towards treating shelves as containers and wrapping access controls around them. Additionally preventing books from being created without being attached to that container (that the user would need access to).

A good comparison would be how RBAC roles work with Resource Groups in Azure. We can provide specific permissions to different sets of users based on how we want to allow them to interact. Some may only view, some may have change access, and some may have no access at all. 

I'll close this PR for now and reevaluate the approach and see where it leads. Thanks again for taking a look!

PR #4344, opened 2023-06-26
Update API Docs

devdot:
I came across a few minor issues in the API documentation. 

(1) In the example response for read shelf/book, the tags have `id` - but they do not have `id` just like at any other endpoint.

(2) In the example response for read book, the page within `contents` does not have a `type` - but a quick test on the demo shows that any object within `contents` has a `type`.

I also noticed some fields are missing in the example responses, e.g. pages have an `editor` field (and sometimes `url`) and users within entities also a `slug` besides `name` and `id`. I assume it's an unnecessary bloat of the example responses to add those?


ssddanbrown:
Thanks @devdot!

> I also noticed some fields are missing in the example responses, e.g. pages have an editor field (and sometimes url) and users within entities also a slug besides name and id. I assume it's an unnecessary bloat of the example responses to add those?

If it's just singular odd fields missing that are currently being provided in API responses, then it's probably not due to bloat, but instead there have been more recent changes, after responses were recorded, what have tangentially affected those old endpoints. Ideally those responses would correctly reflect all fields.

devdot:
I've gone through the example responses and made them match the actual responses as I worked through them for a project I'm working on.

One potential issue I found is that chapters create does not return `book_slug`, even though all other chapters endpoints do return that value. Adding that should be trivial, but it's not in the scope of this PR.

ssddanbrown:
Thank you very much for these updates @devdot!

PR #4356, opened 2023-07-01
Update database.php to fix "1071 Specified key was too long" error

Konstantin-Davidneko:
Simple fix for "1071 Specified key was too long" error in some environments.

### DIFF
--- a/app/Config/database.php
+++ b/app/Config/database.php
@@ -75,7 +75,7 @@ return [
             'prefix'         => env('DB_TABLE_PREFIX', ''),
             'prefix_indexes' => true,
             'strict'         => false,
-            'engine'         => null,
+            'engine'         => 'InnoDB ROW_FORMAT=DYNAMIC',
             'options'        => extension_loaded('pdo_mysql') ? array_filter([
                 PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this @Shultc but is this an active & current problem in any production scenarios?
As far as I'm aware, I've only seen this problem pop up recently for a few specific WAMP based scenarios where very old, and now uncommon, defaults were used. I'd be hesitant to make a database connection change just for those scenarios, since it's these kind of things that can have unknown and unintended consequences.

Konstantin-Davidneko:
I think it is better to test if it have any bad consequences. I have this problem, and I need to add this change after every update, which is cumbersome... And while searching for solution I found users with BookStack who faced same issue. But I fully understand your concern.

What do you think, maybe it will be good solution to move this configuration to .env file, so affected users will be able to change it only once?

If I will make such pull-request, will you approve it? Or maybe you want to make it by yourself?

ssddanbrown:
I'm going to go ahead and close this off.
The consequences are not something we can test for since they're unknown, which is my fear.
If this was a production environment issue I'd consider supporting an added option but, as far as I'm aware, this is really only caused by a strange outdated default in specific software that's rarely used for production environments. Additionally there is already a workaround via changing the database engine to match that of modern MySQL/MariaDB.

PR #4393, opened 2023-07-23
 enable search bar for non-english characters search in normal search mode.

charlietag:
Enable search bar for non-english characters search in normal search mode.
And make it configurable in .env

```
ENHANCE_SEARCH_BAR_COMPATIBILITY=false
```


ssddanbrown:
Thanks for offering this @charlietag but I wouldn't want to take this approach so I'm going to close off this PR.
I don't like to introduce options for branching logic where possible, and this somewhat makes the search index functionality redundant.

At a high level, it's a workaround, whereas the core issue attempting to be addressed should probably be dealt with at an indexing level.

Issue #2743, opened 2021-05-11
Intervention error: Unable to init from given binary data. 


TheMightyOnyx:
https://flareapp.io/share/67O6nykP

**Describe the bug**
Error while changing book cover image.

**Steps To Reproduce**
Change book cover image.

**Configuration:**
 - Exact BookStack Version: v21.04.4
 - PHP Version: PHP 7.4.19 (php:7.4.19-fpm-alpine3.13)
 - Hosting Method: Docker

**Additional context**
Uploading images to pages doesn't work either, the image itself is uploaded and copied to the storage directory correctly, but no thumbs or scaled directories are created. I checked the permissions and everything and it seems fine, I even chmoded the dir to 777 and it still doesn't work. Considering the image is copied but no thumbs are created I imagine the error happens at some point during the thumbnail creation. I imagine there's some php module missing or something misconfigured, but I followed the list of requirements on docs/installation page and it seems all fine. I checked if all modules are enabled and it seems to be the case.

I enabled debugging and linked the stack trace above.

ssddanbrown:
Hi @TheMightyOnyx,
Are you using a common docker image at all? Or are you able to share the scripts/Dockerfile to build your docker image at all? This kind of error can some down to the exact version/build of the GD php extension.

TheMightyOnyx:
This is the php part that installs the modules and composer etc. It's based on `FROM php:7.4.19-fpm-alpine3.13`. If this is not enough and you need to test it yourself I can grant you access to the repository, it's on Gitlab though, so I would need your Gitlab username.
```
....
ARG PROJECT_VERSION=21.04.4
ARG PROJECT=https://github.com/bookstackapp/bookstack/archive/v${PROJECT_VERSION}.tar.gz
....
RUN \
# install dependencies
apk add --no-cache gettext libpng-dev oniguruma-dev libxml2-dev wget && \
# create project directory and download the project files, su into user otherwise umask is ignored
mkdir -p /usr/share/project && \
chown ariamis:ariamis /usr/share/project && \
wget -qO- "${PROJECT}" | su ariamis -s /bin/sh -c "tar -xzf - -C /usr/share/project --strip-components=1" && \
# install php extensions
docker-php-ext-install pdo_mysql mbstring tokenizer gd simplexml dom && \
# remove uneccesary data
apk del libpng-dev oniguruma-dev libxml2-dev wget && \
apk add --no-cache libpng oniguruma libxml2 && \
# copy php configuration
[...]

PR #4422, opened 2023-08-03
Fixed Missing Dependences in Dockerfile

Scemist:
The currently Dockerfile configuration doesn't upload JPG or JPEG, this changes solves this problem, and better format the file.

#2743

### DIFF
--- a/dev/docker/Dockerfile
+++ b/dev/docker/Dockerfile
@@ -3,14 +3,31 @@ FROM php:8.1-apache
 ENV APACHE_DOCUMENT_ROOT /app/public
 WORKDIR /app
 
-# Install additional dependacnies and configure apache
-RUN apt-get update -y \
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this @Scemist, but this PR seems to have a lot more changes than I'd expect for just adding better image support. For example, there's added items for libxml, mcrypt and libtidy. Additionally workarounds for ldap use on M1 macs/arm64 have been removed.

In regards to formatting, I would prefer that tangential formatting would not be done on functional PR changes since it makes it much harder to understand the changes made.

ssddanbrown:
I'm going to go ahead and close this off due to the above concerns. Would be happy though to receive a PR that just makes the necessary changes to fix the issue.

Either way, at some point I'd like to update the dev docker setup to be Ubuntu based & simpler, with all requirements in a single container, which will help avoid these specific kind of extension/compatibility issues. Not sure when I'll get to that though.

PR #4478, opened 2023-08-30
Create tag-export.blade.php

Man-in-Black:
An entity which allow to include tags in a PDF export


ssddanbrown:
Closing [as per my comment in #4479](https://github.com/BookStackApp/BookStack/pull/4479#issuecomment-1699552261)

PR #4479, opened 2023-08-30
Update page.blade.php

Man-in-Black:
Included the page content and the tags from the entity tag-export

### DIFF
--- a/resources/views/exports/page.blade.php
+++ b/resources/views/exports/page.blade.php
@@ -3,7 +3,40 @@
 @section('title', $page->name)
 
 @section('content')
-    @include('pages.parts.page-display')
+    <div dir="auto">
+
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering but I take a very conservative approach to showing extra information/detail in areas.
I would only really do so upon significant proven requirement/purpose, which I don't feel is proven for these changes.

I'd usually instead advise use of the visual theme system if you do need more that what's provided by default in these templates.

For this reasoning, I'm going to close this PR along with #4478.
If you do think changes should be in the core project then you could raise a feature request to explain the value/purpose but it may be something I close off with similar reasoning to the above.

Man-in-Black:
Ok thats fine by me.
I already did this in the visual theme system and found it quite useful, so I thought that this could maybe find its way to the core.

PR #4533, opened 2023-09-11
Fixes #4489: Adding last resort truncation to User Model's getShortName() function

### LINKED ISSUE #4489: Comments : Author's name does not appear if too long
### Describe the Bug

Hello,

we have a user with a long name (26 letters). If he writes a comment on a page, his name does not appear on the top of the comment.
He changed his name with a shorter version (4 letters), and the name appears.

I don't know what is the exact limit, but with me (14 letters), it works.

PS : thank you very much for your work @ssddanbrown, we will plan to update bookstack to get the new notification system, it looks promising.

### Steps to Reproduce

1. Change your name in your profile with a long name.
2. Write a comment in a page.
3. Check if the name is shown on the head of the comment.
4. Retry with a shorter name.

### Expected Behaviour

The name should be shown, or at least with "..." at the end if it is too long.

_No response_

### Browser Details

Brave or Firefox

### Exact BookStack Version

23.06.2

### PHP Version

_No response_

### Hosting Environment

Debian 12

Bajszi97:
This pull request closes issue #4489 

by providing a last resort option to the getShortName() function within the User model. This option is necessary because the function fails to return anything in some cases. If the name is too long and cannot be split by spaces, it will be trimmed to the specified number of characters, followed by two dots ('..').

### DIFF
--- a/app/Users/Models/User.php
+++ b/app/Users/Models/User.php
@@ -345,7 +345,7 @@ class User extends Model implements AuthenticatableContract, CanResetPasswordCon
             return $splitName[0];
         }
 
-        return '';
+        return mb_substr($this->name, 0, $chars) . '..';
     }
 
     /**
--- a/app/Users/Models/User.php
+++ b/app/Users/Models/User.php
@@ -345,7 +345,7 @@ class User extends Model implements AuthenticatableContract, CanResetPasswordCon
             return $splitName[0];
         }
 
-        return mb_substr($this->name, 0, $chars) . '..';
+        return mb_substr($this->name, 0, $chars-3) . '…';
     }
 
     /**
--- a/tests/Entity/CommentTest.php
+++ b/tests/Entity/CommentTest.php
@@ -152,4 +152,20 @@ class CommentTest extends TestCase
         $respHtml = $this->withHtml($this->get($page->getUrl('/edit')));
         $respHtml->assertElementContains('.comment-box .content', 'My great comment to see in the editor');
     }
+
+    public function test_comment_creator_name_truncated()
+    {   
+        $longNamedUser = $this->users->admin();
+        $longNamedUser->name = 'Wolfeschlegelsteinhausenbergerdorff';
+        $longNamedUser->save();
+        $this->actingAs($longNamedUser);
+
+        $page = $this->entities->page();
+
+        $comment = Comment::factory()->make();
+        $this->postJson("/comment/$page->id", $comment->getAttributes());
+
+        $pageResp = $this->get($page->getUrl());
+        $pageResp->assertSee('Wolfeschlegel…');
+    }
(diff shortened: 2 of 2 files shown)


ssddanbrown:
Thanks for this @Bajszi97!

Could you tweak this to use three dots instead of two? To form a full ellipsis. 
Additionally, could you tweak the truncation to take into account the added ellipsis (so reduce truncate length by three)?

It would be good to add a test to cover the originally raised issue in #4489, just a check to view a page with a comment and ensure the expected truncated name is visible.
Could be added to the existing [CommentTest class](https://github.com/BookStackApp/BookStack/blob/0359e2490a347c2bc245b21535e6939254493767/tests/Entity/CommentTest.php). 
Would you want to give that a go? Otherwise I can add that myself before merge.

otherjoel:
Even better, use [the actual ellipses character](https://practicaltypography.com/ellipses.html): `…`

Bajszi97:
> Could you tweak this to use three dots instead of two? To form a full ellipsis. Additionally, could you tweak the truncation to take into account the added ellipsis (so reduce truncate length by three)

Of course, I will change that. I have discarded the idea of reducing the truncated length, as it will cause problems on numbers smaller than three, even though it's very unlikely to call the function with such a small length. I will use the actual ellipses character, if it's considered better practice.

> It would be good to add a test to cover the originally raised issue in #4489, just a check to view a page with a comment and ensure the expected truncated name is visible. Could be added to the existing [CommentTest class](https://github.com/BookStackApp/BookStack/blob/0359e2490a347c2bc245b21535e6939254493767/tests/Entity/CommentTest.php). Would you want to give that a go? Otherwise I can add that myself before merge.

Yes i would. I have never done testing before, but feels like a good way to start. If i can't make it work, i will inform you here.

Bajszi97:
I have applied the changes you mentioned and managed to come up with something for the test, but I have some concerns regarding it. 
I feel like there is a better way to get a user with long enough name for the test case, but i could not figure it out.
The other concern I have is that if the length provided to the function in the [comment.blade](https://github.com/BookStackApp/BookStack/blob/development/resources/views/comments/comment.blade.php) view changes, this test may fail. However, it doesn't necessarily mean that something is wrong.

ssddanbrown:
Thanks @Bajszi97, now merged for next feature release.
I made some tweaks during review in a452092e40110c5251c9b621593f3fa174680059. Mainly formatting and tweaking the test to use some existing helpers, and so the test was performed as another user to avoid the test potentially passing the test from the name elsewhere in the page (like the header bar).

I also adjusted the truncation to two, just because with @otherjoel's suggestion, I roughly measured the visual width of that character to be just over baseline character width, so I rounded up to consider it as 2 characters.

Issue #4394, opened 2023-07-23
Sidebar scrolling difficult and inconsistent at mid-range sceen size breakpoints


LawssssCat:
### Describe the feature you'd like

add a `max_height` attirbute to Page Navigation

<details><summary>Details</summary>
<p>

![ccc](https://github.com/BookStackApp/BookStack/assets/18041500/4e17e530-b611-4d81-b9c5-75db5134312c)

</p>
</details> 
 

### Describe the benefits this would bring to existing BookStack users

No need to scroll to the end of the page to see the content of Page Navigation beyond the screen.

### Can the goal of this request already be achieved via other means?

..

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundemental request

### How long have you been using BookStack?

6 months to 1 year

### Additional context

BookStack v23.06.2

ssddanbrown:
Thanks for reporting @LawssssCat, I'm going to reclassify this as a front-end bug since the behaviour is misaligned with full desktop view.

Note for my future reference: This is specific to the medium screen breakpoint range, where there's a single left-hand sidebar.

ssddanbrown:
Now addressed as of merge of #4562, and follow-up commit of 031067745b9e95a39639f9ff165d71331370bf0b.
Will be part of the next feature release.

PR #4562, opened 2023-09-19
fix Sidebar scrolling at mid-range sceen

LawssssCat:
Reference Issues: https://github.com/BookStackApp/BookStack/issues/4394


ssddanbrown:
Thanks @LawssssCat, Now merged.
I followed this up with some more substantial layout structure changes in 031067745b9e95a39639f9ff165d71331370bf0b to achieve a structure where the two sidebars can be treated as one, at this screen-size, for scrolling and sticky behaviour.

PR #4629, opened 2023-10-29
Modified mail notifications to add the name of the book

Man-in-Black:
As written on discord I've changed the mail notifications to add the title of the book.
It's my first real pull request, so I hope I've done everything right ;)
This time with the right codebase ;)

### DIFF
--- a/app/Activity/Notifications/Messages/PageCreationNotification.php
+++ b/app/Activity/Notifications/Messages/PageCreationNotification.php
@@ -26,7 +26,9 @@ class PageCreationNotification extends BaseActivityNotification
         ];
 
         if ($chapter) {
-            $listMessageData[$locale->trans('notifications.detail_chapter_name') => $chapter->name;
+            $listMessageData += [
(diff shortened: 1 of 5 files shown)


Man-in-Black:
I've just got the idea to extend the whole notification, so as fpr now it is showing the book name as well as the chapter name if present.

without chapter:
image

with chapter:
image

ssddanbrown:
Thanks again for this @Man-in-Black.
I've now merged this in via commit 65ac197be440b90deccfd13d4eb8e6bec11e23b0 (I squashed your changes down to a single commit to clean up the history of changes). 

I followed this up with d41fd7a8dd2a9e4f4117a07c9f87346dbbd661f6, which added testing and tweaked the format a bit so that both the parent chapter and book are instead shown as a "Page Path:" item in the details list, with the names linked to the item in BookStack. This is to keep the details focused on pages being the main context of the email.
The changes in this commit also add testing, permission control and add these details also to the comment notification.

Issue #2208, opened 2020-07-31
web.config File Information Disclosure


isgroup:
**Describe the bug**

When BookStack is installed on an Apache system the "web.config" file is not parsed and exposed. While this has no impact as "web.config" contents are known, BookStack's users will get a Medium severity issue from well-known security scanners.

A simple and effective solution is to disable access to "web.config" in the ".htaccess". This will work both on Apache and IIS.

**Steps To Reproduce**

`GET /web.config HTTP/1.1` on an Apache installation.

**Expected behavior**

Ensure proper restrictions are in place, or remove the web.config file if the file is not required.

**Screenshots**

N/A

**Your Configuration (please complete the following information):**

BookStack Docker Version (latest)

**Additional context**

An information disclosure vulnerability exists in the remote web server due to the disclosure of the web.config file. An unauthenticated, remote attacker can exploit this, via a simple GET request, to disclose potentially sensitive configuration information.

ssddanbrown:
Thanks for reporting @isgroup.

To be honest though, I don't really want to go to any additional effort just to prevent a potential warning in overly sensitive/alertive auto scanners. 

The `web.config` file supplied with BookStack is public. Although I'm not that familiar with IIS, as far as I can tell hiding the file won't really provide any additional benefit, especially when done on a webserver (Apache) that does not even support that config file format.

isgroup-srl:
Hi Dan, thanks for your response.

I have the impression that you did not understood the proposed solution solution: since Apache does not even support that config it's totally safe to forbid access to "web.config". No nasty side effect and it will save users from compliance headaches from overly sensitive/alertive auto scanners, that sadly are commonly used. A `404` could be returned.

Naturally, I agree that this is a very low priority and has an actual security impact only in edge-cases that does not correspond to the currently shipped configuration (the file is public and it doesn't contain sensitive information).

BTW, very cool software!

For future reference:

- https://www.drupal.org/project/drupal/issues/2948579
- https://stackoverflow.com/questions/42949438/restrict-download-of-htaccess-and-web-config-files

PR #4664, opened 2023-11-10
Forbid web.config when running Apache

jsnider2:
Simple rewrite rule to address https://github.com/BookStackApp/BookStack/issues/2208

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -5,6 +5,8 @@
 
     RewriteEngine On
 
+    RewriteRule ^web.config$ - [R=403,L]
+
     # Handle Authorization Header
     RewriteCond %{HTTP:Authorization} .
     RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]


ssddanbrown:
Thanks for offering this @jsnider2, but as per my comments in #2208 I don't really understand the point of this.
I don't see why we should add config, to prevent viewing of a public viewing of a config file for a different server, especially in a manner which will only apply for a subset of environments where that web-server is used (since Apache .htaccess files are less commonly used here).

I'm not keen to make changes just to appease scanners, I'd prefer there to be functional purposes to changes we make.
If you really need to apply this rule, it could be done server-side in that specific environment.

Let me know if I've missed something though or misunderstood.

jsnider2:
The concern you voiced in the previous issue was that you didn't "want to go to any additional effort" which is why I provided a PR that you could merge. We have plenty of other ways to address this, as you point out, but since we're not the first to see this as an issue I thought I'd give it a shot and see if you'd take it. It you don't feel it's in the best interests of your application then feel free to close it.

ssddanbrown:
Thanks for confirming, I'll therefore close this off.

The effort isn't specific to implementation, it's also testing the changes, including thinking through how it'd apply to different environments, while also exanding the scope to consider if we need to be consistent elsewhere in our config, while having more config that might be queried in support.

PR #4691, opened 2023-11-24
Feat more shelves per page

ianroberts-bizclik:


ianroberts-bizclik:
Apologies, pull requested to upstream not origin

Issue #3873, opened 2022-11-25
OpenID Connect: Use group details from user_info endpoint


107142:
### Describe the Bug

It seems the application only parses the `id_token` when enumerating group claims but not the `userinfo` endpoint resulting in missing groups when `user_info` is in use.
We have a large amount of custom claims containing lots of groups making usage of `id_token` impossible (as its size would be simply too much).

### Steps to Reproduce

1. Make sure you IdP uses `user_info` to send claims with groups
2. Configure OIDC to sync groups
3. Dump user detail upon login

### Expected Behaviour

A list of groups should be returned.

### Browser Details

Mozilla/5.0 (X11; Linux x86_64; rv:102.0) Gecko/20100101 Firefox/102.0

### Exact BookStack Version

v22.10.2

### PHP Version

8.1.12

### Hosting Environment

Rancher Kubernetes

Docker image: solidnerd/bookstack:latest

Clean install

107142:
Updated the description as at first I thought there is an issue with the groups claim parsing logic. But after going through the source it seems like the issue is we do not include groups in `id_token`, but in `user_info` and that does not seem to be used.

ssddanbrown:
I'll re-label this as an auth feature request, since we currently don't currently support gaining extra claims from the `user_info` endpoint as you have found, and it's not a bug in existing logic.

felixschloesser:
As this issue was a showstopper for my organisation I hacked something together which works for my needs. I know next to nothing about lavarel or php; so I dont consider this PR worthy, but maybe it can be a starting point for someone else.

I overhauled the get user details function to use the access token to query both name and groups form the userinfo endpoint as defined by the provider. As I didnt get it to work with the Psr\Http\Client\ClientInterface I just resorted to curl.

<details>
<summary>OidcService.php:</summary>

```php
    /**
     * Query the user info endpoint for the user's details.
     *
     * @return array{name: string, email: string, external_id: string, groups: string[]}
     */
    protected function getUserInfo(OidcAccessToken $accessToken): ?array 
    {
        $settings = $this->getProviderSettings();
        $user_info_endpoint = $settings->userInfoEndpoint;

        $headers = [
            'Authorization: Bearer ' . $accessToken,
            'Content-Type: application/json',
        ];

        $curl = curl_init();
        curl_setopt_array($curl, [
            CURLOPT_URL => $user_info_endpoint,
            CURLOPT_RETURNTRANSFER => true,
[...]

PR #4726, opened 2023-12-15
Bugfix: Oidc: Properly query the UserInfo Endpoint

LukeShu:
(This is against the `development` branch; if you think this is worth backporting to `release`, I have a version of it against the `release` branch ready-to-go.)

BooksStack's OIDC Client requests the 'profile' and 'email' scope values in order to have access to the 'name', 'email', and other claims.  It looks for these claims in the ID Token that is returned along with the Access Token.

However, the [OIDC-core specification section 5.4][1] only requires that the Provider include those claims in the ID Token *if* an Access Token is not also issued.  If an Access Token is issued, the Provider can leave out those claims from the ID Token, and the Client is supposed to obtain them by submitting the Access Token to the UserInfo Endpoint.

So I suppose it's just good luck that the OIDC Providers that BookStack has been tested with just so happen to also stick those claims in the ID Token even though they don't have to.  But others (in particular: https://login.infomaniak.com) don't do so, and require fetching the UserInfo Endpoint.)

A workaround is currently possible by having the user write a theme with a ThemeEvents::OIDC_ID_TOKEN_PRE_VALIDATE hook that fetches the UserInfo Endpoint.  This workaround isn't great, for a few reasons:
 1. Asking the user to implement core parts of the OIDC protocol is silly.
 2. The user either needs to re-fetch the .well-known/openid-configuration file to discover the endpoint (adding yet another round-trip to each login) or hard-code the endpoint, which is fragile.
 3. The hook doesn't receive the HTTP client configuration.

So, have BookStack's OidcService fetch the UserInfo Endpoint and inject those claims into the ID Token, if a UserInfo Endpoint is defined. Two points about this:
 - Injecting them into the ID Token's claims is the most obvious approach given the current code structure; though I'm not sure it is the best approach, perhaps it should instead fetch the user info in processAuthorizationResponse() and pass that as an argument to processAccessTokenCallback() which would then need a bit of restructuring.  But this made sense because it's also how the ThemeEvents::OIDC_ID_TOKEN_PRE_VALIDATE hook works.
 - OIDC *requires* that a UserInfo Endpoint exists, so why bother with that "if a UserInfo Endpoint is defined" bit?  Simply out of an abundance of caution that there's an existing BookStack user that is relying on it not fetching the UserInfo Endpoint in order to work with a non-compliant OIDC Provider.

[1]: https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims


ssddanbrown:
Thanks for offering this @LukeShu.

Yeah, I stuck to just the ID token claims when originally building to keep the implementation simple, and since from my testing (and most usage cases so far) it's rare for claims not to be within the token, but happy to expand support to use userinfo if there are legitimate scenarios to support.

It'll probably be the next release cycle (early next year) when I come round to properly look to review and merge this since i've already spent a while on OIDC this release cycle, and don't want to expand it upon existing plans.

Related to #3873.

ssddanbrown:
Continued in #4955

ssddanbrown:
Thanks again @LukeShu.
This has now been merged after some additional work performed in #4955, and is due to be part of the next feature release.

PR #4750, opened 2024-01-01
Added Default Templates for Chapters

Man-in-Black:
I have extended the new feature of default templates for books to chapters.
This makes it possible to specify default templates within individual chapters of a book so that they are automatically selected when a new page is created.
The functionality is exactly the same as with books.
Here is an excerpt from my bookstack (in german):
image
Behind the scenes I've made changes to the controller, the repos, the models and the trash can.
There is also a migration file for the database to alter the table for the chapters so that you can set the default_template_id.

### DIFF
--- a/app/Entities/Controllers/PageController.php
+++ b/app/Entities/Controllers/PageController.php
@@ -6,6 +6,7 @@ use BookStack\Activity\Models\View;
 use BookStack\Activity\Tools\CommentTree;
 use BookStack\Activity\Tools\UserEntityWatchOptions;
 use BookStack\Entities\Models\Book;
+use BookStack\Entities\Models\Chapter;
 use BookStack\Entities\Models\Page;
 use BookStack\Entities\Repos\PageRepo;
 use BookStack\Entities\Tools\BookContents;
@@ -259,7 +260,9 @@ class PageController extends Controller
         $page = $this->pageRepo->getBySlug($bookSlug, $pageSlug);
         $this->checkOwnablePermission('page-delete', $page);
         $this->setPageTitle(trans('entities.pages_delete_named', ['pageName' => $page->getShortName()]));
-        $usedAsTemplate = Book::query()->where('default_template_id', '=', $page->id)->count() > 0;
+        $usedAsTemplate = 
+            Book::query()->where('default_template_id', '=', $page->id)->count() > 0 ||
+            Chapter::query()->where('default_template_id', '=', $page->id)->count() > 0;
 
         return view('pages.delete', [
             'book'    => $page->book,
@@ -279,7 +282,9 @@ class PageController extends Controller
         $page = $this->pageRepo->getById($pageId);
         $this->checkOwnablePermission('page-update', $page);
         $this->setPageTitle(trans('entities.pages_delete_draft_named', ['pageName' => $page->getShortName()]));
-        $usedAsTemplate = Book::query()->where('default_template_id', '=', $page->id)->count() > 0;
+        $usedAsTemplate = 
+            Book::query()->where('default_template_id', '=', $page->id)->count() > 0 ||
+            Chapter::query()->where('default_template_id', '=', $page->id)->count() > 0;
 
         return view('pages.delete', [
             'book'    => $page->book,
--- a/app/Entities/Controllers/PageController.php
+++ b/app/Entities/Controllers/PageController.php
@@ -260,7 +260,7 @@ class PageController extends Controller
         $page = $this->pageRepo->getBySlug($bookSlug, $pageSlug);
         $this->checkOwnablePermission('page-delete', $page);
         $this->setPageTitle(trans('entities.pages_delete_named', ['pageName' => $page->getShortName()]));
-        $usedAsTemplate = 
+        $usedAsTemplate =
             Book::query()->where('default_template_id', '=', $page->id)->count() > 0 ||
             Chapter::query()->where('default_template_id', '=', $page->id)->count() > 0;
 
@@ -282,7 +282,7 @@ class PageController extends Controller
         $page = $this->pageRepo->getById($pageId);
         $this->checkOwnablePermission('page-update', $page);
         $this->setPageTitle(trans('entities.pages_delete_draft_named', ['pageName' => $page->getShortName()]));
-        $usedAsTemplate = 
+        $usedAsTemplate =
             Book::query()->where('default_template_id', '=', $page->id)->count() > 0 ||
             Chapter::query()->where('default_template_id', '=', $page->id)->count() > 0;
 
--- a/app/Entities/Repos/ChapterRepo.php
+++ b/app/Entities/Repos/ChapterRepo.php
@@ -4,6 +4,7 @@ namespace BookStack\Entities\Repos;
 
 use BookStack\Activity\ActivityType;
 use BookStack\Entities\Models\Book;
+use BookStack\Entities\Models\Page;
 use BookStack\Entities\Models\Chapter;
 use BookStack\Entities\Models\Entity;
 use BookStack\Entities\Tools\BookContents;
@@ -46,6 +47,7 @@ class ChapterRepo
         $chapter->book_id = $parentBook->id;
         $chapter->priority = (new BookContents($parentBook))->getLastPriority() + 1;
         $this->baseRepo->create($chapter, $input);
+        $this->updateChapterDefaultTemplate($chapter, intval($input['default_template_id'] ?? null));
         Activity::add(ActivityType::CHAPTER_CREATE, $chapter);
 
         return $chapter;
@@ -57,6 +59,11 @@ class ChapterRepo
     public function update(Chapter $chapter, array $input): Chapter
     {
         $this->baseRepo->update($chapter, $input);
+
+        if (array_key_exists('default_template_id', $input)) {
+            $this->updateChapterDefaultTemplate($chapter, intval($input['default_template_id']));
+        }
+
         Activity::add(ActivityType::CHAPTER_UPDATE, $chapter);
 
         return $chapter;
@@ -99,6 +106,33 @@ class ChapterRepo
         Activity::add(ActivityType::CHAPTER_MOVE, $chapter);
 
         return $parent;
+    }
+
+    /**
+     * Update the default page template used for this chapter.
+     * Checks that, if changing, the provided value is a valid template and the user
+     * has visibility of the provided page template id.
+     */
+    protected function updateChapterDefaultTemplate(Chapter $chapter, int $templateId): void
+    {
+        $changing = $templateId !== intval($chapter->default_template_id);
+        if (!$changing) {
+            return;
+        }
+
+        if ($templateId === 0) {
+            $chapter->default_template_id = null;
+            $chapter->save();
+            return;
+        }
+
+        $templateExists = Page::query()->visible()
+            ->where('template', '=', true)
+            ->where('id', '=', $templateId)
+            ->exists();
+
+        $chapter->default_template_id = $templateExists ? $templateId : null;
+        $chapter->save();
     }
 
     /**
--- a/app/Entities/Repos/PageRepo.php
+++ b/app/Entities/Repos/PageRepo.php
@@ -142,7 +142,7 @@ class PageRepo
         } else {
             $defaultTemplate = $page->book->defaultTemplate;
         }
-        
+
         if ($defaultTemplate && userCan('view', $defaultTemplate)) {
             $page->forceFill([
                 'html'  => $defaultTemplate->html,
--- a/app/Entities/Repos/PageRepo.php
+++ b/app/Entities/Repos/PageRepo.php
@@ -136,7 +136,13 @@ class PageRepo
             $page->book_id = $parent->id;
         }
 
-        $defaultTemplate = $page->book->defaultTemplate;
+        // check for chapter
+        if ($page->chapter_id) {
+            $defaultTemplate = $page->chapter->defaultTemplate;
+        } else {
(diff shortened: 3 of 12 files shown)


app/Entities/Repos/ChapterRepo.php
```diff
@@ -101,6 +108,33 @@ public function move(Chapter $chapter, string $parentIdentifier): Book
...
+        }
+
+        $templateExists = Page::query()->visible()
+            ->where('template', '=', true)
+            ->where('id', '=', $templateId)
+            ->exists();
+
+        $chapter->default_template_id = $templateExists ? $templateId : null;
+        $chapter->save();
+    }
```
ssddanbrown, line 136: Since this is all effectively a duplicate of the book code, could this logic instead be extracted to the BaseRepo, so it can be shared between both the ChapterRepo and Bookrepo?

app/Entities/Repos/PageRepo.php
```diff
@@ -136,7 +136,13 @@ public function getNewDraftPage(Entity $parent)
             $page->book_id = $parent->id;
         }
 
-        $defaultTemplate = $page->book->defaultTemplate;
+        // check for chapter
+        if ($page->chapter_id) {
+            $defaultTemplate = $page->chapter->defaultTemplate;
+        } else {
+            $defaultTemplate = $page->book->defaultTemplate;
+        }
```
ssddanbrown, line 144: This logic presents a breaking change, since currently the book default template option cascades down to apply to pages created in books with that option applied.
With this logic new pages, that are in chapters, won't use the book template if set where the parent chapter does not have its own template assigned.

I think we'd want to retain the existing cascade behaviour of book templates, with a preference of the chapter template if set.

We do have a test to catch this scenario, which is currently failing due to this.

app/Entities/Controllers/PageController.php
```diff
@@ -259,7 +260,9 @@ public function showDelete(string $bookSlug, string $pageSlug)
         $page = $this->pageRepo->getBySlug($bookSlug, $pageSlug);
         $this->checkOwnablePermission('page-delete', $page);
         $this->setPageTitle(trans('entities.pages_delete_named', ['pageName' => $page->getShortName()]));
-        $usedAsTemplate = Book::query()->where('default_template_id', '=', $page->id)->count() > 0;
+        $usedAsTemplate = 
+            Book::query()->where('default_template_id', '=', $page->id)->count() > 0 ||
+            Chapter::query()->where('default_template_id', '=', $page->id)->count() > 0;
```
ssddanbrown: This logic is good, but the page delete view has the following text when this condition is true:

> This page is in active use as a book default page template. These books will no longer have a default page template assigned after this page is deleted.

This is the `entities.pages_delete_warning_template` translation. Best just to update that line to include chapters, rather than having a separate line and adding complexity to detect if it's chapters or books that are using the template.

app/Entities/Tools/TrashCan.php
```diff
@@ -208,6 +208,12 @@ protected function destroyPage(Page $page): int
 
         $page->forceDelete();
 
+        // Remove chapter template usages
+        Chapter::query()->where('default_template_id', '=', $page->id)
+            ->update(['default_template_id' => null]);
+
+        $page->forceDelete();
```
ssddanbrown: There's now a duplicate call to `$page->forceDelete();`.
You can just remove the first one, so that the force delete is the last action to run.

resources/views/chapters/parts/form.blade.php
```diff
@@ -22,6 +22,29 @@
...
+
+            <div class="min-width-m">
+                @include('form.page-picker', [
+                    'name' => 'default_template_id',
+                    'placeholder' => trans('entities.chapter_default_template_select'),
+                    'value' => $chapter->default_template_id ?? null,
+                    'selectorEndpoint' => '/search/entity-selector-templates',
+                ])
+            </div>
+        </div>
```
ssddanbrown: Since this is all pretty much the same as that used for the book form, but just with different `book/chapter` references, could this be extracted out to its own template partial (`entities.template-selector`).

ssddanbrown (changes requested):
Thanks @Man-in-Black for this PR, it's looking good and generally all seems to work apart from one minor condition.
I've added a set of notes on the code to work through.

In addition to these, there are a couple of other areas that would need consideration:

- This functionality would need testing coverage. The tests for book templates [can be found here](https://github.com/BookStackApp/BookStack/blob/v23.12.1/tests/Entity/BookDefaultTemplateTest.php). The tests would be similar, with crossover in testing, so that existing test file could just be changed to `DefaultTemplateTest`, with `book` added to the existing test case names if not already there.
- The REST API needs to be considered, to allow templates to be set/updated for chapters via the API. The API examples, and tests, would then need to be updated to suit.

If you're unsure of anything written, need help, or would instead prefer me to take on the requested work, just let me know.

Man-in-Black:
Thank you Dan for your comments, if it's okay for you I would to try to fix the parts you mentioned.
But if I got stucked somewhere I would be glad if you can give me some advise, as these are my first "bigger" steps into Laravel.

ssddanbrown:
Sure, sounds good! Happy to help and advise where needed.
If you'd prefer help via Discord chat instead of here feel free to create a thread in the help-and-support channel.

app/Entities/Controllers/PageController.php
Man-in-Black: So if I'm getting you right you suggest to update the translation in this case, so that it includes chapters as well? Something like this: `'pages_delete_warning_template' => 'This page is in active use as a book or chapter default page template. These books or chapters will no longer have a default page template assigned after this page is deleted.',` ?

app/Entities/Repos/ChapterRepo.php
Man-in-Black: That should indeed be possible, but I must admit that at this stage I'm not deep enough into the code to make this.

app/Entities/Tools/TrashCan.php
Man-in-Black: you're absolutley right, I've overseen the duplicate call and removed it now. Thanks.

resources/views/chapters/parts/form.blade.php
Man-in-Black: So you suggest to build a new partial file to reference to?

Man-in-Black:
Hi @ssddanbrown 

I'm very happy that you accepted my first "bigger" PR :-)
And I'm also happy that it generally looks good.

I'm trying to get into your comments and suggestions, but to be honest I'm struggling a bit at this point.

- I think I can provide the test cases, but unfortunately I don't know how to run such a test. Maybe you can give me some guidance on how to execute a test case.
- For the part with the API I just found 2 lines to be added in `app/Entities/Controllers/ChapterApiController.php`, but this seems to easy for me.
And also here, how could I start a test case?

The most parts I've done comes from my former PHP knowledge, so it was doable for me. But some years passed since then and after I discovered your project I've started to get back into it, but I think it needs some more time.
I would love to learn something with this PR, but don't be mad at me if I can't resolve all your comments.

I hope this is okay for you and if I'm not able to resolve everything I would be happy if you can take on the work, I would then start to build up my knowledge on your changes.

app/Entities/Repos/ChapterRepo.php
ssddanbrown: Any particular part you're having trouble with? Or would you prefer me to take on this element?

app/Entities/Controllers/PageController.php
ssddanbrown: Yeah, just like that

resources/views/chapters/parts/form.blade.php
ssddanbrown: Yeah, that's right, similar to the tags section that's just above this in the form.

ssddanbrown:
> Maybe you can give me some guidance on how to execute a test case.

Sure, I have specific guidance covering exactly this [on this page here](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md). Let me know if there's something missing from that, or if you have trouble with any of the details there.

> For the part with the API I just found 2 lines to be added in app/Entities/Controllers/ChapterApiController.php, but this seems to easy for me.

That might be all that's needed to be honest, most of the logic is handled in the repos so we can share it across the API and normal app code, we just have to make sure those properties are accepted through validation.

In regard to testing the API changes, you can see my changes to `BooksApiTest` [in the changes for this PR](https://github.com/BookStackApp/BookStack/pull/4721/files#diff-f331abb9608442a27742799ad7eab3c4a1ea72581129250e408ee3a9ab95fa97). I didn't add any specific API tests for this, but just updated existing tests to also include a default template.

> I would love to learn something with this PR, but don't be mad at me if I can't resolve all your comments.
> I hope this is okay for you and if I'm not able to resolve everything I would be happy if you can take on the work, I would then start to build up my knowledge on your changes.

I'm not mad at all, happy to take the time where possible to get others involved and to help folks learn.
Just feel free to let me know if you get to a blocking point, need help, or want me to take on parts.

app/Entities/Controllers/PageController.php
Man-in-Black: ok done ;)

app/Entities/Repos/ChapterRepo.php
Man-in-Black: I would be happy if you can take on this element.

app/Entities/Repos/PageRepo.php
Man-in-Black: If you could also take on this part I would also be happy.

resources/views/chapters/parts/form.blade.php
Man-in-Black: I've updated it as far as I could do it, I hope this fits as I chaned some translations to match it generally also with the book templates.
I hope I've done it the right way

Man-in-Black:
> Sure, I have specific guidance covering exactly this [on this page here](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md). Let me know if there's something missing from that, or if you have trouble with any of the details there.

Thanks for the guide. I followed it but I can't get it to run, I always get the following error:
~~~
root@bookstack:/var/www/bookstack/tests# composer test tests/HomepageTest.php
No composer.json in current directory, changing working directory to /var/www/bookstack
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Continue as root/super user [yes]?
> phpunit 'tests/HomepageTest.php'
PHP Fatal error:  Uncaught Error: Class "Tests\TestCase" not found in /var/www/bookstack/tests/HomepageTest.php:8
Stack trace:
#0 /usr/share/php/PHPUnit/Util/FileLoader.php(65): include_once()
#1 /usr/share/php/PHPUnit/Util/FileLoader.php(49): PHPUnit\Util\FileLoader::load()
#2 /usr/share/php/PHPUnit/Runner/StandardTestSuiteLoader.php(43): PHPUnit\Util\FileLoader::checkAndLoad()
#3 /usr/share/php/PHPUnit/Runner/BaseTestRunner.php(146): PHPUnit\Runner\StandardTestSuiteLoader->load()
#4 /usr/share/php/PHPUnit/Runner/BaseTestRunner.php(112): PHPUnit\Runner\BaseTestRunner->loadSuiteClass()
#5 /usr/share/php/PHPUnit/TextUI/Command.php(120): PHPUnit\Runner\BaseTestRunner->getTest()
#6 /usr/share/php/PHPUnit/TextUI/Command.php(96): PHPUnit\TextUI\Command->run()
#7 /usr/bin/phpunit(73): PHPUnit\TextUI\Command::main()
#8 {main}

Next PHPUnit\TextUI\RuntimeException: Class "Tests\TestCase" not found in /usr/share/php/PHPUnit/TextUI/Command.php:98
Stack trace:
#0 /usr/bin/phpunit(73): PHPUnit\TextUI\Command::main()
#1 {main}
  thrown in /usr/share/php/PHPUnit/TextUI/Command.php on line 98
Script phpunit handling the test event returned with error code 255
~~~

> That might be all that's needed to be honest, most of the logic is handled in the repos so we can share it across the API and normal app code, we just have to make sure those properties are accepted through validation.

Ok I've added the two lines that seemed right to me.

> I'm not mad at all, happy to take the time where possible to get others involved and to help folks learn. Just feel free to let me know if you get to a blocking point, need help, or want me to take on parts.

I'm very glad about that. I just commented on some of your comments to see if you can adopt them.
I have processed all other comments to the best of my knowledge.

ssddanbrown:
>  I always get the following error:

Ah, you may still have PHP (composer) dependancies installed for production.
Try running `composer install`, you should see a few extra things get installed.

By default, we state to run `composer install --no-dev` which avoids autoloading test code, or downloading test/development requirements.

Man-in-Black:
Thanks for the clarification. It realy installed some new packages.
But after I installed them my bookstack instance stopped working with an error 500.
Maybe because of this I'm getting a bunch or errors while running the tests.

I think at the moment I've reached my limits in this regards and I do need to learn more about it before writing a valid test case.

ssddanbrown:
Thanks again @Man-in-Black for offering this and progressing it upon my requests.
I've now merged it in after taking on the remaining parts you wanted me to take on, or you reached your limit with.

My further changes can be seen in 4137cf9c8fc1f9088f8792b64fcc889f787e8958 and 43a72fb9a5dd40885c70c41bc578bffbec9c25ee.

Man-in-Black:
Thanks @ssddanbrown for taking on this.
After I reviewed your changes it now seems quite obvious to me what you have done.
Thanks again and I'm very happy that I got the opportunity to contribute to this great product which everyone I know is loving and is also using it after my recommendation.

Issue #4156, opened 2023-04-03
Increase resolution of Draw.io diagrams


vincentbernat:
### Describe the feature you'd like

Draw.io diagrams should be generated at double the resolution.

### Describe the benefits this would bring to existing BookStack users

This will make diagrams look nicer on HiDPI devices. Currently, they look a bit blurry, notably the fonts. This may also be noticeable on regular displays.

### Can the goal of this request already be achieved via other means?

Another possibility would be to render them as SVG (see #1170). As we don't need them to be interactive, they can be rendered in `<img>` tags (but this does not solve all the problems reported in #1170).

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundemental request

### How long have you been using BookStack?

0 to 6 months

vincentbernat:
As a proof of concept, I tried this patch:

```diff
diff --git a/resources/js/services/drawio.js b/resources/js/services/drawio.js
index 67ac4cc0ecb3..4b1389afa6df 100644
--- a/resources/js/services/drawio.js
+++ b/resources/js/services/drawio.js
@@ -55,7 +55,7 @@ function drawEventExport(message) {
 }
 
 function drawEventSave(message) {
-    drawPostMessage({action: 'export', format: 'xmlpng', xml: message.xml, spin: 'Updating drawing'});
+    drawPostMessage({action: 'export', format: 'xmlpng', scale: 2, xml: message.xml, spin: 'Updating drawing'});
 }
 
 function drawEventInit() {
diff --git a/resources/js/wysiwyg/plugin-drawio.js b/resources/js/wysiwyg/plugin-drawio.js
index 9f4a065adf86..57fbb9629795 100644
--- a/resources/js/wysiwyg/plugin-drawio.js
+++ b/resources/js/wysiwyg/plugin-drawio.js
@@ -26,7 +26,7 @@ function showDrawingManager(mceEditor, selectedNode = null) {
                 pageEditor.dom.setAttrib(selectedNode, 'drawio-diagram', image.id);
             });
         } else {
-            const imgHTML = `<div drawio-diagram="${image.id}" contenteditable="false"><img src="${image.url}"></div>`;
+            const imgHTML = `<div drawio-diagram="${image.id}" contenteditable="false"><img src="${image.url}" style="with:50%;height:50%;"></div>`;
             pageEditor.insertContent(imgHTML);
         }
     }, 'drawio');
```

It does not work:

[...]

Issue #3743, opened 2022-09-21
Enable HighDPI images from Diagrams.net


skaag:
### Describe the feature you'd like

Right now if you own a HighDPI monitor the images returned from Diagrams.net look fuzzy. To solve this we need to request HighDPI image generation from Diagrams.net

### Describe the benefits this would bring to existing BookStack users

This would make pages more readable on HighDPI displays (aka "Retina" displays)

### Can the goal of this request already be achieved via other means?

Not without some hacking...

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundemental request

### How long have you been using BookStack?

0 to 6 months

ssddanbrown:
Some useful discussion in closed duplicate #4156

hbertsch:
Have the same issue on high res monitors (which almost all new monitors and devices come equipped with). Especially for diagrams with a lot of content the blurriness is so high, that the information can not properly be read. Would be great to have the option to increase diagram renderings

ssddanbrown:
I've just been doing some investigation into this for a BookStack support service customer, so sharing my findings to ensure I don't forget them:

- The rasterization to PNG during the export inherently introduces some softness. If possible, it may be worth looking into how the diagrams.net PNG rasterization is done, to see if we can achieve something there.
   - This might just be the nature of mapping & scaling to pixels during export, and differences of actual text going via text rendering.
- Most blur gets introduced due to browser scaling of the image as soon as the image goes beyond the bounds of the content area and needs to be resized. 
- Exporting at 2/3x scale can help (assuming the extra complication of handling sizes can be managed) but it's not a full solution since you'd still get browser scaling at width limits. Mostly helps for smaller-width drawings on HiDPI screens. Can help add a little clarity for those other cases at >100% monitor scales, but not perfect.
- Something I haven't looked into: potentially extract vector client side and replace in browser. Might not SVG in png, or might not be accessible. 
  - Could store vector alongside, and swap back-end, but complexities and edge-cases there.

skaag:
I believe this is the solution for this problem:

https://www.drawio.com/doc/faq/export-higher-resolution

If you can simply pass that setting over, we should be getting back PNGs at 200% or 300% of the current resolution (as shown on that page). This would be a wonderful fix, since most laptops and desktops these days have high-dpi displays and the blurred images are inadequate.

mgabor3141:
[...]

PR #4805, opened 2024-01-26
Enhanced diagrams.net export functionality

cody0704:
### Description

This Pull Request introduces the editor-drawio::export event to the BookStack platform, following the discussion in issue #4156. The primary goal of this enhancement is to provide additional flexibility and customization for the export settings of the diagrams.net integration within BookStack.

### New Public Event - editor-drawio::export

* This event allows for dynamic modification of the diagrams.net export parameters. It is triggered during the export action in the diagrams.net editor embedded within BookStack.
* Users can now customize export settings based on their specific requirements, such as adjusting the scale, format, or any other diagrams.net supported parameters.

### Testing and Validation:
* The functionality has been tested by implementing custom scripts in the 'Custom HTML Head Content' section of BookStack.
* Specifically, I verified that adjusting the scale parameter of the export configuration through the new event works as expected, without any issues.

How to Test:

To test this feature, you can add the following script to the 'Custom HTML Head Content' section in BookStack:

```script
window.addEventListener('editor-drawio::export', event => {
    const config = event.detail.config;
    config.scale = 2; // Adjust the scale as needed
});
```

This script will dynamically adjust the export scale of diagrams.net drawings.

### DIFF
--- a/dev/docs/javascript-public-events.md
+++ b/dev/docs/javascript-public-events.md
@@ -96,6 +96,27 @@ window.addEventListener('editor-drawio::configure', event => {
 });
 ```
 
+### `editor-drawio::export`
+
+This event is triggered when an export action is initiated in the embedded diagrams.net drawing editor. It allows for dynamic configuration of the export parameters of diagrams.net, offering greater flexibility and control over the export process.
+See [this diagrams.net page](https://www.drawio.com/doc/faq/embed-mode) for details on the available options for the configure event.
+
+#### Event Data
+
+- `config` - The configuration object used for the export function in diagrams.net.
+  - The default settings include format: 'xmlpng', xml: message.xml, and spin: 'Updating drawing'.
+  - You can modify this object in-place to customize the export parameters as per your requirements.
+
+##### Example
+
+```javascript
+// Set export scale parameter to 2
+window.addEventListener('editor-drawio::export', event => {
+    const config = event.detail.config;
+    config.scale = 2;
+});
+```
+
 ### `editor-tinymce::pre-init`
 
 This event is called before the TinyMCE editor, used as the BookStack WYSIWYG page editor, is initialised.
--- a/resources/js/services/drawio.js
+++ b/resources/js/services/drawio.js
@@ -21,9 +21,11 @@ function drawEventExport(message) {
 }
 
 function drawEventSave(message) {
-    drawPostMessage({
-        action: 'export', format: 'xmlpng', xml: message.xml, spin: 'Updating drawing',
-    });
+    const config = {
+        format: 'xmlpng', xml: message.xml, spin: 'Updating drawing',
+    };
+    window.$events.emitPublic(iFrame, 'editor-drawio::export', {config});
+    drawPostMessage({action: 'export', ...config});
 }
 
 function drawEventInit() {


ssddanbrown:
Thanks for offering this @cody0704.
Do you have an active use for this added public event? I just want to make sure we're only adding and maintaining added events that are useful.

I ask because I'm not sure how altering any of the export options would be useful, without needing to make further changes to the BookStack source to alter how the exports are then managed.
You indicate adjusting the scale, and reference #4156, but as far as I know that alone won't achieve #4156 without issue (unless the drawing was already at page-container-width) since drawings will just be presented 2x (or the scale ratio) bigger.

If it is for scaling, then I'm also not 100% sure if we should add specific events as workarounds (with side affects) for issues that are open (#3743) as something we'd probably want to address in the future.

cody0704:
I'm delighted to have the opportunity to discuss this with @ssddanbrown

> Thanks for offering this @cody0704. Do you have an active use for this added public event? I just want to make sure we're only adding and maintaining added events that are useful.

I indeed utilize this feature to resolve the issue of blurriness in diagrams saved in BookStack, specifically when created with draw.io. The default scale often results in unclear images, particularly for text-rich diagrams. Allowing the export parameters to be adjustable can significantly enhance the clarity of saved images. This enhancement is particularly beneficial for intricate diagrams where zooming into specific sections is common, potentially leading to blurriness. Hence, I recommend enabling user-defined adjustments, as the necessity for clarity varies based on the complexity and content of the diagrams drawn.

> I ask because I'm not sure how altering any of the export options would be useful, without needing to make further changes to the BookStack source to alter how the exports are then managed. You indicate adjusting the scale, and reference #4156, but as far as I know that alone won't achieve #4156 without issue (unless the drawing was already at page-container-width) since drawings will just be presented 2x (or the scale ratio) bigger.

After implementing this functionality, I have been actively using it, setting the scale to 3. My primary use case involves creating flowcharts, which can become quite intricate. When zoomed in, the content tends to get blurry. Adjusting the scale to 3 significantly enhances the clarity of the content. Below is an example for your reference.

### Drawio - AWS18
Screenshot 2024-01-31 at 11 32 46
### scale: default value (1) - chrome zoom 500%
Screenshot 2024-01-31 at 11 33 22
### scale: 3 - chrome zoom 500%
Screenshot 2024-01-31 at 11 33 55

This might not be a great example, but what I want to convey is that after saving in drawio, there will be more pixels, resulting in clearer images. This also relates to whether the patterns in the drawio libraries support scaling. However, at the very least, the text does support scaling

cody0704:
#  Should I respond to this question here or would it be better to reply in #3743?

> If it is for scaling, then I'm also not 100% sure if we should add specific events as workarounds (with side affects) for issues that are open (#3743) as something we'd probably want to address in the future.

I'm not certain about the exact problems users are facing. However, in my setup, which includes a MacBook Pro 13" 2019 with a Retina display, I addressed the issue using the [viewerjs](https://github.com/fengyuanchen/viewerjs?tab=readme-ov-file) library. This JS library enables image scaling, which, when used in conjunction with the increased export scale of draw.io images(resulting in higher pixel clarity as per the official draw.io documentation at https://www.drawio.com/doc/faq/export-higher-resolution which states that a higher scale results in higher DPI), significantly improves the display quality. The effectiveness of this approach is evident in the attached image.

* Custom HTML Head Content
```
<!-- viewerjs -->
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/viewerjs/1.11.3/viewer.min.css" integrity="sha512-zdX1vpRJc7+VHCUJcExqoI7yuYbSFAbSWxscAoLF0KoUPvMSAK09BaOZ47UFdP4ABSXpevKfcD0MTVxvh0jLHQ==" crossorigin="anonymous" referrerpolicy="no-referrer" />
<script src="https://cdnjs.cloudflare.com/ajax/libs/viewerjs/1.11.3/viewer.min.js" integrity="sha512-f8kZwYACKF8unHuRV7j/5ILZfflRncxHp1f6y/PKuuRpCVgpORNZMne1jrghNzTVlXabUXIg1iJ5PvhuAaau6Q==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
<script>
window.onload = () => {
  document.querySelectorAll("div#main-content a").forEach((element) => {
    if (/\.(jpg|png|gif)$/.test(element.href)) {
  element.removeAttribute("href");
    }
  })
  
  document.querySelectorAll('div#main-content img').forEach((element) => {
    new Viewer(element, {
      title: false,
      navbar: false,
      zoomRatio: 0.3,
      toolbar: {
        rotateLeft: {
          size: 'large',
        },
        zoomIn: {
          size: 'large',
        },
        reset: {
          size: 'large',
        },
        zoomOut: {
          size: 'large',
        },
        rotateRight: {
          size: 'large',
        },
      }
    });
  })
}
</script>
```

### The first image appears quite small. Clicking on it allows for enlargement to focus on specific sections.
Screenshot 2024-01-31 at 11 33 06
Screenshot 2024-01-31 at 11 34 45

PR #4828, opened 2024-02-05
Update PWA manifest

shashinma:
Changed the orientation settings in PwaManifestBuilder.php from 'portrait' to 'any'. This allows the PWA to adjust to any screen orientation, enhancing user flexibility.

> [!IMPORTANT]
> A decision was made to propose a configuration change after attempting to use BookStack on a tablet. Without this > change, webapk opens BookStack in portrait mode only.
> 
> **These changes will enhance the positive user experience for users who do not possess the required technical skills.**

### DIFF
--- a/app/App/PwaManifestBuilder.php
+++ b/app/App/PwaManifestBuilder.php
@@ -26,7 +26,7 @@ class PwaManifestBuilder
             "launch_handler" => [
                 "client_mode" => "focus-existing"
             ],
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks @shashinma, this change makes sense, although portrait would probably be most popular on mobile, I don't see why we should force people into that.
Will merge to be part of the next feature release.

PR #4831, opened 2024-02-06
Implement OIDC_AUTO_REGISTER setting

AMDHome:
## Overview

This PR adds a setting for OIDC_AUTO_REGISTER. This behaves similar to the other third party authentication setting XXXX_AUTO_REGISTER.

## Behavior

Defaults to true

When set to true everything behaves as it currently does. If an account does not exist for an OIDC login, then it will be automatically created for them. If it does exist then the user can log in.

When set to false, Oidc logins will fail if a user was not created beforehand. Failure message is set to `auth.failed`. In order to log in you must first create an account with another admin account and manually input the **External Authentication ID**.

## Why this would be helpful

I work for a university, and our department would like to implement bookstack while using our university's authentication services to handle logins. 

Problem is anyone affiliated with the university has a login, but I only want people from my department to have access to our bookstacks instance. Everyone at this university knows their own External Authentication ID so it makes it easy for us to create the accounts manually and control who has access.

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -274,6 +274,7 @@ OIDC_GROUPS_CLAIM=groups
 OIDC_REMOVE_FROM_GROUPS=false
 OIDC_EXTERNAL_ID_CLAIM=sub
 OIDC_END_SESSION_ENDPOINT=false
+OIDC_AUTO_REGISTER=true
 
 # Disable default third-party services such as Gravatar and Draw.IO
 # Service-specific options will override this option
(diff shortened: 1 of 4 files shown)


ssddanbrown:
Thanks for offering this @AMDHome, but I really try to avoid adding new options unless very much proven to be required, especially where added to meet business/process logic rather than to meet the spec when it comes to auth. This would ideally be done via an issue thread to gain feedback and allow opportunity to think about alternative options.

Could you create an issue instead to start this off? You could still link to this PR as a potential implementation example, even if closed.

ssddanbrown:
Closing in favour of alternative solution discussed in #4833

Issue #1856, opened 2020-01-22
Configurable Password Policy


Cave-Johnson:
**Describe the feature you'd like**
I would like the ability for an administrator to be able to configure a minimum requirement for a password policy. 

**Describe the benefits this feature would bring to BookStack users**
This would allow for admins to increase the minimum password complexity for users accounts increasing the overall security of the bookstack instance.
Currently it is possible for a user to set a password of "aaaaaa"

**Additional context**
With the nature of a wiki, some sensitive information can be stored within bookstack. It would be good to minimise the chance sensitive information is accidentally exposed by a weak password.

A simple implementation of this would be to have the following check boxes in the admin settings page
 - A box to set a custom minimum password length

Check boxes to enable the following
[-] Require Digits
[-] Require Uppercase and Lowercase Characters
[-] Require special characters
[-] Disable repetitive or sequential characters (e.g. ‘aaaaaa’, ‘1234abcd’)

An added bonus would be the option to integrate haveibeenpwned API to disallow the use of known compromised passwords
https://haveibeenpwned.com/API/v3

This is obviously a hotly debated topic, however a summary of the new NIST guidelines can be found here which sum it up https://spycloud.com/new-nist-guidelines/

This would also go quite nicely with https://github.com/BookStackApp/BookStack/issues/1118

makrele568:
**+1** Is there a way to change the password length from 8 to 6 characters?

HungryHowies:
+1

melat0nin:
I think this is absolutely necessary, especially for instances used by non-technical folk who might not understand the importance of strong passwords, and might not understand/be willing to use MFA. 

Personally I don't think a GUI is necessary, since those setting the policy are likely the same people setting up BookStack and therefore will likely be familiar with modifying the `.env` file.

Since these requirements are built into Laravel it seems like a no-brainer to implement them in BookStack.

ssddanbrown:
BookStack is now [managed on Codeberg](https://codeberg.org/bookstack/bookstack).

As part of this change, we're closing all existing GitHub issues and pull requests.
You'll instead find this issue on Codeberg here: 
https://codeberg.org/bookstack/bookstack/issues/1856
  
Any further discussion for this will now be done on Codeberg.
Therefore this GitHub thread will be locked.

PR #4837, opened 2024-02-12
Added password history, complexity, and expiry validations

bosaleh:
This PR implements the following validations for passwords:

1- Password history: stops the user from re-using a password they have previously used. The number of passwords to be validated can be set in the PASSWORD_HISTORY environment variable.

2- Password expiry: forces the user to reset their password after a specified period. The number of days before expiry can be set in the PASSWORD_MAX_AGE environment variable.

3- Password lock: Stops the password from being changed until a specified period has elapsed. The number of days until a password can be changed can be set in the PASSWORD_MIN_AGE environment variable.

4- Password complexity: uses Laravel's built-in password validation to enforce password policies that can be set as environment variables. Please see .env.example.complete for the complete list of complexity validations.

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -84,6 +84,27 @@ MAIL_VERIFY_SSL=true
 
 MAIL_SENDMAIL_COMMAND="/usr/sbin/sendmail -bs"
 
+# Password security configuration
+# Number of recent passwords to check against, only enforced if set
+PASSWORD_HISTORY=5
+
+# Days before a password is forced to change, only enforced if set
+PASSWORD_MAX_AGE=60
+
+# Days before a password is allowed to change, only enforced if set
+PASSWORD_MIN_AGE=2
+
+# Built-in laravel password valdation
+# min defaults to 8, rest default to false
+# For more information please refer to https://laravel.com/docs/9.x/validation#validating-passwords
+PASSWORD_MIN=8
+PASSWORD_REQUIRE_LETTERS=true
+PASSWORD_REQUIRE_NUMBERS=true
+PASSWORD_REQUIRE_MIXED_CASE=true
+PASSWORD_REQUIRE_SYMBOLS=true
+PASSWORD_REQUIRE_UNCOMPROMISED=false
+
+
 # Cache & Session driver to use
 # Can be 'file', 'database', 'memcached' or 'redis'
 CACHE_DRIVER=file
--- a/app/Access/Controllers/LoginController.php
+++ b/app/Access/Controllers/LoginController.php
@@ -74,6 +74,7 @@ class LoginController extends Controller
 
         try {
             if ($this->attemptLogin($request)) {
+                //logout and redirect to password reset page if the user's password is expired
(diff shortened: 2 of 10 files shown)


ssddanbrown:
Thanks for offering this PR @bosaleh, but to be honest I'm not keen on adding a whole range of options to support, which can add new scenarios/paths to consider, especially where much of this functionality has not been previously requested.

The most relevant existing request related to this would be #1856, to allow a configurable password policy. 
We may want to take things in a different approach rather than just providing many direct options (env or UI) to control requirements. You can comment there, or :+1: the original post, to show support.

Otherwise, for the other features, you could open specific feature requests to provide an opportunity to describe the specific benefits of adding such options, and allow community interest to be gauged over time to indicate requirement/need from our user-base.

bosaleh:
Thanks @ssddanbrown, I agree that these are niche features that few will need, but I had to implement them on our installation anyway.

If you want I could create a new PR with just the password complexity rules mentioned in #1856 as that only adds a few lines in AuthServiceProvider.

ssddanbrown:
Thanks for your understanding @bosaleh!

> If you want I could create a new PR with just the password complexity rules mentioned in https://github.com/BookStackApp/BookStack/issues/1856 as that only adds a few lines in AuthServiceProvider.

Thanks for the offer, but to be honest, I'm still not too comfortable adding a variety of options for this, especially where fairly tightly tied to the underling Laravel options. I'd want to see a bit more demand & input regarding scenario/control before adding official env/UI options.

An alternative option is that we could maybe make it easier to use [our logical theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md) to define custom password rules (may already be possible to define custom rules via this system), but not sure if there'd be active use of this if added as an option.

melat0nin:
> Thanks @ssddanbrown, I agree that these are niche features that few will need, but I had to implement them on our installation anyway.
> 
> If you want I could create a new PR with just the password complexity rules mentioned in #1856 as that only adds a few lines in AuthServiceProvider.

Would you be willing to share this as a gist, for those willing to make (and maintain) the changes in their own instance?

melat0nin:
Just found this snippet on Discord from @ssddanbrown, which does the trick: https://discord.com/channels/578552496637739008/985537128505163786/1050088597686071386

ssddanbrown:
@melat0nin Ah, sorry, I had forgot to respond to your comment above.

Edit: To help others that might come across this, here's the referenced content from Discord:

---

Can be hacked via the logical theme system.
Following the getting started part of the logical theme system docs:

Use a functions.php following this:

```php
<?php

use BookStack\Facades\Theme;
use Illuminate\Validation\Rules\Password;

Theme::listen(ThemeEvents::WEB_MIDDLEWARE_BEFORE, function() {
    Password::defaults(function() {
        return Password::min(12)
            ->mixedCase()
            ->symbols()
            ->numbers();
    });
});
```


Tweak rules as desired.
Note: This is customized logic and not officially supported, may break upon updates.
Changes may not be correctly reflected in certain errors (Maybe error mesages, API doc details etc...)

melat0nin:
@ssddanbrown no problem, I see you've been plenty busy with v24.02!

PR #4849, opened 2024-02-15
Turned off autocomplete for TOTP codes

ImMattic:
Small QOL change to turn off autocomplete when entering TOTP codes since they're one time use only.

### DIFF
--- a/resources/views/mfa/parts/verify-totp.blade.php
+++ b/resources/views/mfa/parts/verify-totp.blade.php
@@ -2,7 +2,7 @@
 
 <p class="small mb-m">{{ trans('auth.mfa_verify_totp_desc') }}</p>
 
-<form action="{{ url('/mfa/totp/verify') }}" method="post">
+<form action="{{ url('/mfa/totp/verify') }}" method="post" autocomplete="off">
     {{ csrf_field() }}
     <input type="text"
            name="code"


ssddanbrown:
Thanks for offering this @ImMattic, Seems like a sensible improvement.
I just double checked the [auto-complete attribute on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete), and I saw there's a specific `one-time-code` option that looks like it's for this kind of thing.

I wonder if that would also work to prevent standard autofill, while also making it known to  browser/extension auth systems for potential autofill?

Also, thanks for the sponsorship!

ImMattic:
No problem! I'm always a big proponent of supporting open source whenever possible, especially if I use it.

I didn't know about that OTP attribute. I did just try it on my instance though and unfortunately it doesn't prevent autofill (at least on Firefox). I definitely see how having the OTP attribute would help extensions like Bitwarden to autofill the OTP though. Is it maybe possible to stack attributes together?

ssddanbrown:
Thanks @ImMattic,
Yeah, looking further it's not a supported attribute in Firefox. 
Setting `off` at the form level, then `one-time-code` at the input level seems to still prevent autofill (tested in Firefox and chrome) while still having `one-time-code` there for potential other helpful usage.

Additional tweaks made in bc24a1360f8a1f7e3be3656de9392a0802328d90 and d5a689366c7ceca8f7949caf0d919bb13fb56499, which also copies these changes for backup code MFA, and adds some tests to cover.

Now all merged to be part of the next patch release.

Issue #4845, opened 2024-02-14
Make it easier to clear the formatting of tables


ssddanbrown:
Our clear formatting option does work on tables, but only the cell contents where the cells are selected, which can leave akward styles left. It is possibly to clear all table formatting to reset it, but you have to know to drag-select from the block after the table to cover the whole table. This may be better as a table toolbar action.

Note: Ensure we retain table structure elements like merged cells.

ssddanbrown:
Added via #4850

Issue #4844, opened 2024-02-14
Table cell selection causes scrollbars to appear


ssddanbrown:
Specifically on Chromium/Edge.
Doesn't seem to be an issue on Firefox from a quick test.

![image](https://github.com/BookStackApp/BookStack/assets/8343178/37b54f6d-e60b-4159-ae41-9ee665cf3ddd)

Issue #4843, opened 2024-02-14
WYSIWYG Text direction & Alignment controls can fail when set in nested structures


ssddanbrown:
Specifically in tables where there may be subformatting.
Especially when copied in from other apps like word which may bring more complex content structures.

Probably best if, on text direction or alignment control usage, we traverse to closest block element to apply it, then remove all alignment/direction (attribute and styles) of all child elements, to keep these block-level.

Additionally, text direction controls don't currently apply during multi-table cell selections, only the first.

ssddanbrown:
Addressed in c290d01adbbdeeb68c4b0d791fe919a02432d457.
Probably not perfect, but now combined with the better clear formatting controls, this should provide the user much more control, or ways to clean & gain control.

Issue #985, opened 2018-08-23
WYSIWYG - Easier header rows


ssddanbrown:
At the moment, to set a header on a table you have to click into the row > Click the table options in the top toolbar > Row > Row Properties > Change row type to header.

We should look to either make the first row, if more than one row, be a header or make it easier to set a header row.

Might be limited on TinyMCE control, Possibly hook into such an event like this:
https://github.com/tinymce/tinymce/blob/master/src/plugins/table/main/ts/api/Events.ts#L14

Might be tricky.

alfonsrv:
Confluence solved this by having the first row be a header by default and then adding a dedicated menu row for then a table is currently selected.

Roemer:
I added a "hack" that adds a button for toggling row/column headers in the context menu of the table and at the end of the main toolbar (couldn't figure out a way to add it to the table menu of the main toolbar yet).
The hack is here: https://github.com/BookStackApp/hacks/issues/7

Edit: As an alternative, a second row similar to Confluence can be added. This would look like this:
![image](https://github.com/BookStackApp/BookStack/assets/393641/8a84ea11-07e7-48de-bdfc-d6cf61aa5350)
and is done with this customization:
```js
<script>
  window.addEventListener('editor-tinymce::pre-init', event => {
    const mceConfig = event.detail.config;
    mceConfig.toolbar = [
      mceConfig.toolbar,
      'table | tableinsertrowbefore tableinsertrowafter tabledeleterow | tablecutrow tablecopyrow tablepasterowbefore tablepasterowafter | tableinsertcolbefore tableinsertcolafter tabledeletecol | tablemergecells tablesplitcells | tablerowheader tablecolheader | tablecellbackgroundcolor | tabledelete'
    ];
      
    mceConfig.table_toolbar = 'tableprops tabledelete | tableinsertrowbefore tableinsertrowafter tabledeleterow | tableinsertcolbefore tableinsertcolafter tabledeletecol | tablerowheader tablecolheader';
  });
</script>
```

ssddanbrown:
Thanks for the input and sharing of hack/workaround @Roemer, this promoted me to put in a solution for the next release.
Within 8f1d8cef9e71a324b599f012622d7be8f8e1f09a the header row toggle is dynamically shown as part of the existing table toolbar, when within the first row. Unlike your hack I havn't added the column header toggle, as I've not really seen much demand for header columns.

platnub:
[...]

PR #4850, opened 2024-02-15
Range of WYSIWYG Editor Table Handling Improvements

ssddanbrown:
Range of fixes and improvements to handling tables, especially in regard to cleaning complex scenarios from tables imported from apps like word. We won't support all word-like content/formatting, but we can make handling that easier.

Related to #4845, #4844, #4843, #985

### Changes

- Updates TinyMCE to latest version.
- Updates cell selection styles to avoid overflow scrolling on chromium browsers.
- Adds some custom clear-format action handling to properly clear table cell styles on cell range selection.
- Adds "Clear table formatting" menu option to remove styles/sizes from the table and everything within.
- Adds "Resize to contents" menu option to remove just sizes from the table and everything within.

### Todo

- [x] See if we can dynamically show table header row toggle on first row selection.
- [x] See if we can alter table menu to add:
  - [x] Clear all table formatting option.
  - [x] Clear all table sizing option (fit cells to content).
- [x] Fully test on edge in Win11 environment, with copied content from Word.


ssddanbrown:
Preview of new table format clear and table resize menu options:

[Screencast from 2024-02-16 14-40-22.webm](https://github.com/BookStackApp/BookStack/assets/8343178/520443b6-aeb8-4a7a-b7d3-010282ef0f00)

Roemer:
This is already pretty cool, a few questions:
- Any specific reason that you didn't also add `tablecolheader` to make columns headers?
- Why allow headers only on the first row? Others could be headers too.
- I am trying to use your `tableclearformatting` and `tableclearsizes` in a custom header like:
```
    mceConfig.toolbar = [
      mceConfig.toolbar,
      'table tableinsertdialog tableclearformatting tableclearsizes | tableinsertrowbefore tableinsertrowafter tabledeleterow | tablecutrow tablecopyrow tablepasterowbefore tablepasterowafter | tableinsertcolbefore tableinsertcolafter tabledeletecol | tablemergecells tablesplitcells | tablerowheader tablecolheader | tablecellbackgroundcolor | tabledelete'
    ];
```
but it does not add the items, any idea here? Are they only menu items?

ssddanbrown:
> Any specific reason that you didn't also add tablecolheader to make columns headers?

Because I hadn't seen any demand for it. I try to avoid any additions without clear need, even if simple to add.

> Why allow headers only on the first row? Others could be headers too.

Because this is the most significant use-case. You can still make other rows headers if needed via the menu system, but the aim was to make the common case accessible. Plus it keeps the toolbar minimally complex unless there's a mucher higher chance of needing that action.

> but it does not add the items, any idea here? Are they only menu items?

Yeah, don't think the main toolbars directly support having menu items added to. 
You'd need to create buttons that have the same actions (exec the relevant commands).

PR #4895, opened 2024-03-12
Dockerfile improvements

C0rn3j:
Don't cache 50MB of lists, remove two layers, use heredoc for sanity, don't install compose in 5 steps, make the apt install list obvious and remove redundant flags.


ssddanbrown:
Thanks @C0rn3j, now merged!

PR #4904, opened 2024-03-16
15KB lossless optimization via oxipng(PNG) and svgo(SVG)

C0rn3j:
It is not much, but it is not nothing.


ssddanbrown:
Thanks for this @C0rn3j.
Before I merge, please could you remove any changes from the `tests/` directory?
Also, please could you share the commands/options/configurations/script used for the conversions made?  Would be good to have that as a reference for how this optimization was done, and should be useful to have on-hand for future changes.

C0rn3j:
> Before I merge, please could you remove any changes from the tests/ directory?

Sure thing, will do so when I find a bit of time

> Also, please could you share the commands/options/configurations/script used for the conversions made?

See ~~https://invent.kde.org/plasma/plasma-workspace-wallpapers/-/merge_requests/17#note_895818~~ https://rys.rs/optimization

C0rn3j:
Took out the test dir

C0rn3j:
Found out GIFs can also be losslessly optimized, so another 2KB down on the loading animation, plus 0.6KB on the existing PNG optims due to additional --zopfli flag.

ssddanbrown:
Thanks @C0rn3j, looks good! Will merge for next feature release.

PR #4970, opened 2024-04-22
Honeypot against Bot registrations

nesges:
This PR introduces a form-field named "username" to the registration form. This field must not be filled, but it acts as a honeypot for bots. If it is filled the Validator of RegisterController would invalidate a registration attempt. To the human viewer the honeypot is made invisible through css.

The CSS deliberately omits `display:none` because it can be assumed that bots would easily recognize this. Instead it used a similar CSS like Bootstraps visually-hidden class.

Caveat: To ensure that the form remains accessible, aria-hidden=true is used to tell screenreaders to ignore the honeypot. Bots may be programmed intelligently enough to recognize the attribute and ignore the field. But I do think that accessibility comes first and I haven't seen a bot that recognizes aria-hidden yet

### DIFF
--- a/app/Access/Controllers/RegisterController.php
+++ b/app/Access/Controllers/RegisterController.php
@@ -87,6 +87,7 @@ class RegisterController extends Controller
             'name'     => ['required', 'min:2', 'max:100'],
             'email'    => ['required', 'email', 'max:255', 'unique:users'],
             'password' => ['required', Password::default()],
+            'username' => ['prohibited'], // this is a honeypot for bots that must not be filled in
         ]);
     }
 }
--- a/resources/sass/_forms.scss
+++ b/resources/sass/_forms.scss
@@ -400,7 +400,7 @@ input[type=color] {
     border:none !important;
     overflow: hidden !important;
     clip: rect(0,0,0,0) !important;
-	white-space: nowrap !important;
+    white-space: nowrap !important;
 }
 
 .title-input input[type="text"] {
--- a/resources/sass/_forms.scss
+++ b/resources/sass/_forms.scss
@@ -389,6 +389,17 @@ input[type=color] {
   }
 }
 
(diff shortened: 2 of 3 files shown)


ssddanbrown:
Thanks for offering this PR @nesges,
I'm not too keen on playing cat-and-mouse with bots, but since a honeypot like this is quite simple I think this would be a reasonable default addition.

A couple of questions:

- Is the use of the `ambrosia-container` class name specific or significant in any way? I understand you need to uniquely target that section but just want to understand that in case of future changes.
- Would you be able to add a couple of [Registration Test cases](https://github.com/BookStackApp/BookStack/blob/development/tests/Auth/RegistrationTest.php)? Just to ensure the field is on the page, and the request is rejected on that field being filled. ([extra guidance on our tests here](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md))

nesges:
The class name `ambrosia-container` is totally arbitrary and can easily be swapped. I just wanted to avoid naming it `honeypot` because I'm always overstating bot cleverness. 

Will gladly look into the Test cases and add some to RegistrationTest. However, I'm just getting ready to go on vacation and not sure if I can do it at short notice

ssddanbrown:
@nesges thanks for confirming on the class name.

> Will gladly look into the Test cases and add some to RegistrationTest. However, I'm just getting ready to go on vacation and not sure if I can do it at short notice

Okay, don't add stress before your vacation then, I can throw these in during testing/review.
Enjoy the vacation!

nesges:
That's kind of you, thanks! :-)

ssddanbrown:
Thanks again @nesges, now merged for next feature release with some testing and minor tweaks made in 5c28bcf8651474ffd7f15831e3de705aae5a7ef8.

Issue #5029, opened 2024-05-23
GIF Thumbnail creation fails, breaking the image picker window


bytebone:
### Describe the Bug

When uploading a gif, Bookstack often fails to extract/generate a thumbnail image. This is not communicated in the GUI or the realtime application logs. But it is visible on the filesystem, where the original file is saved, but no thumbnail gets created in the respective subfolder, and an error gets logged in the laravel.log file (see below).
Once this has happened, the image picker no longer loads any images and gets stuck on an infinite loading animation. In the dev panel, after about a second of the image picker being open, the gallery endpoint returns a 502 error code. 

To resolve the issue, the original gif file needs to be deleted from the uploads folder. The image picker will load again and display a broken image item, which can then be deleted from the picker/database. 

### Steps to Reproduce

1. Go to any page and start editing
2. open the image picker
3. drag a gif into the picker for uploading
4. The upload finishes but does not let you place the image in the page
5. Close the image picker, then open it again
6. The images no longer load

### Expected Behaviour

Should be obvious

### Screenshots or Additional Context

I am using the Linuxserver Docker image to run Bookstack, but by the nature of the error assume that its not an issue with their Docker wrapper, but with the core app. 

Error Log:

<details><summary>Details</summary>
<p>

```
[2024-05-22 15:08:18] production.ERROR: BookStack\Uploads\ImageRepo::getEntityFiltered(): Argument #5 ($uploadedTo) must be of type int, null given, called in /app/www/app/Uploads/Controllers/GalleryIma
geController.php on line 31 {"userId":1,"exception":"[object] (TypeError(code: 0): BookStack\\Uploads\\ImageRepo::getEntityFiltered(): Argument #5 ($uploadedTo) must be of type int, null given, called i
n /app/www/app/Uploads/Controllers/GalleryImageController.php on line 31 at /app/www/app/Uploads/ImageRepo.php:79)
[stacktrace]
#0 /app/www/app/Uploads/Controllers/GalleryImageController.php(31): BookStack\\Uploads\\ImageRepo->getEntityFiltered()
#1 /app/www/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): BookStack\\Uploads\\Controllers\\GalleryImageController->list()
#2 /app/www/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(43): Illuminate\\Routing\\Controller->callAction()
[...]

PR #5030, opened 2024-05-23
Fix TypeError in GalleryImageController by Ensuring Default Value for uploadedTo

kalebalebachew:
## Problem
- The application encountered a TypeError when the uploadedTo parameter was passed as null to the getEntityFiltered method within the GalleryImageController. 
- This error was triggered during GIF uploads that failed to generate thumbnails, subsequently causing the image picker to malfunction (stuck in an infinite loading state with a 501 error) as stated one issue #5029

## Root Cause
- The list function in GalleryImageController did not enforce a non-null default for the uploadedTo parameter when calling getEntityFiltered. This oversight allowed null values to be passed, which are not compatible with the expected integer type, leading to a TypeError.

## Changes Made
- Modified the $uploadedToFilter variable assignment in the list method to default to 0 if no value is provided.
Included comments to explain the rationale behind the default value.
## Impact
- This fix prevents the server error encountered during GIF uploads and ensures the image picker continues to function correctly even if a thumbnail generation fails. It also enhances the robustness of the method by ensuring type consistency for critical parameters.

### DIFF
--- a/app/Uploads/Controllers/GalleryImageController.php
+++ b/app/Uploads/Controllers/GalleryImageController.php
@@ -22,28 +22,29 @@ class GalleryImageController extends Controller
      * Can be paged and filtered by entity.
      */
     public function list(Request $request, ImageResizer $resizer)
-    {
-        $page = $request->get('page', 1);
-        $searchTerm = $request->get('search', null);
-        $uploadedToFilter = $request->get('uploaded_to', null);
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering a PR @kalebalebachew, but I don't see this as the correct solution here.
While this may prevent that exception being thrown, the exception scenario likely indicates other issues in that environment.

PR #5036, opened 2024-05-29
Fixed incorrect code shortcut reference

bradenterpstra01:
I believe I have identified a small error in the TinyMCE shortcut description for insert inline code.

Ctrl+8 instead of Ctrl+Shift+8

I assume Mac is the same but I do not have a Mac to test with.


ssddanbrown:
Thanks for fixing this @bradenterpstra01! Will merge for next patch release.

PR #5052, opened 2024-06-04
Update docker-compose.yml

michaelortnerit:
Remove the version: because it is obsolete. See: https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-optional


ssddanbrown:
Thanks @michaelortnerit!

Issue #5068, opened 2024-06-13
test_frame_src_csp_header_set fails on customized ALLOWED_IFRAME_SOURCES


mueller-contria:
### Describe the Bug

`test_frame_src_csp_header_set()` (in `tests/SecurityHeaderTest.php`) tests the CspHeader->frame-src against a fixed string, identical to the default-setting for ALLOWED_IFRAME_SOURCES. 
If we customize ALLOWED_IFRAME_SOURCES in our .env, the test fails. 

Comparable Environment Variables are defined in the phpunit.xml, but unfortunately not ALLOWED_IFRAME_SOURCES

I will provide a Pull-Request for phpunit.xml soon. 

### Steps to Reproduce

1. Add a line like `ALLOWED_IFRAME_SOURCES="https://*.draw.io https://*.youtube.com https://*.youtube-nocookie.com https://*.vimeo.com https://*.google.com"` to your `.env`-file
2. Execute `vendor/bin/phpunit --filter test_frame_src_csp_header_set`
3. The test fails 
```
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'frame-src 'self' https://*.draw.io https://*.youtube.com https://*.youtube-nocookie.com https://*.vimeo.com'
+'frame-src 'self' https://*.draw.io https://*.youtube.com https://*.youtube-nocookie.com https://*.vimeo.com https://*.google.com'
```

### Expected Behaviour

The test should pass. 

The tests should not rely on the ALLOWED_IFRAME_SOURCES-value set in `.env`. Instead, `phpunit.xml` should define the value of this env-var. 

### Browser Details

CLI / PHPUnit

### Exact BookStack Version

v24.03-dev

ssddanbrown:
Closed with merge of #5069, thanks again @mueller-contria!

PR #5069, opened 2024-06-13
Add ALLOWED_IFRAME_SOURCES to phpunit.xml

mueller-contria:
Fix for bug #5068
test_frame_src_csp_header_set fails when .env-file has customized ALLOWED_IFRAME_SOURCES


ssddanbrown:
Thanks @mueller-contria! Will merge now for next patch release.

PR #5086, opened 2024-06-30
Fix draw.io/diagrams.net "export" function

### LINKED ISSUE #4710: When viewing/editing a draw.io drawing, user is not able to export draw.io drawing in vsdx format on Safari
### Describe the Bug

When viewing/editing a draw.io drawing in a document, user is not able to export draw.io drawing in vsdx format on Safari.  The same process works on Chrome.

### Steps to Reproduce

1. Go to a page that contains a draw.io drawing and choose to Edit the page.
2. Double click on the drawing to go to "Edit" mode in embedded draw.io tool.
3. In draw.io took, go to "File"->"Export as"->"VSDX(beta)"
4. In the popup window, enter the file name in "Save As:", then click on "OK" button.
5. Nothing happens.  File is not saved

### Expected Behaviour

It is supposed to save/download a copy of the draw.io drawing in vsdx format.

_No response_

### Browser Details

Safaro Version 17.1.2 (19616.2.9.11.12)

### Exact BookStack Version

v23.08.3

vincentbernat:
Exporting a diagram is done using the "blob:" scheme. This is not allowed by the current CSP. Add it automatically when we have draw.io/diagrams.net integration enabled.

Fix #4710

### DIFF
--- a/app/Util/CspService.php
+++ b/app/Util/CspService.php
@@ -143,6 +143,7 @@ class CspService
             $drawioSourceParsed = parse_url($drawioSource);
             $drawioHost = $drawioSourceParsed['scheme'] . '://' . $drawioSourceParsed['host'];
             $hosts[] = $drawioHost;
+            $hosts[] = 'blob:';
         }
 
         return $hosts;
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this @vincentbernat, but when looking before I found it hard to understand the full potential security impact of adding the blob scheme to CSP (which is why I haven't added it so far).
Where there's a non-understood risk, I'd rather that this be opt in (via adding to `ALLOWED_IFRAME_SOURCES`) rather than enabled by default.

ssddanbrown:
I'll go ahead and close this off under the reasoning provided above.

PR #5096, opened 2024-07-03
Wraps file extension comparison components in strtolower()

### LINKED ISSUE #5095: Image replacement is case-sensitive when it should not be
### Describe the Bug

When replacing an image in the image editor, if the case of the extension differs the replacement fails.

### Steps to Reproduce

1. Upload an image named "file.PNG"
2. Attempt to replace the image with "newfile.png"
3. Observe the error

### Expected Behaviour

As `png` and `PNG` are not indicators of different files, you should be able to replace a file ending in PNG with one ending in png.

_No response_

### Browser Details

_No response_

### Exact BookStack Version

v24.02.2

DanielGordonIT:
Created to fix #5095 
This avoids the issue where replacing file.PNG with newfile.png fails due to "PNG" not being equal to "png"

### DIFF
--- a/app/Uploads/ImageRepo.php
+++ b/app/Uploads/ImageRepo.php
@@ -166,7 +166,7 @@ class ImageRepo
      */
     public function updateImageFile(Image $image, UploadedFile $file): void
     {
-        if ($file->getClientOriginalExtension() !== pathinfo($image->path, PATHINFO_EXTENSION)) {
+        if (strtolower($file->getClientOriginalExtension()) !== strtolower(pathinfo($image->path, PATHINFO_EXTENSION))) {
             throw new ImageUploadException(trans('errors.image_upload_replace_type'));
         }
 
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks @DanielGordonIT for providing this!

If you're confident with such a codebase, it would also be good to have a test to cover this scenario.
We have [some test guidance here](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md).
Here's an existing test related to this scenario, so an added test would be similar:

https://github.com/BookStackApp/BookStack/blob/14837e34fb7524b8ace87a49c8b808542addd55d/tests/Uploads/ImageTest.php#L122-L139

If you're not able to add this, that's fine, just let me know and I can write the test during review.

DanielGordonIT:
I will work on creating a test for this tonight! If I am not able to make the test (PHP is not a language I am very familiar with, not that it's stopped me from using the theme system), I will let you know.

DanielGordonIT:
~~@ssddanbrown I've added what I think would be a functional test case, but please do give it a review.~~

Wait no I did not the commit didn't work

ssddanbrown:
This all looks good @DanielGordonIT, Thank you very much again for providing this and for putting in the extra effort to write a test! Merging for next patch release.

Issue #5099, opened 2024-07-04
create-page link which can be traced back


cod3monk:
### Describe the feature you'd like

We want to preprint labels for people to document their projects around our makespace. The label comes in two parts: introductions and the sticky label to place on the project for reference. The instructions should contain a link (QR-code) that brings up a create-page in a chapter to start a project documentation page. The sticky label should have a corresponding link (QR-code) for other to scan and visit the documentation page.

I could be the same url in both cases, where the page is created if it does not exist and otherwise is viewed.

Related is #3217, because many users will scan the initial (create-page) QR-code before logging in.

### Describe the benefits this would bring to existing BookStack users

Pregenerated links for creating and viewing pages, that can be passed on physical media.

### Can the goal of this request already be achieved via other means?

Maybe through a custom website that leverages the bookstack API? Not sure.

Alternatively we pre-create 100 empty pages and use regular permalinks, but it will be tedious keep an overview of empty/unused pages vs relevant ones.

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

1 to 5 years

cod3monk:
Looking into the `Page` model and `PageController`, a change to the model would be necessary to store an externally passed in identifier. Other than that the an additional controller function should be reasonably easy (I guess).

ssddanbrown:
BookStack is now [managed on Codeberg](https://codeberg.org/bookstack/bookstack).

As part of this change, we're closing all existing GitHub issues and pull requests.
You'll instead find this issue on Codeberg here: 
https://codeberg.org/bookstack/bookstack/issues/5099
  
Any further discussion for this will now be done on Codeberg.
Therefore this GitHub thread will be locked.

PR #5115, opened 2024-07-11
show-or-create for pages

cod3monk:
This PR adds two new routes:
`/books/{bookSlug}/show-or-create?name={pageSlug}` and
`/books/{bookSlug}/chapter/{chapterSlug}/show-or-create?name={pageSlug}`

The concept is, that with these URL a user will either drop into the edit view of a new Page, or -if the page already exists- be redirected to the page itself.

This solves the feature request mentioned in #5099.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -79,19 +79,13 @@ Big thanks to these companies for supporting the project.
 </a></td>
 </tr>
 <tr>
-<td align="center" colspan="2"><a href="https://practinet.be/" target="_blank">
+<td align="center"><a href="https://practinet.be/" target="_blank">
     <img width="240" src="https://media.githubusercontent.com/media/BookStackApp/website/main/static/images/sponsors/practinet.png" alt="Practinet">
 </a></td>
+<td align="center"><a href="https://transporttalent.com" target="_blank">
+    <img width="240" src="https://media.githubusercontent.com/media/BookStackApp/website/main/static/images/sponsors/transport-talent.png" alt="Transport Talent">
+</a></td>
 </tr></tbody></table>
-
-## 🛣️ Road Map
-
-Below is a high-level road map view for BookStack to provide a sense of direction of where the project is going. This can change at any point and does not reflect many features and improvements that will also be included as part of the journey along this road map. For more granular detail of what will be included in upcoming releases you can review the project milestones as defined in our [Release Process](dev/docs/release-process.md) documentation.
-
-- **Platform REST API** - *(Most actions implemented, maturing)*
-    - *A REST API covering, at minimum, control of core content models (Books, Chapters, Pages) for automation and platform extension.*
-- **Permission System Review** - *(In Progress)*
-    - *Improvement in how permissions are applied and a review of the efficiency of the permission & roles system.*
 
 ## 🛠️ Development & Testing
 


ssddanbrown:
Thanks for offering this @cod3monk, but I'm really not keen on maintaining & supporting new endpoints that are built/intended for specific external business processes.

These kinds of additions would be a good use case of using the [logical theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md) to add your own endpoints [as somewhat reflected here](https://www.bookstackapp.com/blog/bookstack-release-v23-12/#logical-theme-system-events-to-register-routes), which would allow you to add these custom endpoints to your own instance without altering core app code.

cod3monk:
Would you be willing to include a `?createIfNotFound` property in the show endpoints to redirect to an edit view on `NotFoundException`?

I am working on porting the endpoint to the logical theme system, which turns out to be rather tedious. Due to a lack of understanding of the `Facades` and thus how to port my original `showOrCreate` PageController. Documentation seems to be very sparse or too abstract on bookstack and laravel side.

ssddanbrown:
> Would you be willing to include a ?createIfNotFound property in the show endpoints to redirect to an edit view on NotFoundException?

No.

Here's my take on a port of your changes to a logical theme `function.php` file (tested on my dev instance which is pretty much matching the current release in terms of back-end logic):

<details><summary>functions.php</summary>
<p>

```php
<?php

use BookStack\Entities\Queries\EntityQueries;
use BookStack\Entities\Repos\PageRepo;
use BookStack\Exceptions\NotFoundException;
use BookStack\Facades\Theme;
use BookStack\Http\Controller;
use Illuminate\Http\Request;
use Illuminate\Routing\Router;

class CustomPageActionsController extends Controller
{
    public function __construct(
        protected PageRepo $pageRepo,
        protected EntityQueries $entityQueries,
    ) {
    }

    /**
     * If page exists, redirect to show page, otherwise to create page
     */
    public function showOrCreate(Request $request, string $bookSlug, string $chapterSlug = null)
    {
        $data = $this->validate($request, [
            'name' => ['required', 'string', 'max:255'],
        ]);

        $pageName = $data['name'];

        try {
            // Redirect to the page if an existing page is found
            $page = $this->entityQueries->pages->findVisibleBySlugsOrFail($bookSlug, $pageName);
            return redirect($page->getUrl());
        } catch (NotFoundException $e) {
            // Otherwise create a new page and start editing it
            if ($chapterSlug) {
                $parent = $this->entityQueries->chapters->findVisibleBySlugsOrFail($bookSlug, $chapterSlug);
            } else {
                $parent = $this->entityQueries->books->findVisibleBySlugOrFail($bookSlug);
            }

            $this->checkOwnablePermission('page-create', $parent);

            $page = $this->pageRepo->getNewDraftPage($parent);
            $this->pageRepo->publishDraft($page, [
                'name' => $pageName,
            ]);

            return redirect($page->getUrl('/edit'));
        }
    }
}

// Register the routes for our custom controller & action
Theme::listen(ThemeEvents::ROUTES_REGISTER_WEB_AUTH, function (Router $router) {
    $router->get('/books/{bookSlug}/show-or-create', [CustomPageActionsController::class, 'showOrCreate']);
    $router->get('/books/{bookSlug}/chapter/{chapterSlug}/show-or-create', [CustomPageActionsController::class, 'showOrCreate']);
});
```

</p>
</details> 

I've made changes to the method code but the general logic/routes/endpoints/request-data should be the same with the desired overall output.

cod3monk:
Thanks! Code-wise I had almost the same, but it was missing some of the `use` imports, thus running into using `Facade` versions of the same objects and not catching the correct `NotFoundException`.

Issue #1712, opened 2019-10-12
Sorting option „Popularity“


Wookbert:
**Describe the feature you'd like**
In the Books view (/books) the left side navigation panel lists popular books. Assuming that this popularity is measured by how often a book gets opened/loaded, I would find it extremely useful to have this as a sorting option in the Books view.

**Describe the benefits this feature would bring to BookStack users**
IMO sorting by popularity is a much better (or should I say: best) sorting option, when you have many books to deal with.

ssddanbrown:
Thanks for the request here @Wookbert, but since this has had very little further request/desire I'm going to go ahead and close this off.

PR #5132, opened 2024-07-19
Add popularity sorting options for books

fuszti:
Now the user can sort the /books by the view_count. It is the #1712 feature request.

Unit tests are added too.

### DIFF
--- a/app/Entities/Controllers/BookController.php
+++ b/app/Entities/Controllers/BookController.php
@@ -40,20 +40,10 @@ class BookController extends Controller
     public function index(Request $request)
     {
         $view = setting()->getForCurrentUser('books_view_type');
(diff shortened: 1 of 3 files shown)


fuszti:
@CodiumAI-Agent /review

ssddanbrown:
Thanks for offering this @fuszti and sorry for the late response. Personally I'm not too convinced of the value of adding this feature, since there's been little desire for this so far and we already list a few of the most popular books in this view. Plus if we add this here it would also be expected in the shelves view, adding a little complexity there too.

ssddanbrown:
I'm going to go ahead and close this off as per my comment above.

fuszti:
Oh wow, I somehow missed your previous comment. Thank you for responding to my PR. I understand your point, no worries.

PR #5153, opened 2024-08-06
Update sponsor image URLs in readme

Alexander-Wilms:
In the rich diff, one can see that some of the images are still broken:

image

Codeberg returns an error 500 for these.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -49,7 +49,7 @@ Big thanks to these companies for supporting the project.
 
 <table><tbody><tr>
 <td align="center"><a href="https://www.federated.computer/bookstack/" target="_blank">
-    <img width="480" src="https://media.githubusercontent.com/media/BookStackApp/website/main/static/images/sponsors/federated-computer.png" alt="Federated.computer">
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for letting me know @Alexander-Wilms, seems like either archiving the website repo, or being at GitHub LFS limits broke these links.

I also updated these to now just use the Bookstack website instead, rather than involving other platforms.

Alexander-Wilms:
Yes, https://[codeberg.org/bookstack/website/media/branch/main/static/images/sponsors/cloudabove.png](https://codeberg.org/bookstack/website/media/branch/main/static/images/sponsors/cloudabove.png) shows

```
version https://git-lfs.github.com/spec/v1
oid sha256:8a45011ad86493b0718ee9628947e1ad9363c26c9291122a70956f6494474bb9
size 21923
```

ssddanbrown:
@Alexander-Wilms Yeah, currently a bit stuck. Don't think I had all LFS data pulled locally, so it's not all on Codeberg yet, but I can't pull from GitHub due to being over the LFS bandwidth limit (which you can't control on a public repo). Waiting for that limit to reset in a few days time :cry:

Issue #4271, opened 2023-05-26
OIDC: Support `picture` claim for use as user avatar


Ghost-chu:
### Describe the Bug

Although the OIDC response contains the `picture` field, the Bookstack still use default user avatar.

```json
{
  "sub": "<censored>",
  "iss": "<censored>",
  "aud": "7acd8e81792f80dc48e9",
  "preferred_username": "<censored>",
  "name": "<censored>",
  "email": "<censored>",
  "picture": "https://cdn.<censored>/casdoor/avatar/<censored>/Ghost_chu.png?t=1685018195637388715"
}
```

### Steps to Reproduce

1. Setup the OIDC for Bookstack
2. Create a new user and upload a avatar from your OIDC provider management
3. Login to Bookstack
4. Bookstack use default blue avatar as new user default avatar

### Expected Behaviour

Bookstack should use the avatar from OIDC response instead the default avatar

### Browser Details

Brave 1.51.118 Chromium: 113.0.5672.126（Release） （64 bit）

### Exact BookStack Version

v23.05.2

### Hosting Environment

debian-11.7 - Bookstack Docker Image by LinuxServer
```
      - AUTH_METHOD=oidc
      - AUTH_AUTO_INITIATE=true
      - OIDC_NAME=<censored>
      - OIDC_DISPLAY_NAME_CLAIMS=name
      - OIDC_CLIENT_ID=<censored>
      - OIDC_CLIENT_SECRET=<censored>
      - OIDC_ISSUER=<censored>
      - OIDC_ISSUER_DISCOVER=true
```

ssddanbrown:
Thanks for raising, but I have recategorised this as a feature request, and updated the title to suit, since this is not a break in existing logic. We've just never specifically supported user avatars via the picture claim.

cal940:
hello, is there any new progress on this issue? 

would be nice to see this feature in the following releases.

jasonpincin:
Plus one on this one, fwiw.

rubentalstra:
I’ve opened a merge request (#5429) adding optional support for fetching user avatars from the OIDC ```picture``` claim. It reuses our existing avatar logic so there’s no major code duplication. A new ```oidc.fetch_avatars``` config option must be enabled for this feature to take effect. Feedback and testing are welcome!

ssddanbrown:
This has now been added via  #5429 and #5626.
Thanks to @rubentalstra for providing an implementation for this.
[...]

PR #5177, opened 2024-08-24
Add support for OIDC picture

jasonpincin:
This is a naive first attempt at adding support for OIDC pictures. It works (for me) which is about all I can say about it. 

Tossing it up there in case it's useful, or for feedback. Happy to develop the PR further if there's broader interest in the feature. If I were to develop it further, I'd be curious if there's thoughts on better ways of plumbing the picture path through, and if there's any mechanism for refreshing the user's avatar that would need to be updated. It wasn't immediately evident to me that this existed.

### DIFF
--- a/app/Access/Oidc/OidcService.php
+++ b/app/Access/Oidc/OidcService.php
@@ -214,7 +214,8 @@ class OidcService
             $user = $this->registrationService->findOrRegister(
                 $userDetails->name,
                 $userDetails->email,
-                $userDetails->externalId
+                $userDetails->externalId,
+                $userDetails->picture,
             );
         } catch (UserRegistrationException $exception) {
             throw new OidcException($exception->getMessage());
--- a/app/Access/Oidc/OidcUserDetails.php
+++ b/app/Access/Oidc/OidcUserDetails.php
@@ -11,6 +11,7 @@ class OidcUserDetails
         public ?string $email = null,
         public ?string $name = null,
         public ?array $groups = null,
+        public ?string $picture = null,
     ) {
     }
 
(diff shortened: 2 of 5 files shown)


jasonpincin:
This relates to a feature request captured here: https://github.com/BookStackApp/BookStack/issues/4271

ssddanbrown:
Thanks for offering this @jasonpincin, and sorry for my delayed response. The code here looks good, but there's a few things I think we'd need to consider, or some changes to the code to avoid complexity.

### High Level Review

- Fetching resources from an external domain can be dangerous (Open to server-side-request exploitation) so this needs to be done in a considered approach, rather than being default enabled. I thought about maybe being able to define "trusted origins" for this, but that might be over complicating scope. Might just be best to have an extra simple boolean config option to whether avatars are fetched from OIDC or not. Then we can add commentary to that option in documentation regarding the risks (eg. Only enable if picture URLs would be considered trusted).

### Code Review

- A lot of methods here are gaining added complexity just to carry across the `$picture` through the whole default registration process. Instead I think it would be better if this can the be left as-is, and and extra `assignToUserFromUrl` method be added to the `UserAvatars` class, [much like this method](https://github.com/BookStackApp/BookStack/blob/4b4d8ba2a18fd6daa0face55742d50dc0230a4b4/app/Uploads/UserAvatars.php#L44-L54) which is what we use for LDAP avatar images. The `UserAvatars` class can then be injected into the `OidcService` and used directly.
- As with all added back-end features, we need to ensure it's covered with testing. There's [some guidance for this here](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md), with existing [OIDC tests here](https://github.com/BookStackApp/BookStack/blob/development/tests/Auth/OidcTest.php), but this can get complex with all the OIDC and external HTTP call mocking required so don't stress with this if it's not familiar or difficult, I can add that after if needed.

---

Would you be happy to take changes following the above? Otherwise I can continue this if needed so don't stress on it. Just let me know if anything is unclear or if you need extra guidance on anything.

ssddanbrown:
Since there's been no further follow-up, I'm going to close this off.
I'd still be happy to receive a PR though if it followed the above considerations.

jasonpincin:
Yeah sorry for the delay; happy to make the adjustments @ssddanbrown  - time just not on my side at the moment.

PR #5198, opened 2024-09-09
Add OpenSearch support

### LINKED ISSUE #5122: Add OpenSearch support

Add XML-file and HTML meta-tag so browsers can autodiscover the search of BookStack instances. See:

https://developer.mozilla.org/en-US/docs/Web/OpenSearch

### Describe the benefits this would bring to existing BookStack users

Easily add a BookStack instance as search-engine to browsers supporting this standard (like Firefox)

### Can the goal of this request already be achieved via other means?

Not easily, Firefox for example has no UI to manually add search engines, as far as I know.

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

1 to 5 years

### Additional context

I can prepare a pull request if you accept this new feature.

maximilian-walter:
Add XML-file and HTML meta-tag so browsers can autodiscover the search of BookStack instances. See:

https://developer.mozilla.org/en-US/docs/Web/OpenSearch

Closes BookStackApp/BookStack#5122

### DIFF
--- a/app/App/MetaController.php
+++ b/app/App/MetaController.php
@@ -64,4 +64,14 @@ class MetaController extends Controller
             'jsLibData' => file_get_contents(base_path('dev/licensing/js-library-licenses.txt')),
         ]);
     }
+
+    /**
+     * Show the view for /opensearch.xml.
+     */
+    public function opensearch()
+    {
+        return response()
+            ->view('misc.opensearch')
+            ->header('Content-Type', 'application/opensearchdescription+xml');
+    }
 }
--- a/lang/en/common.php
+++ b/lang/en/common.php
@@ -107,4 +107,7 @@ return [
     // Not directly used but available for convenience to users.
     'privacy_policy' => 'Privacy Policy',
     'terms_of_service' => 'Terms of Service',
+
+    // OpenSearch
+    'opensearch_description' => 'Search :appName',
 ];
--- a/resources/views/layouts/base.blade.php
+++ b/resources/views/layouts/base.blade.php
@@ -32,6 +32,9 @@
     <link rel="manifest" href="{{ url('/manifest.json') }}">
     <meta name="mobile-web-app-capable" content="yes">
 
+    <!-- OpenSearch -->
+    <link rel="search" type="application/opensearchdescription+xml" title="{{ setting('app-name') }}" href="{{ url('/opensearch.xml') }}">
+
     @yield('head')
 
     <!-- Custom Styles & Head Content -->
--- a/resources/views/misc/opensearch.blade.php
+++ b/resources/views/misc/opensearch.blade.php
@@ -1,0 +1,11 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+  <ShortName>{{ setting('app-name') }}</ShortName>
+  <Description>Search {{ setting('app-name') }}</Description>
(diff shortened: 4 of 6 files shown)


ssddanbrown:
Thanks for offering this @maximilian-walter!

Here's my feedback from a quick review (have not actually pulled down and tested yet):

- Can the images be reduced down to just the 32 & 64? Can't see evidence of large images being used here, and I'd prefer to keep things to what's required/used where possible.
- The URL has a `method` attribute, but I can't see this as [part of the spec](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-url-element)?
- The `ShortName` [must be limited to 16 characters](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-shortname-element) or less.
- The `Description` is lacking translation support.
- We'd want to have [some tests to cover this](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md), ([Simple example of tests for licenses endpoint](https://github.com/BookStackApp/BookStack/blob/development/tests/LicensesTest.php)).

Are you happy to take on these changes? Just let me know if you need further help/guidance, or if you're not able to perform further work for any reason.

maximilian-walter:
Thanks @ssddanbrown for your feedback!

> Can the images be reduced down to just the 32 & 64? Can't see evidence of large images being used here, and I'd prefer to keep things to what's required/used where possible.

I don't know if higher resolutions would be used. I would expect that, but I did not validate it. Because we have them already I would keep them, but I am also fine if you wish to reduce the number of links.

> The URL has a method attribute, but I can't see this as [part of the spec](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-url-element)?

You are right, I removed the attribute. We don't need it anyway.

> The ShortName [must be limited to 16 characters](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-shortname-element) or less.

The hardlimit of 16 characters is now enforced.

> The Description is lacking translation support.

I moved the description to the language file. I hope this was the correct way of doing it in regards of Crowdin.

> We'd want to have [some tests to cover this](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md), ([Simple example of tests for licenses endpoint](https://github.com/BookStackApp/BookStack/blob/development/tests/LicensesTest.php)).

I added some basic test based on your HtmlTest-class. Because the Symfony Crawler is used, I assumed it is fine to feed XML to it. :)

A proper XML schema validation would be much better, but I did not find any DTD file for this standard.

ssddanbrown:
Thanks @maximilian-walter! Could you just also restore the two `.gitignore` files removed in this PR?
BTW, I'll probably squash merge this to contain the feature to one commit, and prevent files being removed/re-added in git history if that's okay.

maximilian-walter:
@ssddanbrown Sorry, I added the missing files again and rebased my branch so the files do not get removed and added again. Squashing is still a good idea. I used multiple small commits only for a easier code review. :)

ssddanbrown:
Now merged for next feature release, thanks again @maximilian-walter!

While testing I found that chrome would totally ignore opensearch discovery if not hosting on the root of the domain, whereas Firefox was totally happy with that. Will have to remember this for future potential support scenarios!

Issue #1091, opened 2018-10-26
Add a "Cascade Shelf Permissions" command to run the "Copy Permissions to Books" action for all shelves


Joonake:
**Describe the feature you'd like**
I could not find any way to automatically inherit permissions from bookshelf to all books under it. There is a button in shelf permission settings but I would like to automate this so every book moved to 'secret shelf' would automatically inherit the permissions of 'secret shelf'

**Describe the benefits this feature would bring to BookStack users**
Automated inheriting would eliminate the need of clicking 'Copy Permissions to Books' after creating every new or moving existing books to shelf.

I tried to dig up the source code and found php function 'copyPermissions' but I have no idea how to call this function outside the whole framework.

ssddanbrown:
Hi @Joonake, Thank you for your request.

Bookshelf permissions do not automatically cascade for technical reasons. 
A single Book can be on multiple shelves. In this event you could get conflicting permissions. I think any auto-resolution of this could cause odd, unexpected permission behaviour. 

I can see the use-case though. Maybe, If we add a command line action to auto-set permission? If a book is in multiple bookshelves then the latter shelf would take priority. You could then set a cronjob or similar to run this action every day or so. Would that suffice?

Joonake:
Dear @ssddanbrown, thank you for your reply. Yeah just noticed that there would be a problem if book is on multiple shelves. In my case it would be sufficient if one could run shelf-specific `copyPermissions` function from command line. Then all the books on specific shelf would be protected no matter what other shelves they are on or what order they were added to each shelf.

Beeez:
I second this as an option. It would be great to at least include this as an option on the shelf permissions page. It is easy enough to create a new Book under another shelf if needed to have a duplicate. 

Perhaps if it is found that a book _is_ in fact on multiple shelves, to report that as an error noting the shelves it is a part of? And possibly giving the option to split the book into two separate books from there

ssddanbrown:
> Perhaps if it is found that a book is in fact on multiple shelves, to report that as an error noting the shelves it is a part of?

@Beeez Thanks for the suggestion but I wouldn't want to show an error for this, since this was intended to be the main unique feature of shelves.

Beeez:
@ssddanbrown Understandable. I guess we could just Copy Permissions To Books after every book is created under a shelf.

ssddanbrown:
Issue title updated to reflect the discussed resolution

joaomezzari:
[...]

Issue #4411, opened 2023-07-27
Permission Question


wedowhateverwewant:
### Attempted Debugging

- [X] I have read the debugging page

### Searched GitHub Issues

- [X] I have searched GitHub for the issue.

### Describe the Scenario

So I get that you create roles and add members to group, then add that role to the shelves, but then the permissions don't cascade down to the books and chapters and pages unless you do the " copy permission" is there a way to have whatever permission shelve has apply to everything under it to NEW and existing items automatically ? some sort of script or is there an option ? 

### Exact BookStack Version

v23.06.2

### Hosting Environment

Docker Compose
Nginx Proxy Manager
Bookstack
SAML2 for SSO

ssddanbrown:
There's no in-platform automatic way. Open request in #1596.

There is a ["Copy Shelf Permission" command](https://www.bookstackapp.com/docs/admin/commands/#copy-shelf-permission) which can do this externally from command line (including being scheduled if desired). This can run on all shelves in the system. Just be aware that books on multiple shelves will only get the permissions from one of the shelves (Whichever runs through the process later).

Alternatively, if you have some PHP skills, you could listen to book creation events via the [logical theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md) then trigger this process upon such events.

wedowhateverwewant:
okay thanks for the info. however if one edits and makes some changes to  php files and afterward theres an update to bookstack does that update erase the past changes ?

ssddanbrown:
If you're making changes to core app files, then yes, that may conflict with git during update (on a normal install) or be wiped (if in a docker container).

The logical theme system allows customizations outside of core app files, so they run as an extension to the platform.
This may still conflict with updates (if your logic is depending on internal code that's changed) but the changes should persist and they can be toggled off via "toggling off" (removing the option) the theme. They wouldn't stop the updates or be lost in the say way as changing core app files.

wedowhateverwewant:
okay this makes sense, thanks

Sulkar:
Hello, thanks for the hint. I managed to achieve the result with the following code in the theme's functions.php.

```php
Theme::listen(ThemeEvents::ACTIVITY_LOGGED, function (string $activityType, $detail) {

    $validTypes = [ActivityType::BOOK_CREATE];
    if (!in_array($activityType, $validTypes)) {
        return;
    }

    if ($detail instanceof Book) {
[...]

PR #5205, opened 2024-09-17
Add the option for new books to inherit permissions when created

### LINKED ISSUE #1596: Book Permissions
Would it be possible to have new books inherit the permissions of the shelve they are created under (assuming they are indeed created under a shelf)?

My understanding is the permissions set on shelves do not propagate unless you use the copy permissions, therefore I don't think this would cause a conflict, but it would mean there is one less step many users need to perform, by default.

anbcodes:
Adds an option to bookshelves to copy permissions to books created under them. This fixes the issue where someone has permission to create a book but cannot view the book they created. This fixes #1596 by implementing the suggestion in [this comment](https://github.com/BookStackApp/BookStack/issues/1091#issuecomment-2250465959).

### DIFF
--- a/app/Entities/Controllers/BookController.php
+++ b/app/Entities/Controllers/BookController.php
@@ -19,6 +19,7 @@ use BookStack\Facades\Activity;
 use BookStack\Http\Controller;
 use BookStack\References\ReferenceFetcher;
 use BookStack\Util\SimpleListOptions;
+use BookStack\Settings\SettingService;
 use Illuminate\Http\Request;
 use Illuminate\Validation\ValidationException;
 use Throwable;
@@ -31,6 +32,7 @@ class BookController extends Controller
         protected BookQueries $queries,
         protected BookshelfQueries $shelfQueries,
         protected ReferenceFetcher $referenceFetcher,
+        protected SettingService $settingService,
     ) {
     }
 
@@ -115,6 +117,17 @@ class BookController extends Controller
         if ($bookshelf) {
             $bookshelf->appendBook($book);
             Activity::add(ActivityType::BOOKSHELF_UPDATE, $bookshelf);
+
+            $should_inherit_perms = $bookshelf->new_books_inherit_perms ?? $this->settingService->get("app-new-books-inherit-perms");
+
+            if ($should_inherit_perms) {
+                // Copy permissions from the shelf to the book on creation
+                // https://github.com/BookStackApp/BookStack/issues/4411#issuecomment-1962302337
+                $shelfPermissions = $bookshelf->permissions()->get(['role_id', 'view', 'create', 'update', 'delete'])->toArray();
+                $book->permissions()->delete();
+                $book->permissions()->createMany($shelfPermissions);
+                $book->rebuildPermissions();
+            }
         }
 
         return redirect($book->getUrl());
--- a/app/Entities/Controllers/BookshelfController.php
+++ b/app/Entities/Controllers/BookshelfController.php
@@ -16,6 +16,7 @@ use BookStack\Util\SimpleListOptions;
 use Exception;
 use Illuminate\Http\Request;
 use Illuminate\Validation\ValidationException;
+use Illuminate\Validation\Rule;
 
 class BookshelfController extends Controller
 {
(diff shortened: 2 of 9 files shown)


ssddanbrown:
Thanks for offering this @anbcodes but I'm really not keen on introducing layers of settings for process control like this.
From an implementation point of view I'd be more open to a single option during book creation, but even that adds complication when considered with expectations around permission control, and I'm not sure it meets the desired process of many.

Part me of things we should just document a [logical theme solution on our hacks](https://www.bookstackapp.com/hacks/) since that could provide a flexible base to be used & customized without complicating anything in the core platform.

intellasoft:
Hi @ssddanbrown,

I'm helping coordinate this change that @anbcodes made that we badly need.  Right now it's impossible for our team to effectively use bookstack without this change.  And I'm not the only one with this issue, given the popularity of https://github.com/BookStackApp/BookStack/issues/1596

Default Behaviors when you segment different permissions to various shelves
- When someone creates a book, they have no access to it and need an admin to come rescue the book and set the permissions
- When someone creates a book on a shelf in for example the 'Dev' shelf.  Other 'Dev' members have no access unless an administrator goes to the copy shelf permissions to books button.  Again needing an administrator to come to the rescue.

As much as I am completely in love with Bookstack... we had none of these issues with our old system.

There's already a button to do this process manually.  Why not make it automatic for those that need it?  I know we're certainly not alone in needing automatic permissions, especially when there's a 1:1:1 shelf,book,page and nothing is re-used anywhere sort of situation most people find themselves in.

ssddanbrown:
> Why not make it automatic for those that need it? 

@intellasoft I'm not against the idea, just not keen on adding layers of process specific setting, especially as they would still not solve all desires or assumptions in the behaviour here, and could even introduce new considerations/complications.

Is there any reason using a logical theme hack would not be suitable in your scenario given that you badly require this functionality?

intellasoft:
It sounds like if there were some hooks for create book to then run a plugin for whatever the user wants to do.. that would simplify the core.

The hacks method sounds.... well... hackey.  Ie: having to drop in code or replace a script for each version does not sound ideal.

I'm a huge fan of callbacks and plugin hooks.  Maybe we should rework this to use new hooks...

ssddanbrown:
@intellasoft the logical theme system (which would be used for such a hack) is essentially a hook system. The code you use within the hooks is not deemed supported/stable, but needing to update would not be common.

You don't have to drop-in code or replace scripts on each update, the code is placed in its own folder, and does not have to be re-applied on each BookStack change.

We emit a bookshelf-update then books are added via the shelf > create-book flow, so it should be possible to hook into that event, find newly assigned & created books, then copy permissions to them.
[This comment](https://github.com/BookStackApp/BookStack/issues/4411#issuecomment-1962302337) reflects a very similar solution (although depending on slightly less stable session/process-based internal functionality).

intellasoft:
Ah got it.  And then... is there a hook to add additional configuration elements to the portal?

intellasoft:
But... I also disagree (as a user) that this should be a plugin.

I don't get the design decision that would allow for utilizing the permission system in such a way that the built-in inherent behaviors make the system unusable for affected users.  Especially since the (manual) process already exists.  We're just automating it.

Like.. what's the purpose of allowing a book to be created that the user then has no access to?

ssddanbrown:
>  And then... is there a hook to add additional configuration elements to the portal?

No. We don't generally provide hooks for UI components but, depending on what you might need to do, you may be able to override UI using the visual theme system.

> Like.. what's the purpose of allowing a book to be created that the user then has no access to?

The permission system is complex and there are likely naturally many scenarios where you could encounter logic that does not suit your process, especially where the system is used with different intentions from their original design (using shelves as top-level permission containers / owners of books, using permissions to open access rather than to lock access). 
It's not really a goal to ensure every one of these scenarios is addressed via options, since that's a road to added complexity and often would introduce complications of their own.

I'd rather provide flexible customization solutions (logical theme system) for those that need it rather than complicate the core system to have options for the different processes, locking shelves further into a role they're not really built/intended for.

ssddanbrown:
I'm going to go ahead and close this off as per the reasoning provided in my prior comments.

Issue #5251, opened 2024-10-10
Add theme category in settings page


LachTrip:
### Attempted Debugging

- [X] I have read the debugging page

### Searched GitHub Issues

- [X] I have searched GitHub for the issue.

### Describe the Scenario

Trying to make an additional settings page for my bookstack theme. I can add a new category to settings.layout and have a custom settings page, but to be able to access it I need to add an entry to the settingCategories in app/Settings/SettingController.php, which does not seem possible through current theming options.

### Exact BookStack Version

v24.03-dev 

### Hosting Environment

Debian docker

ssddanbrown:
Hi @LachTrip,
Yeah, the categories are not made customizable in any way.

I guess we could change this up so, instead of validating the categories by fixed array, we could validate them by checking for an existing matching view file. Would probably be best to move the existing setting category layouts into their own view folder though to prevent any pollution in that respect.

ssddanbrown:
Now achieved via your PR #5255, thanks again @LachTrip!

PR #5255, opened 2024-10-11
Validate settings categories by checking for existing views

LachTrip:
Removes fixed categories array and validates by checking the relevant view exists. Allows for custom settings categories to be added via theme system. #5251


ssddanbrown:
Thanks @LachTrip! Now merged to be part of the next release!

PR #5257, opened 2024-10-11
Export: convert tables from HTML to Markdown

WilliamDEdwards:
When exporting pages to Markdown, tables were still HTML. This is because `league/html-to-markdown` does not convert tables to Markdown by default, "because it is not part of the original Markdown syntax" (source: https://packagist.org/packages/league/html-to-markdown)


ssddanbrown:
Thanks for offering this @WilliamDEdwards.
While these days I'd prefer to stick to commonmark, and I'm hesitant of varied MD table syntax/support, I was thinking this may be desired & sensible since we handle MD tables in input.

That said, the handling provided by the conversion library is (understandably) simplistic and does not handle a range of potentially common scenarios like merged cells. I'd rather leave this as HTML than attempt a conversion that potentially breaks the structure.

I did look at performing a simple scan on the table for merges and then conditionally apply the conversion (for simple tables) but looks like that may be tricky via the current library structure, and I don't think the overall effort would be worthwhile to add/support/maintain for the relatively low demand.

Therefore I'm going to close this off. Feel free to open a feature request for this though to record demand and/or for further discussion around this.

PR #5288, opened 2024-10-24
Add firstName and lastName variables for LDAP

### LINKED ISSUE #1684: LDAP_DISPLAY_NAME_ATTRIBUTE merge multiple values
Ability to merge 2 values in `LDAP_DISPLAY_NAME_ATTRIBUTE` like:
`LDAP_DISPLAY_NAME_ATTRIBUTE={givenName sn}`

Merging first and last name from LDAP when fhere is no full name filled in LDAP

MatthieuLeboeuf:
Hello !

To resolve #1684 

I added two fields, LDAP_FIRST_NAME_ATTRIBUTE and LDAP_LAST_NAME_ATTRIBUTE, which, if both are filled, allow replacing LDAP_DISPLAY_NAME_ATTRIBUTE.

I did not set default values to avoid breaking existing installations.

### DIFF
--- a/app/Access/LdapService.php
+++ b/app/Access/LdapService.php
@@ -82,10 +82,12 @@ class LdapService
         $idAttr = $this->config['id_attribute'];
         $emailAttr = $this->config['email_attribute'];
         $displayNameAttr = $this->config['display_name_attribute'];
+        $firstNameAttr = $this->config['first_name_attribute'];
+        $lastNameAttr = $this->config['last_name_attribute'];
         $thumbnailAttr = $this->config['thumbnail_attribute'];
 
         $user = $this->getUserWithAttributes($userName, array_filter([
-            'cn', 'dn', $idAttr, $emailAttr, $displayNameAttr, $thumbnailAttr,
+            'cn', 'dn', $idAttr, $emailAttr, $displayNameAttr, $firstNameAttr, $lastNameAttr, $thumbnailAttr
         ]));
 
         if (is_null($user)) {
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Hi @MatthieuLeboeuf,
Thanks for offering this.

I'm not too keen on expanding our support of LDAP options, based on such little demand, but we do support something similar for SAML2/OIDC so it maybe makes sense to align functionality.

Rather than supporting three variables just for the name, could this instead follow the convention of our SAML2/OIDC implementation and allow the attributes to be defined under the single option, split via pipe, for example:

```bash
LDAP_DISPLAY_NAME_ATTRIBUTE=firstname|lastname
```

With found attributed values joined using a space. That way it's aligned with our other auth options, and allows any number of attribute values to be joined.

Also, if possible, would be good to have a test to cover any added/changed functionality here.
- [Existing ldap display name test as example](https://github.com/BookStackApp/BookStack/blob/e4ca3bf1329ab674ac7ca0220da5bf7befdc49b3/tests/Auth/LdapTest.php#L581C21-L604)
- [Testing docs](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md)

MatthieuLeboeuf:
Hello :)

For SAML2, SAML2_DISPLAY_NAME_ATTRIBUTES is used, and for OIDC, OIDC_DISPLAY_NAME_CLAIMS is used.

Should we rename LDAP_DISPLAY_NAME_ATTRIBUTE to LDAP_DISPLAY_NAME_ATTRIBUTES ?
However, this would introduce a breaking change.

ssddanbrown:
@MatthieuLeboeuf No, just keep it setting name as-is, we'll just live with the misaligned pluralisation.
I checked via a [quick search](https://ff1959.wordpress.com/2012/03/04/characters-that-are-permitted-in-attribute-names-descriptors/) that `|` should not be part of an existing value LDAP attribute name, so we should be able to add this to the existing option without breaking any instances.

MatthieuLeboeuf:
Okay :+1:

PR #5293, opened 2024-10-28
simpifly Dockerfiles

johnroyer:
some updates:
- Remove deprecated syntax
- use `mlocati/docker-php-extension-installer` to install PHP extensions: easier to read and install dependencies automatically
- use multiple docker images: docker image layers are re-usable, it can accelerate build speed while working on Dockerfile

Notes: everything should be works on `docker compose up`

### DIFF
--- a/dev/docker/Dockerfile
+++ b/dev/docker/Dockerfile
@@ -1,34 +1,34 @@
 FROM php:8.3-apache
 
-ENV APACHE_DOCUMENT_ROOT /app/public
-WORKDIR /app
-
-RUN <<EOR
 # Install additional dependencies
-apt-get update
-apt-get install -y \
-	git \
-	zip \
-	unzip \
-	libpng-dev \
-	libldap2-dev \
-	libzip-dev \
-	wait-for-it
-rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this @johnroyer but I personally prefer the original means of installing extensions and composer to be honest, rather than getting another container and external script involved.

johnroyer:
How do you think if I revert the commit of php extension installer ?

ssddanbrown:
Assuming that reverts both composer and extension install changes, then yeah should be good.
Just an FYI, I'll probably squash merge this rather than having (potentially) 7 commits for altering a single file.

johnroyer:
> Assuming that reverts both composer and extension install changes

OK, i will revert both

I will rebase onto `develop` before push.

johnroyer:
I merge changes to 2 commits. Seems everything is fine on build.

ssddanbrown:
Thanks @johnroyer, now merged.
While testing, I followed this up with 227c5e155b60ba9a442d23c64a9846366c3e15ac to address some other issues (unrelated to this PR).

johnroyer:
Thanks for the update

PR #5295, opened 2024-10-28
Added support for concatenating multiple LDAP attributes in displayName

### LINKED ISSUE #1684: LDAP_DISPLAY_NAME_ATTRIBUTE merge multiple values

MatthieuLeboeuf:
Hello

To resolve #1684 and the next of #5288 

I've made the requested changes, but I can't seem to fix the test error. Do you have any ideas ?

### DIFF
--- a/app/Access/LdapService.php
+++ b/app/Access/LdapService.php
@@ -72,6 +72,28 @@ class LdapService
     }
 
     /**
+     * Calculate the display name.
+     */
+    protected function getUserDisplayName(array $displayNameAttr, array $userDetails, string $defaultValue): string
+    {
+        $displayName = [];
+        foreach ($displayNameAttr as $dnAttr) {
+            $dnComponent = $this->getUserResponseProperty($userDetails, $dnAttr, null);
+            if ($dnComponent !== null) {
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Thanks again for providing this @MatthieuLeboeuf!
I followed this up with 90341e0e00a832e78df8a394487a073a475003ed just to mainly add test coverage for this feature.

Not certain what test was originally failing for you, but it might have been due to the ordering being changed via the merge of attributes when pulling back user data, which would have meant (after filtering away of a null thumbnail attribute) that the display name attribution in the test would be at a different index to what's expected.
In my commit I've altered this to be array expansion at the original array location instead.

### Docs Updates

- Document ability to use mulitple attributes via `LDAP_DISPLAY_NAME_ATTRIBUTE="firstname|lastname"` syntax.

PR #5302, opened 2024-11-01
Add support for STORAGE_S3_TOKEN

### LINKED ISSUE #5273: Add support for STORAGE_S3_TOKEN

Add support for defining **STORAGE_S3_TOKEN** in addition to **STORAGE_S3_KEY** and **STORAGE_S3_SECRET** for S3 storage.

Temporay S3 credentials generated by **STS** ([Security Token Service](https://docs.aws.amazon.com/STS/latest/APIReference/welcome.html)) need to define **credentials['token']** in addition to **credentials['key']** and **credentials['secret']**.

See AWS SDK docs about this: https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_temporary.html#providing-temporary-credentials-to-the-sdk-php

### Describe the benefits this would bring to existing BookStack users

This permit to use **temporary** and **limited-privilege** S3 credentials which increase security and permit some automation process.

### Can the goal of this request already be achieved via other means?

No.

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

1 to 5 years

### Additional context

I only use temporary and auto-generated credentials everywhere i can as a security principle.

vlaborie:
Hello,

To resolve #5273 and support temporary S3 credentials generated by STS.

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -131,6 +131,7 @@ STORAGE_ATTACHMENT_TYPE=local_secure
 # Amazon S3 storage configuration
 STORAGE_S3_KEY=your-s3-key
 STORAGE_S3_SECRET=your-s3-secret
+STORAGE_S3_TOKEN=your-s3-token
 STORAGE_S3_BUCKET=s3-bucket-name
 STORAGE_S3_REGION=s3-bucket-region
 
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this, but I'm not keen on expanding our support for S3 specific options, at least without some significant proven need/demand, whereas this seems to be desired for a one-off scenario specific context.

If it helps, you can override our config files using our [logical theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md), so you could effectively apply the same config value to your instance at runtime using the follow `functions.php` theme file:

```php
<?php

use BookStack\Facades\Theme;

Theme::listen(ThemeEvents::APP_BOOT, function () {
    config()->set('filesystems.disks.s3.token', 'mytokenvaluehere');
});

```

PR #5324, opened 2024-11-15
update text color on delete buttons

johnroyer:
text color modified:
- delete user button
- delete MFA method buttons

Those actions can not be undone, better to add warning.

----

Is there CSS style can make button in red background and white text on buttons? Thanks


ssddanbrown:
Hi @johnroyer,

Thanks for the PR, but I'm not keen on making these buttons stand out like this. 
I agree with having clear warnings (and confirmations), but these actions already have their own additional confirmation/warning steps.

johnroyer:
I understand. Close the PR.

PR #5337, opened 2024-11-26
Improve Oidc userinfo response Content-Type matching

wesbiggs:
Updates the OIDC userinfo endpoint request to allow for a `Content-Type` response header with optional parameters, like `application/json; charset=utf-8`. This was causing an issue (failure to find a valid subject claim) when integrating with [node-oidc-provider](https://github.com/panva/node-oidc-provider).


ssddanbrown:
Thanks for this @wesbiggs, this might explain some previously unexplained scenarios that have been reported before!
I followed this up with bc1f1d92e52ef5b103bdcbd739a001aece605b5d just with a little extra value normalisation, and to add a test to cover. 
Now merged to be part of the next patch release.

wesbiggs:
Great, thanks for taking a look!

PR #5339, opened 2024-11-27
Ability to prefix an S3 bucket

Topazjames:
We are running two bookstack instances. One for production and one for development.
This allows the ability to use a single s3 bucket but prefix where files are stored.
e.g. dev could use my-s3-bucket/dev and production could use my-s3-bucket/production


ssddanbrown:
Thanks for offering @Topazjames, But I'm not keen on adding, testing & supporting further options in this area without wider & established proven need. Feel free to open a feature request if you feel there is a wider need for this.

If this for some reason is essential in your environment, you could [follow the example here](https://github.com/BookStackApp/BookStack/pull/5302#issuecomment-2455025745) to make an unofficial unsupported hack to set this setting just in your instance.

PR #5350, opened 2024-12-03
fix x_shelves translation pt_BR

WillxRv:
fix x_shelves translation pt_BR


ssddanbrown:
Thanks for offering this @WillxRv but translation changes should really be made via the crowdin project instead: https://crowdin.com/project/bookstack

If you don't want to sign-up there, as a one-off I can copy this change to crowdin if you'd like.

ssddanbrown:
I've gone ahead and updated this is Crowdin. Used to be correct before anyway but was changed recently.
Change should be part of the next feature release.

PR #5357, opened 2024-12-11
Add link to missing setup.md from Codeberg

sedefkazan:
The BookStackApp installation guide is not explanatory for new users. There are missing parts in the .env file  database connection for BookStackApp. Laravel and PHP installations are not for beginners. The installation steps could be more explanatory and detailed for beginners. It is necessary to make sure that some folders are writable, but the more information section does not work. This statement means that users expect more detailed instructions in this section. How to make the storage and bookstrep/cache folders writable is missing.   
It is available on Codeberg at the following location:

[Installation.md on Codeberg] ( https://codeberg.org/sedefkazan/website/src/commit/73856595cc081fdc3a1180f2c8675668ef3b5173/content/docs/admin/installation.md )

Change in read.md for content/docs/admin/installation.md path in Codeberg.


ssddanbrown:
Thanks for offering this, but our installation guidance is purposefully kept platform abstract and minimal.
Exact steps where needed are left to our install scripts, community scripts and video guides.

ssddanbrown:
I'm going to go ahead and close this off as per my comments above, and since this is relevant to the [website repo](https://codeberg.org/bookstack/website) rather than the core app repo.

Issue #5287, opened 2024-10-24
RFC2307 (openLDAP/Posix) style LDAP group memberships


bennet0496:
### Describe the feature you'd like

There are two major ways groups can be resolved in LDAP Directory Servers. Either via the an attribute on the user (usually `memberOf`, RFC2307bis). Or via the group object just listing its members (usually with a `memberUid`), with a second query required to to find a user's groups (RFC2307). Bookstack currently only support the former to sync user groups. I'd like Bookstack to support both modes of group membership to also support Posix conform directory servers.

### Describe the benefits this would bring to existing BookStack users

OpenLDAP servers (or maybe other Non-AD Servers as well), predominately used in Unix/Linux environments, by default use the RFC2307 mode of managing group memberships, requiring extra setup to use the other via the `memberOf`-overlay. While it is certainly not uncommon to the have this overlay configured in the openLDAP world. Both modes are mutually exclusive, and can not trivially migrated from one to the other. This leaves users that in the past setup their OpenLDAP server, the default posix way, either without group sync capabilities in Bookstack or with the burden of migrating their LDAP directory and connected services to use RFC2307bis instead.

### Can the goal of this request already be achieved via other means?

No

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

1 to 5 years

### Additional context

I provided an exemplary implementation in #5281. However, it is currently unclear how large the user base of this  feature would be. And as @ssddanbrown pointed out, it might not be worth adding this additional complexity for only a handful of users. Therefore this feature request is primarily meant to gauge interest of the broader user base for this feature.

@ssddanbrown also noted that it might be an option to add custom group sync capabilities via logical themes. However, while this might be a reasonable middle ground, I'm not sure if there are that may other (useful) ways to sync groups (even with other login methods) and whether the complexity is better spend adding the feature directly or indirectly by letting user hook into the sync process.

bennet0496:
After testing around a bit, the following logical theme seams to to the trick
```php
Theme::listen(ThemeEvents::AUTH_LOGIN, function (string $authSystem, User $user) {
    $groupSyncService = App::get(GroupSyncService::class);
    $ldapService = App::get(LdapService::class);

    $getUserGroups = Closure::bind(function (string $userName) {
        $memberAttr = "memberUid";
        $groupBase = "ou=groups,ou=members,dc=example,dc=com";

        $ldapConnection = $this->getConnection();
[...]

PR #5387, opened 2025-01-03
Add Dual-Method Group Membership Verification for LDAP Integration

ghost:
This PR introduces an enhancement to BookStack’s LDAP integration by adding support for a dual-method verification system to handle group membership checks. Currently, BookStack relies solely on the "memberOf" attribute to determine group memberships, which works well for LDAP systems like Microsoft Active Directory but poses challenges for systems like OpenLDAP that do not support "memberOf" by default.

### DIFF
--- a/app/Access/LdapService.php
+++ b/app/Access/LdapService.php
@@ -357,6 +357,14 @@ class LdapService
         return $formattedGroups;
     }
 
+    public function isUserInGroup(string $userName, string $groupName): bool
+{
+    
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering but think something may be missing since this is just adding an unused function?

Either way, I'm only looking to expand options in this area with proven need, so I'm going to close this off, but feel free to add comment to #5287 if that specifically covers your need, or open a new issue if you have a specific different need to gain support input from others.

PR #5402, opened 2025-01-10
Allow api search to filter by book_id and chapter_id

helderjsm:
In Api we should be able to search by book_id and or chapter_id to reduce the set of results.

Example search using book_id filtering:
http://localhost:6875/api/search?count=100&page=1&query=p{type:page}{book_id:5}
Example search using chapter_id filtering:
http://localhost:6875/api/search?count=100&page=1&query=p{type:page}{chapter_id:2}

### DIFF
--- a/app/Search/SearchRunner.php
+++ b/app/Search/SearchRunner.php
@@ -454,6 +454,20 @@ class SearchRunner
         }
     }
 
+    protected function filterBookId(EloquentBuilder $query, Entity $model, string $input, bool $negated)
+    {
+        if ($model instanceof Page || $model instanceof Chapter) {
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this PR @helderjsm.
Happy for such a feature to be implemented since it meets my main criteria for API additions (mirrors what we provide in the UI, and we provide chapter/book specific search).

I'm not too sure about adding them via filters though.
I think the use is fundamental enough for them to be considered as their own API endpoints.
So `/api/search/book/{id}`/`/api/search/chapter/{id}`.
This way the type filtering can be hard-coded/limited instead of causing confusion in the need to apply them.

ssddanbrown:
Since there's been no further follow-up here I'm going to go ahead and close this off.

PR #5418, opened 2025-01-16
Add `.svg` extension to supported image file extensions

tjone270:
I've noticed that SVG images are not in the hardcoded list of supported image extensions. Having tested SVGs on my BookStack instance in the WYSIWYG editor these behave normally, so think it's worthwhile having this as an out-of-the-box supported image.

### DIFF
--- a/app/Exports/ZipExports/Models/ZipExportImage.php
+++ b/app/Exports/ZipExports/Models/ZipExportImage.php
@@ -32,7 +32,7 @@ class ZipExportImage extends ZipExportModel
 
     public static function validate(ZipValidationHelper $context, array $data): array
     {
-        $acceptedImageTypes = ['image/png', 'image/jpeg', 'image/gif', 'image/webp'];
+        $acceptedImageTypes = ['image/png', 'image/jpeg', 'image/gif', 'image/webp', 'image/svg+xml'];
         $rules = [
             'id'    => ['nullable', 'int', $context->uniqueIdRule('image')],
             'name'  => ['required', 'string', 'min:1'],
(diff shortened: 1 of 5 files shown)


ssddanbrown:
Related to #1103

Thanks for offering this @tjone270.
This is a path I've thought about and tested before: 

- https://github.com/BookStackApp/BookStack/pull/3452
- https://github.com/BookStackApp/BookStack/issues/1170#issuecomment-1136182504

My issue comes down to SVG being a complex format that we can't assure wider support for (like in PDF exports) and which brings a wider range of expectations upon normal images (interactivity).
Edit: Also need to think the security implications of displaying SVG images, especially when accessed direct via browser in public storage. Quite likely a range of security concerns.

Maybe the demand/value is enough for implementation regardless of compromise.
Think I need to dwell on this, I might ask about this via the issue for feedback.

thajohns:
With regards to security, I think [this chapter from the SVG standard](https://www.w3.org/TR/SVG2/conform.html#referencing-modes) is useful.

morrolinux:
I deperately need this feature for migrating to bookstack. 
How about making it an opt-in until all the rest is figured out @ssddanbrown ?

ssddanbrown:
Thanks again @tjone270 for offering this, but I'm going to close this off as SVG support makes me a little too uneasy right now, when it comes to security and feature support, especially relative to demand (since would only cover a portion of the SVG demand, as some existing requests are for interactivity).

It may be something to return to in the future, but it probably won't be anytime real soon so will close this pending PR.

PR #5429, opened 2025-01-20
Add optional OIDC avatar fetching from the ```picture``` claim

### LINKED ISSUE #4271: OIDC: Support `picture` claim for use as user avatar

rubentalstra:
This update enables BookStack to **optionally** fetch user avatars from the OIDC `picture` claim. The implementation:

- closes: https://github.com/BookStackApp/BookStack/issues/4271

1. Introduces a `fetch_avatars` config flag in `config/oidc.php` to toggle avatar retrieval.  
2. Uses `UserAvatars->assignToUserFromUrl($user, $picture, $accessToken)` to support **both** public and **private** (Bearer token-protected) endpoints.  
3. Fetches the `picture` claim from the user’s ID token or userinfo response, if provided.  
4. Works with various OIDC providers (Google, Okta, Keycloak, Azure, etc.), provided they include or allow retrieving a valid `picture` URL.  
5. Includes SSRF protection notes, advising users to only enable this if they trust the domains in the `picture` field.

This approach does **not** break existing behavior; avatar fetching is off by default. If enabled, BookStack will try to update a user’s avatar upon login, using the token to authenticate if necessary.

### DIFF
--- a/app/Access/Oidc/OidcService.php
+++ b/app/Access/Oidc/OidcService.php
@@ -11,6 +11,7 @@ use BookStack\Exceptions\UserRegistrationException;
 use BookStack\Facades\Theme;
 use BookStack\Http\HttpRequestService;
 use BookStack\Theming\ThemeEvents;
+use BookStack\Uploads\UserAvatars;
 use BookStack\Users\Models\User;
 use Illuminate\Support\Facades\Cache;
 use League\OAuth2\Client\OptionProvider\HttpBasicAuthOptionProvider;
@@ -26,7 +27,8 @@ class OidcService
         protected RegistrationService $registrationService,
         protected LoginService $loginService,
         protected HttpRequestService $http,
-        protected GroupSyncService $groupService
+        protected GroupSyncService $groupService,
+        protected UserAvatars $userAvatars
     ) {
     }
 
@@ -226,6 +228,10 @@ class OidcService
         }
 
         $this->loginService->login($user, 'oidc');
+
+        if ($this->config()['fetch_avatars'] && $userDetails->picture) {
+            $this->userAvatars->assignToUserFromUrl($user, $userDetails->picture, $accessToken->getToken());
+        }
 
         return $user;
     }
--- a/app/Access/Oidc/OidcUserDetails.php
+++ b/app/Access/Oidc/OidcUserDetails.php
@@ -11,6 +11,7 @@ class OidcUserDetails
         public ?string $email = null,
         public ?string $name = null,
         public ?array $groups = null,
+        public ?string $picture = null,
     ) {
     }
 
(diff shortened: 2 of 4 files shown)


ssddanbrown:
Thanks for providing this @rubentalstra.

Is the authorization header in the picture URL request based upon any actual need or standard/specification?
I'd be wary of sending the access token out, especially to other domains, but would prefer not to send at all if there's no need.

Also, I wouldn't look to have this run on each login, just the first registration/sync as per all existing details in support SSO options. We'll also need to add testing to cover the added functionality. I'm happy to make these changes/additions before merge, but I really just need to know about the authorization point above.

rubentalstra:
@ssddanbrown hi, thank you for taking the time. The header is for sure needed if you use Microsoft because it will call a global endpoint. And based on the header it will return the users profile.

ssddanbrown:
Thanks for this @rubentalstra.

I can confirm via testing.
Really not sure I want to go outside the spec just to suit Microsoft though.
Also, the origin between picture url and issuer is different so we can't put this behind a basic origin check.
Could have extra config options but that's getting dumb for a minor side features of an optional auth option.

I'm siding towards not sending any auth tokens, leaving any non-spec services like Entra/AzureAD use our logical theme hooks to work around their awkardness.

Another consideration for this PR: it's currently saving images as png, which probably will be the most common, but this is not assured; The spec does not confirm exact formats.

ssddanbrown:
This has now been merged for the next release, thanks again @rubentalstra.
I made some changes as could be seen via #5626. These included the removal of sending auth tokens as per my last comment.

rubentalstra:
> This has now been merged for the next release, thanks again @rubentalstra.
> 
> I made some changes as could be seen via #5626. These included the removal of sending auth tokens as per my last comment.

@ssddanbrown thank you for reviewing my PR. I'm happy to see that it was not for nothing 😉

rubentalstra:
- close: https://github.com/BookStackApp/BookStack/issues/4271

Issue #395, opened 2017-05-12
Feature: Collaborative editing (at least conflict/locking)


bios-ben:
Desired Feature:
Currently if you have multiple people editing the same document, the most recent person to save just overwrites the other. At the very least this should use lock files and note when there's conflicts or try to auto merge. But the better feature would be to collaborative edit using sockets like other editing tools such as etherpad or hackmd.

Shackelford-Arden:
While I +1 this, I'd also say that it would be a nice addition to have the option between Check In/Out method and collaborative writing. Some may not want the collaborative ability. Personally I would, but I know my last job would simply want a check in/out functionality.

ssddanbrown:
Yeah, I can see the benefit of something like this. Collaboration will be hard to do technically, Especially on the current laravel codebase without the app structure getting extra complicated. My concerns with locking is that it kind of relies on the user to lock/unlock which we can the work around with notifications and such but it does add a lot to the system.

At the moment the system will notify upon edit if someone else has started editing but there's nothing forceful in place.

therealscottcarlow:
Locking or check-in/check-out is going to be necessary for my org at scale. I would love to see this implemented. Collab is OK, but realistically if I wanted to collab on a doc I wouldn't use Bookstack as the platform to do that in. That's my 2 cents.

bios-ben:
What I'd like to see in an app, if I were to make something ideal from scratch, is to be able to edit a file and choose collaborative, real-time read-only, or private lock. In my org the main thing I want a lot of the time is to have multiple people live viewing the edits as they go in (in final formatted form), while one person is doing the editing using whatever editor (markup, wysiwyg, etc). This allows people to essentially audit the documenting party, while only one person is responsible for actual dictation. I haven't seen software like this, and if I had the cycles/funding I'd build it. 

----- Original Message -----

dsaundersWizmo:
+1 for Check In/Check Out.

Would also need admin level force Check In and probably a way to do that in bulk.  Check Out expiration would be nice.

vdavid:
Re: @ssddanbrown

> At the moment the system will notify upon edit if someone else has started editing but there's nothing forceful in place.

Actually, that'd be cool but it doesn't seem to do that.
[...]

PR #5451, opened 2025-01-28
Try and handle conflict on update: it's a plan

bendem:
This is a plan to fix the data loss part of #395. I know absolutely nothing of the codebase and my years of php are far. I already have a functional PoC, but that's not the best user experience and I'm afraid getting it polished is beyond what I'm able to do for now.

Video of it in action: 

https://github.com/user-attachments/assets/93d005e0-a72f-449f-bb8c-9fc04de2023e

### DIFF
--- a/app/Entities/Controllers/PageController.php
+++ b/app/Entities/Controllers/PageController.php
@@ -222,6 +222,23 @@ class PageController extends Controller
         $page = $this->queries->findVisibleBySlugsOrFail($bookSlug, $pageSlug);
         $this->checkOwnablePermission('page-update', $page);
 
+        $editorRevision = $request->get("revision");
+        // TODO This query is repeated at app/Entities/Tools/PageEditorData.php:73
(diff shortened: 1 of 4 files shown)


ssddanbrown:
Thanks for the suggestion and PR @bendem.

I'm not opposed to introducing a higher barrier to saving overlapping changes, but I wouldn't want to outright prevent it like in this implementation, I don't want to be leading users into hard blocks/dead-ends.

We could do something similar to this but:
- On save compare created dates between actual page latest revision and user draft (if existing).
- If user draft less created before page latest revision, 
  - Save submitted content to user draft.
  - Return to editor with warning about potential conflicts.
  - Require secondary confirmation (save again after being shown warning) before actual save.

bendem:
Agreed, this was mostly to illustrate [my comment](https://github.com/BookStackApp/BookStack/issues/395#issuecomment-2188330452). I'm not in a position to push this further, but I hope it can spark a contribution from someone more familiar with php and the codebase.

Issue #1271, opened 2019-02-11
Only allow a certain social auth instead of traditional email login


aditya-aggarwal-1:
If using in a business setting over the public internet, it would be beneficial to only allow a certain Social Login option, and not show the default Email/password login.

**Describe the benefits this feature would bring to BookStack users**
Use-case examples would include the following use case:
I work in a start-up where we use Bookstack for our documentation purposes to share data between business as well as engineering teams. Since we do not use an LDAP, and mostly rely on Google Auth, it would be one less effort for my ops-team to ensure that the user doesn't have an account left Active on Bookstack after he/she leaves the organization. As we deactive their Google Login, it would deactive their access altogether. This is somewhat difficult to control in the current system, since we do not know if a user registered using Google or basic sign up.

Simply being able to set auth method to "google" or "socialite" may help.

ssddanbrown:
Thanks for the suggestion @aditya-aggarwal-1.

This seems very similar to #1090. If there are differences to your request please let me know otherwise I'll soon close this issue as a duplicate.

the-cybersapien:
@ssddanbrown yes, it's the same as #1090. Weird that when I searched the issues to find something like this, I didn't get this issue in the list. Wrong keywords I guess.
Please feel free to close this as a duplicate.

ssddanbrown:
Thanks for confirming!

PR #5461, opened 2025-01-31
Enable Social Login with Planning Center Online

monoxane:
Hey, I'm probably being a bit of a pest opening a PR without an associated issue so sorry for that.

This PR adds support for Social auth with [Planning Center](https://www.planningcenter.com), a scheduling and management app typically used with Churches, but also somewhat popular with some types of non-profits.

The rationale behind enabling social auth using the existing Socialite dependency is that PCO has their own OAuth2.0 implementation that does not conform 100% to the OIDC spec, and therefore cannot be used with the generic OIDC provider in BookStack (See [planningcenter/developers #1271](https://github.com/planningcenter/developers/issues/1271) for me complaining about it)

I don't really have a full PHP environment set up as I usually do golang/react but I've tested this locally using the docker-compose stack as well as building the LinuxServer.io image using it, and deploying that to my prod env. Changes behave as expected and I am able to log in using a PCO account via Oauth, and the account is successfully registered and marked as being from PCO. I've ran static check and the linter and they both passed.


ssddanbrown:
Thanks for offering this PR @monoxane, but I'm not longer accepting/adding new social auth platforms since they are problematic to fully support/maintain.

We do however provide a way to add custom additional social auth options via [our logical theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md#logical-theme-system).
There's specific guidance for this towards the bottom of the logical theme system docs: https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md#custom-socialite-service-example

I'm therefore going to close this PR off but feel to continue discussion here or via another issue or via Discord etc... if you need further help/guidance on this.

Issue #5468, opened 2025-02-03
MFA Option: Email based OTP


emirhanuysall:
### Describe the feature you'd like

Currently, BookStack does not support OTP-based authentication for email logins.

We propose adding an optional OTP feature that integrates with existing MFA systems, allowing users to enter a one-time code sent via email during login. 

This would enhance security while maintaining compatibility with BookStack's existing authentication flow.

### Describe the benefits this would bring to existing BookStack users

Enhanced Security & Flexibility

-Improved Account Protection: Adds an extra layer of security by requiring a one-time password (OTP) during email login, reducing the risk of unauthorized access.

-Seamless MFA Integration: Allows users to enable OTP authentication without modifying BookStack’s core authentication system, making it easier to integrate with existing MFA setups.

-Better Compliance: Helps organizations meet security and compliance requirements by providing an additional authentication step.

-User-Friendly Authentication: Ensures a simple and effective way to enhance security without requiring external authentication providers.

### Can the goal of this request already be achieved via other means?

No.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

1 to 5 years

aliozgur:
@ssddanbrown, @emirhanuysall is on my team, and this feature was developed based on a requirement.

To give you more context: we use BookStack for customer-specific documentation, and we have hundreds of customers and users who log in to BookStack with their corporate email addresses. Customer documents usually contain private details specific to the customer. When a BookStack user associated with a customer leaves, we want to cut off their access to our BookStack instance.

In practice, knowing which user has left a customer is not always possible, as our customers do not notify us in such situations. As a result, we thought the Email MFA option would be a valuable feature, as users who no longer have access to their corporate accounts would be unable to log in to BookStack.

We would appreciate it if this PR were merged.

ssddanbrown:
@aliozgur Thanks for offering development of the PR but as said in #5469 I'd want to see some proven demand/need. Need for a single scenario/business and provision of a PR are not really core reasons I'd consider for inclusion and growing the project scope.

> As a result, we thought the Email MFA option would be a valuable feature, as users who no longer have access to their corporate accounts would be unable to log in to BookStack.

[...]

PR #5469, opened 2025-02-03
Adding OTP Support for Email Login #5468

emirhanuysall:
Currently, BookStack does not support OTP-based authentication for email logins.

We propose adding an optional OTP feature that integrates with existing MFA systems, allowing users to enter a one-time code sent via email during login.

This would enhance security while maintaining compatibility with BookStack's existing authentication flow.

### DIFF
--- a/app/Access/Controllers/EmailController.php
+++ b/app/Access/Controllers/EmailController.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace BookStack\Access\Controllers;
+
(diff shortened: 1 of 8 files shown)


ssddanbrown:
Thanks for offering this PR @emirhanuysall but I'd want to see some proven desire/need before adding options.
Therefore I'm going to close this, but #5468 will remain open to receive feedback.

PR #5473, opened 2025-02-06
bug fix for Search Does Not Find Words Inside Guillemets quotes (« »)

inv-hareesh:


ssddanbrown:
Thanks for opening this PR @inv-hareesh but this has a mix of changes. That combined with having no PR description makes me think this may be an accidental or bot/ai PR, and I don't look to merge a mix of changes via a PR, so I'm going to therefore close this off.

PR #5475, opened 2025-02-07
Fix search issue for words inside Guillemets (« ») without spaces

inv-hareesh:
This commit fixes an issue where words enclosed in Guillemets (« ») were not indexed correctly for search when there was no space after the opening quote. The search engine was not recognizing such words as separate terms, leading to failed search results.

The fix involves updating the list of delimiter characters in SearchIndex::$delimiters to include Guillemets (« and »), ensuring that words within them are properly tokenized and indexed.

### DIFF
--- a/app/Search/SearchIndex.php
+++ b/app/Search/SearchIndex.php
@@ -16,7 +16,7 @@ class SearchIndex
     /**
      * A list of delimiter characters used to break-up parsed content into terms for indexing.
      */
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks @inv-hareesh, now merged for next feature release.
I followed this up with some tests to cover via f1b8e857bf68eb83ecc2db40da53f5d4626a16de

Issue #5458, opened 2025-01-31
Users can edit and rename shelves


Salla205:
### Describe the Bug

The Asset Permissions set under Settings → Roles do not automatically apply to added shelves. A shelf will only become visible to users and its functions will be activated once one of the options "View," "Create," "Update," or "Delete" is selected. The specific permissions assigned to a shelf override the general role settings and can render them ineffective.

Here are a few images for better understanding.

![Image](https://github.com/user-attachments/assets/bc8f35e1-e6e5-4b58-9abc-f4d0fb05bc13)

![Image](https://github.com/user-attachments/assets/678749cf-efd4-481c-adf8-efe29192d4c9)

![Image](https://github.com/user-attachments/assets/1d93e9ba-b6d5-4c7f-a0bc-e1a36bc75405)

Background on Company Usage
We plan to introduce BookStack company-wide and provide each department with its own shelf. Additionally, the IT department will provide shelves containing central documentation.

Department Shelves: Visible only to the respective department and admins – not to other departments.
Permissions in Department Shelves: Users can create books but cannot edit the shelf itself. Within their own department, users are allowed to delete books, chapters, and pages.
IT Department Shelves: Documentation with view-only permissions must not be copied.

Additionally, only shelves should be displayed, and users should not be able to create books via "Book", as these are only visible to themselves.

![Image](https://github.com/user-attachments/assets/65730b23-bf71-4904-bc5b-67f01c3177b2)

### Steps to Reproduce

Go to Settings → Roles and create a new role.
Assign Asset Permissions as shown in the image above. No System Permissions are selected.

Now, add this role to a shelf under Shelves → Add Role, but do not check "View," "Create," "Update," or "Delete." The shelf will not be visible to the user.

- The shelf only becomes visible when "View" is selected.
- The "Create" option has no effect.
- "Update" allows users to create books but also edit shelves—which should not be possible.
- "Delete" enables users to delete shelves, even though this permission is not explicitly selected.

### Expected Behaviour

Either the "Create" permission must be enabled to allow book creation, or the specific shelf permissions should not override the Asset Permissions.

Additionally, there should be a button to hide Books in the top navigation bar and display only Shelves, since users should not create their own books. This option should only be available to Admins or IT users.

### Exact BookStack Version

BookStack v24.10

ssddanbrown:
Hi @Salla205,
I'm not really sure I see a bug here.

Quoting the reproduction steps:

[...]

PR #5514, opened 2025-03-03
bug fix for Users can edit and rename shelves #5458

inv-hareesh:
Fixed an issue where all users could edit and rename shelves by adding a check to ensure only admins have permission to perform these actions.


ssddanbrown:
Thanks for offering @inv-hareesh but there's not really an issue to be solved here, there are already relevant permission checks on theses actions, therefore I'm going to close this off.

PR #5549, opened 2025-03-23
Update _text.scss to add a bottom border / underline under titles

seabee33:
Adds a grey border line under the title on each page view
the screenshots below show an example
Screenshot from 2025-03-23 13-28-36
Screenshot from 2025-03-23 13-28-55


ssddanbrown:
Thanks for offering this @seabee33, but this seems to be a style preference rather than something to be changed in core, so I'm going to close this off. 
If you wanted this, you should be able to add this as a customization via the "Custom HTML Head Content" customization setting.

Issue #5542, opened 2025-03-18
Sorting by name not working with "É"


serinf-lauza:
### Describe the Bug

Hi! I created a sorting rule that only has Name - Alphabetical (Asc) in it. When I apply it to a book, the letter "É" gets sort out at the end of everything. It should be between "D" and "F", the previous method for sorting by name still works and put it as it's rightful place (Sort by name). My BookStack is in French, so we use a lot of "É" in our titles!
Thank you! :)

### Steps to Reproduce

1. Settings > Sorting
2. Create a rule with Name - Alphabetical (Asc)
3. Choose the Auto sort Option created previously to sort a chapter
4. See where "É" goes

![Image](https://github.com/user-attachments/assets/99cb8676-7417-43e4-adfa-7e73ac94af1e)

### Expected Behaviour

It should be between "D" and "F" when sorted by name. 

### Browser Details

Chrome

### Exact BookStack Version

v25.02.1

bernardo-campos:
In `BookStack\Sorting\SortSetOperationComparisons`, the `nameAsc` and `nameDesc` functions use:

`return strtolower($a->name) <=> strtolower($b->name);`

The `<=>` operator combined with `strtolower()` doesn’t reliably handle accented characters across all configurations. For example, it might sort "é" after "z" leading to inconsistent results in languages like Spanish or French.

Using `Collator` from the **intl extension** would be more appropriate, as it offers locale-aware comparisons for accurate and culturally correct sorting of accented characters.

I’m working on a PR to address this.

ssddanbrown:
Thanks to the work done by @bernardo-campos in #5550 this has now been improved, with the applied changes due to be part of the next patch release.

Thanks @serinf-lauza for raising.

PR #5550, opened 2025-03-23
Fix issue #5542 Sorting by name

bernardo-campos:
Related Issue: [Issue #5542](https://github.com/BookStackApp/BookStack/issues/5542)

**Summary:**

Replaces the `<=>` operator with PHP’s Collator class for locale-aware string comparisons in `App\Sorting\SortSetOperationComparisons`. Using `<=>` with `strtolower()` mishandles accented characters (e.g., "é" sorts after "z"), which is unreliable for languages with diacritics. Collator ensures accurate sorting based on the user’s locale.

**Changes:**

Added a new protected static method `getCollator()`:

Inside class `App\Sorting\SortSetOperationComparisons`, I've added
```
 protected static function getCollator(): ?\Collator
 {
    $locale = user()->getLocale()->isoLocale(); // e.g., 'eu_ES', 'fr_FR'
    return class_exists(\Collator::class) ? collator_create($locale) : null;
}
```
- `$locale` is fetched from the user’s settings via `BookStack\Translation\LocaleManager`.
- `class_exists(\Collator::class)` checks for the **intl extension**, returning `null` if unavailable (graceful fallback).
- The returned `Collator` object (or `null`) is used in the class’s comparison functions to handle sorting.

**Impact:**

- Improves sorting for users with locales like fr_FR or eu_ES.
- No breaking changes; users without intl fall back to existing behavior.

**TL;DR:** Swaps <=> for Collator to fix accented character sorting, with a fallback for missing intl.

### DIFF
--- a/app/Sorting/SortSetOperationComparisons.php
+++ b/app/Sorting/SortSetOperationComparisons.php
@@ -2,6 +2,7 @@
 
 namespace BookStack\Sorting;
 
+use voku\helper\ASCII;
 use BookStack\Entities\Models\Chapter;
 use BookStack\Entities\Models\Entity;
 
@@ -13,12 +14,12 @@ class SortSetOperationComparisons
 {
     public static function nameAsc(Entity $a, Entity $b): int
     {
-        return strtolower($a->name) <=> strtolower($b->name);
+        return strtolower(ASCII::to_transliterate($a->name)) <=>  strtolower(ASCII::to_transliterate($b->name));
     }
 
     public static function nameDesc(Entity $a, Entity $b): int
     {
-        return strtolower($b->name) <=> strtolower($a->name);
+        return strtolower(ASCII::to_transliterate($b->name)) <=>  strtolower(ASCII::to_transliterate($a->name));
     }
 
     public static function nameNumericAsc(Entity $a, Entity $b): int
--- a/tests/Sorting/SortRuleTest.php
+++ b/tests/Sorting/SortRuleTest.php
@@ -198,6 +198,8 @@ class SortRuleTest extends TestCase
         $namesToAdd = [
             "Beans",
             "bread",
+            "Éclaire",
+            "egg",
             "Milk",
             "pizza",
             "Tomato",


ssddanbrown:
Thanks for offering @bernardo-campos, but I'm not too keen on us having diverging logic for minor functionality depending on available extensions. I'd also be concerned about the use of locale, since the user locale isn't assured to match the content and I don't want sort logic to change depending on user performing the action which causes changes.

The codebase already uses the https://github.com/voku/portable-ascii library via Laravel, could we maybe use the `to_transliterate` function of that instead? Keeping the unknown chars as a fallback instead of replacing/removing.
That looks to use the intl extension logic under the hood, with fallbacks for environments without the extension which should hopefully cover most common cases.

Also, it would be good to have some added tests to cover this scenario, similar to our existing name sort tests: https://github.com/BookStackApp/BookStack/blob/13dae24cbe3f95ba6747f1ffe37bc3e1a4db4af9/tests/Sorting/SortRuleTest.php#L190

bernardo-campos:
Ok. I see the point. Yep, using `to_transliterate` it works like a charm!

bernardo-campos:
I’ve changed:
```return strtolower($a->name) <=> strtolower($b->name);```
to 
```return strtolower(ASCII::to_transliterate($a->name)) <=> strtolower(ASCII::to_transliterate($b->name));```

I know the logic got a bit lengthy, but I realized that just using `ASCII::to_transliterate` alone didn’t work because it preserves uppercase and lowercase letters, causing the comparison to sort them separately (uppercase first, then lowercase). Using `strtolower` inside doesn’t help either, since `strtolower('É')` still keeps it as **É**.

Another idea is to use laravel helper `Str::slug($somestring, ' ')` maybe it is simpler and slug function is already used in several parts of the project

ssddanbrown:
Thanks @bernardo-campos, although I think we'd want to specifically pass in `null` as the second parameter to the `to_transliterate` calls, since otherwise unknown non-ascii chars will be converted to `?`, whereas i think it's better to retain the original utf8 values so that sorting remains stable in such cases.

> Another idea is to use laravel helper Str::slug($somestring, ' ') maybe it is simpler and slug function is already used in several parts of the project

That would lead to a bit too much conversion of the original text, similar to what I mention above, which would lead to less stable/expected sorting outcomes.

ssddanbrown:
Thanks again @bernardo-campos,
Changes now all merged in for next patch release.

I followed this up with 1ba0d26fdd1e673dfcaefc87e2fa38f08cac61ad to address my comment above.

PR #5552, opened 2025-03-24
Vector Search & LLM Integration

ssddanbrown:
This is a proof of concept for some level of somewhat native LLM integration.
This manages vector indexing, searching and LLM querying, with vectors stored & queried in the local BookStack database, and an external LLM service (OpenAI, Ollama etc...) queried for embeddings & query execution using locally found vector-based results for context.

The result is a LLM generated response to user query, with a list of relevant BookStack content used for the query as reference.

### Issues & Questionables

- MySQL lacks proper support. MySQL 9 supports vector columns and indexes, but you can't query distance in the community versions. MariaDB is good from 11.7.
  - Not sure how I feel about MariaDB specific features/functionality. Current implementation uses MariaDB specific functions.
- No idea how non-en/mixed language content works for these systems yet.
- Do we set official targets of support? How common is OpenAI API compatibility across other providers?
- Does embedding size vary much across difference services/models?
  - Seems to: `all-minilm`=384, `openai:text-embedding-3-small`=1536,  `openai:text-embedding-3-large`=3072
  - Inserting 384 length embeddings into `vector(1536)` mariadb column seemed to result in 0 hex values. Seems like the column size has to be correct, can't insert under?
    - Response from Sergei at MariaDB on this: 
      - "all vectors in a column are strictly the same size, because the distance between two vectors of different sizes is not defined. if you store 384-dim vectors in vector(1536) you can pad them with zeros and the result will be exactly the same as storing them in vector(384) except that it'll use about 5x more storage and memory"
    - Will need to provide controls/mechanism to change this.
- How is those accessed/shown in the UI? Option in the search bar maybe to toggle search type? Action/reference from the normal search page?
- What *needs* to be configurable? Looks like a lot of things that could be tweaked, but don't want to be maintaining loads of specific options. Need to assess what's most important.

### Considerations

- This will really require a queue worker so indexing can be done in the background.
  - This does mean that search index may be behind content.
- Would need migration handling due to limited DB support, so maybe have the migration have widespread support by default but add vector parts via a command?
- For a nice UX, would probably need to have results queried via JS rather than results on page load. Not a big deal, but I generally try to have read/view operations be non-JS friendly. 
- Lots of context-window/vector-size considerations to be made that have not been implemented yet.
- We'd probably want to cap the context documents (at least those shown to a user) at a certain score to avoid showing completely irrelevant content.

### Todo

- [ ] Resolve above issues/considerations where needed.
- [ ] Need to ensure vectors handled on content delete/update.
- [ ] Resolve code TODOs
- [ ] Cover functionality with testing
- [ ] Show warning for LLM/AI-based output (inaccurate)
- [ ] Extract text from views
- [ ] Check against dark mode
- [ ] Check accessibility
- [ ] Add JS server-side-event lib to readme attribution list?

### Implementation Notes

- In some scenarios extra web-server config may be needed to prevent server-side buffering from messing with server-side-event handling. Not crucial, as won't break things, but can delay the response until finished on server, providing a worse experience.  For example, for nginx proxies: `proxy_buffering off; proxy_cache off;`.


ssddanbrown:
Preview:

https://github.com/user-attachments/assets/d710d8a9-3aa5-4203-a271-dd9fb65c5cc2

jrejaud:
ChatGPT just opened integrations, would be lit if I could use ChatGPT to search Bookstack :)

ItsNoted:
This is awesome Dan! Would be even more amazing to see integration with Ollama so we can self host everything locally and not have to worry about calling ChatGPT :)

ssddanbrown:
@ItsNoted local LLM support would be an important requirement really for an implementation since I know that matters a lot to our kind of audience, and when building I tested this in bulk using Ollama (since they are fairly OpenAI API compatible) and all seemed fairly functional.

ivano-buffa:
Hi @ssddanbrown,
Any update on this?

Little suggestion: perhaps CrewAI might help? Less mature that LangGraph but highly powerful.

The idea of indexing your own instance of Bookstack to privately "talk" to your own data and "retrieve" what you need, will definitely give to Bookstack the boost that it deserves.

My best.
~Ivano

ssddanbrown:
@trendpx I spent some time on this last month, but got to a point where I was not happy with the vector search system. It didn't feel it was working well enough to justify the addition cost in a way which would generally work well out of the box. It can work for sure, but with a lot of environment/model/context specific tweaking.

I want to revisit this with an alternative & simpler approach, without vectors, to test this kind of flow:

```txt
User enters query -> LLM consulted to extract key search terms -> Terms used for BookStack search -> Relevant pages and original query sent to LLM for response.
```

I'd prefer not to require any specific extra external platforms/libraries where possible.

samschultzponsys:
This looks incredible. I would love to embed an agent into a site if thats possible, I know that would be out of the scope of bookstack and is more up to openwebui etc.

HelloGithubUniverse:
are there any updates on this?

ssddanbrown:
@HelloGithubUniverse Since the start of the year I've been playing with some options, initially trying what I mentioned above back in September, but again wasn't really happy with the results.

I then started playing with a new approach, where this could be set-up as an extension to existing version of BookStack, using our theme systems, supported with an external API and LLM/vector database stack specifically suited for this kind of thing. Will mean it's not an official part of the core, and not supported in the same manner, but will be more flexible to tweaking and/or extension to suit specific environments and features.

I'm aiming to have it all as part of a docker-compose stack, so you can just bring it up, then add the required BookStack theme files, then go, at least initially. Just doing some latter design and testing bits on that now. Aiming to share it during this next coming week.

After then, I'm thinking I'll use it as a test-bed for taking our extension/theme system to the next step (allowing easier piecing together of hacks) for the next BookStack feature release.

CrazyWolf13:
That sounds awesome!

I hope not only docker but also bare metal installations will be officially supported?

HelloGithubUniverse:
Thats awesome! Really looking forward to it.

ssddanbrown:
> I hope not only docker but also bare metal installations will be officially supported?

@CrazyWolf13 The stack required for the extra database, API and LLM parts will be provided as a docker compose stack for convenience, although you could technically run any parts outside of docker. It should be usable with any up-to-date version of BookStack though, regardless of how its hosted (might be some network considerations depending on environment).

---

Almost there in regards to sharing the initial preview version. Just some extra docs and testing to do.
Here's a preview of how things are looking:

 
https://github.com/user-attachments/assets/799a900d-4e57-431f-9287-82d4e99bb666

CrazyWolf13:
Thanks, yeah I can understand that.

We'll try our best for the proxmox helper scripts to make it available though.

Btw the preview looks awesome!

ssddanbrown:
Here's an initial publish of the stack mentioned above:
https://codeberg.org/bookstack/query

Once it's had some further refinement, and after I've changed things to work with future theme system developments, I'll probably put together a video guide on setup and usage.

I'm going to close off this PR for now though since I'm getting more sure about going in that direction of an extension/hack instead of being part of the core.

nekromoff:
I am planning on integrating NotebookLM with Bookstack - probably via automatic export of MD files into drive and linking those into NotebookLM. I will report once I have the integration set up.

nekromoff:
I can confirm NotebookLM works fine with one large MD file as a source (exported all wiki as MD files and then just joined in under one MD).

I have described my solution here:
https://ambience.sk/use-bookstack-wiki-with-notebooklm/

PR #5570, opened 2025-04-07
Configurable log options

miloslavkostir:
In my Laravel projects I'm used to create log options configurable. Projects run in Docker container and the log dir is mapped out of the container.

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -54,6 +54,12 @@ APP_THEME=false
 # Can alternatively be set to "*" to trust all proxy addresses.
 APP_PROXIES=null
 
+# Log details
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this PR @miloslavkostir, but I'm not keen on adding extra env options without proven wider need.

Not sure why you can't map the existing directory/file out of a container already.
Otherwise, some other options are:
- Patch the option as part of your container build process.
- Use the [logical theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md#logical-theme-system) to override the option.

Issue #1265, opened 2019-02-07
Inline commenting


abbeyjackson:
This is a suggestion added to issues as per Dan’s instructions via Twitter. 

Inline commenting (selecting a chubk of text to comment on or having a comment associated with a specific line number) is a feature that is missing from Bookstack which I feel would really add value to the product.

JohnnyWild89:
Looking forward to get this feature! Bookshelf is perfect replacement for any products which use as WIKI. Inline commentaries would greatly expands work possibilities.

cze-karel:
This would be a really great feature that would improve the possibility of people collaborating on the documents. Is it possible this feature will make it into the roadmap?

ssddanbrown:
Further request and ideas proposed within #3063

TiemeWoldman:
Any plans on adding annotations (Inline comments) to BookStack? This would be a great addition. 

I tried [hypothes.is](https://hypothes.is/) for adding annotations to BookStack pages, but the BookStack text select pop-up interferes with hypothes.

EDMJohnny:
As far as I can tell BookStack is pretty close to perfect for a writing / collaboration platform. Super simple to use, very intuitive.  Inline comments would make it perfect!  Is this something you are working on or have an ETA for?
Thanks

ssddanbrown:
@EDMJohnny

>  Is this something you are working on or have an ETA for?

Not at this time.

EDMJohnny:
Thanks for responding.  Is this something you'd be willing to put a bounty on / get paid for completing?  Do you have a ballpark for cost of adding this feature?

ssddanbrown:
@EDMJohnny I'm not really a fan of public bounties, can just cause issues. Not really sure on ballpark costs either as this would need fleshing out into a full proposal since this request could be imagined/desired in different forms.

I've quoted features before (although never accepted) but only for features with clear scope and platform fit. Something makes me uneasy about rushing through a proposal for the sake of money, feel that risks providing an unbalanced larger voice to those with money, not sure the right way to balance that. If I was to quote it'd be somewhere upwards of £1k, even for a minimal implementation.

danhansen:
Why not integrate something like https://github.com/recogito/recogito-js ? It looks like a significant portion of the functionality is already implemented.

BloodyIron:
[...]

PR #5584, opened 2025-04-18
Content Comments

ssddanbrown:
Related to #1265

## Todo

- [x] Inline comment display
  - [ ] ~~Option for controlling display~~
    - Omitting this for initial implementation as we'll gather some feedback about best placement and control level.
  - [x] Selection zone representation (on focus/click?)
  - [x] Indicator for if the comment is outdated
- [x] WYSIWYG display?
- [x] Update reference hash to be focused on text content, otherwise too much variance (eg, when loaded in editor).
- [ ] ~~MD editor display?~~
  - The MD editor cycle is too loose to retain accurate references, so retaining editor support to the new WYSIWYG.
- [x] Comment archival handling
- [x] Test changes to pointer work across presentation variation (change of edit/comment permissions etc...)
- [x] Test accessible use of pointer remains, including added comment action
- [x] Test use in old WYSIWYG (references remain inactive but don't break anything).


MfeaR:
Hi! Could you please clarify if this feature will be implemented, and if so, when? We're evaluating documentation apps, and this is the only feature preventing us from choosing BookStack.

ssddanbrown:
@MfeaR This is a work in progress attempt at a feature. My aim is for this to be in the next feature release (so maybe late May or June) but can't promise anything.

EDMJohnny:
Hi - thank you for implementing this feature.  One bug I've discovered is that after a comment is created, if you edit the comment, the reference link breaks, so the comment is no longer attached to the inline selection that started it.

Issue #4575, opened 2023-09-24
Default admin user account handling plan


ssddanbrown:
This issue lays out the plan to change the way that the default admin access is granted on new instances.
Currently, a default admin account is created via migrations with the "admin@admin.com" email, and `password` password.
There are two main goals of this change:

- Stop using the `admin@admin.com` email address.
- Stop using predetermined/known credentials, causing a security issue if left unchanged.

While we could make changes easily and directly just thinking about BookStack itself, we need to think about the ways BookStack is hosted/provided to ensure we don't break the setup/install flows in various use-cases.

## Plan

Over a range of feature releases we'll make changes to work toward this incrementally.

1. Expand options of the `create-admin` command.
2. Notify BookStack hosting/service providers and projects, with advisory of options.
3. Remove the default user creation.

There probably should be a couple of feature releases' worth of time between 1 and 3.

### Create admin command changes

To provide options, and potentially seamless routes to handle these changes, we'll update the `create-admin` command with extra options:

`--initial` : This option will create the admin user, only if there are no pre-existing admin users in the system. If there's only the default admin account (`admin@admin.com`), this will instead update that account with the provided details. If this does not result in an admin account being created/updated, this will return an error status code so the caller can acknowledge if the account can be used.

`--generate-password` : This option will generate a random password for the user and return it via stdout on the command line. This should be the only output when this option is used.

With the `--initial` option added, setup providers can migrate to using this option after it's released, but before default user creation is removed, so a seamless transition can be achieved. You could even use the existing default credentials to switch to the new system without any user-facing change (could help simply or delay changes but not advisable since won't help toward main goals). The behaviour of this option means it can be ran via init scripts, without needing to query BookStack, or the environment, first to see if this is a fresh system without a (or with default) admin user.

The password gen option, combined with the initial option, can open some options to providers to help migrate to a more secure setup.

### Official guidance/script changes

On the official side of things, we'll need to:

- Change our currently supported install scripts to create a user with a random password, then output to user.
- Update our docs with this noted in our install and script guidance.
- Pin comments to our install videos to advise of this change.
- Probably worth creating a post in the Subreddit for SEO purposes, for when folks search for default credentials.

We can update our scripts after the initial addition of the command options.
These can then be used as an example for others to update their scripts/setups with.

## Setup Providers to Notify

We'll need to notify these folks when the time comes (After commands have been implemented).
This list will be used for tracking when the time comes.
**Please don't notify these projects/groups/people yet!**

[...]

PR #5596, opened 2025-04-30
Add an artisan command to change password

ChrisHills463:
With this change you can change a password from the cli with artisan. This is especially useful when automating the deployment of BookStack using Ansible or similar.

Usage:-

```php artisan bookstack:change-password --email admin@admin.com --password Sup3rS3cur3Pa55w0rD```

### DIFF
--- a/app/Console/Commands/ChangePasswordCommand.php
+++ b/app/Console/Commands/ChangePasswordCommand.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace BookStack\Console\Commands;
+
+use BookStack\Users\Models\Role;
+use BookStack\Users\UserRepo;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Hash;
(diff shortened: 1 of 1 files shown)


ssddanbrown:
Thanks for offering this @ChrisHills463, and sorry for the late response.
This could be considered related to #4575. 

In that case could be better potentially served by a general "edit-user" command, with ability to also change basic user details (name,email,password,external_auth_id) so this could be used for creating unique defaults, in addition to helping other sysadmin level tasks (like scripting updates to the `external_auth_id`).

ssddanbrown:
Thanks again for offering this @ChrisHills463, but I'm going to close it off in favour of changes made to the create admin command in  #5749, which aim to address the fundamental scenario of being able to change the first-user-password, in a forward compatible manner as described in #4575. 

I'm not totally closed to the idea of introducing some form of user edit commands if needed beyond the changes in #5749, but I'd want to see some wider need/desire before introducing something new.

PR #5606, opened 2025-05-13
Add settings for number of books/shelves that will be displayed per page

### LINKED ISSUE #2343: Feature Request - Define number of shelves to display on one page
The ability for an admin to set a global value for the number of shelves to display on one page before showing the next page option.

Allow users to chose how many shelves are displayed on a single page allowing a level of customisation (personal preference is a single page with all shelves rather than breaking it up in to pages). I understand there is probably a reason on the performance aspect which led to the current number.

Xenoamor:
A trivial addition to allow setting the number of paginated results on the books/shelves tab. Unfortunately I could not find a way to do this using the Logical Theme System

I'm not sure if these settings would be better placed under the "Customization" tab but I have placed them under "Sorting" for the moment. Unfortunately this does require 4 new strings in the localization files

This would close #2343

### DIFF
--- a/app/App/HomeController.php
+++ b/app/App/HomeController.php
@@ -83,7 +83,7 @@ class HomeController extends Controller
         if ($homepageOption === 'bookshelves') {
             $shelves = $this->queries->shelves->visibleForListWithCover()
                 ->orderBy($commonData['listOptions']->getSort(), $commonData['listOptions']->getOrder())
-                ->paginate(18);
+                ->paginate(intval(setting('sorting-shelves-per-page', '18')));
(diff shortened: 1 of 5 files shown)


ssddanbrown:
Thanks for providing this PR @Xenoamor, and sorry for the long wait on a response for this one.

This is now merged in, to be part of the next release.
I made some follow-up changes in https://github.com/BookStackApp/BookStack/commit/10f5ceee35c234ee638f4bfd3b1d4ea790041a3e, to add some testing, to add search results page counts, and some other bits to generally fit these options into the UI a bit better, and to make their usage a bit easier.

PR #5662, opened 2025-06-18
Add MetricsApiController with /api/status endpoint

Jandalslap:
Introduced MetricsApiController to expose basic system stats for BookStack. Provides a JSON response with counts for shelves, books, pages, and users. Registered the endpoint at /api/status for use in monitoring and dashboards. 

Specifically getHomepage in my case.

Screenshot 2025-06-19 095603

Using the customapi service feature settings below:

- Documentation:
    - bookstack:
        icon: /icons/bookstack.png
        href: "{{HOMEPAGE_VAR_BOOKSTACK_HREF}}"
        ping: http://bookstack:80
        description: document manager
        widget:
            type: customapi
            url: http://bookstack:80/api/status
            refreshInterval: 60000
            method: GET
            display: block
            mappings:
                - field: shelves
                  label: Shelves
                  format: number            
                - field: books
                  label: Books
                  format: number
                - field: pages
                  label: Pages
                  format: number
                - field: users
                  label: Users
                  format: number

### DIFF
--- a/app/Api/MetricsApiController.php
+++ b/app/Api/MetricsApiController.php
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * MetricsApiController
+ * 
+ * Provides a lightweight endpoint for retrieving basic BookStack usage statistics,
+ * including the total count of shelves, books, pages, and users. This endpoint
+ * is useful for health checks, dashboards, or external monitoring integrations.
+ * 
+ * Route: GET /api/status
+ */
+
+namespace BookStack\Api;
+
+use BookStack\Http\Controller;
+use BookStack\Entities\Models\Bookshelf;
+use BookStack\Entities\Models\Book;
+use BookStack\Entities\Models\Page;
+use BookStack\Users\Models\User;
+
+class MetricsApiController extends Controller
+{
+    /**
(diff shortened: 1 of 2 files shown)


ssddanbrown:
Thanks for offering this @Jandalslap, but I'm not keen on merging because:

- This does not provide new data, but is a collection of what can be achieved via other endpoints.
- The benefit of this appears use-case specific, and is not something I've had notable prior demand of.

Jandalslap:
Yes, you're right, I was just trying to consolidate into one endpoint. Homepage is a popular dashboard though and I was hoping to incorporate Bookstack into one of their permanent app ready services. A lot of self hosting enthusiasts see those services offered so its good exposure but i understand if its not what you’re after. Their customapi service requires only one endpoint however as a built in service it can probably use what’s already available. I was trying to show what was possible first.

ssddanbrown:
Okay. I'll therefore close this off.

You could technically add your combined custom endpoint via the [logical theme system](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/logical-theme-system.md), but a custom built service for homepage sounds like a better fit if it can suit other instances by default out of the box without needing to apply customizations.

PR #5666, opened 2025-06-19
Diagram Rendering

ssddanbrown:
Test to see if we can provide a hybrid approach to drawings, by using png images as we do now, but extract and display their contents in a viewer for interactive convenience.


ssddanbrown:
Closing as I resorted to a hack-based solution instead.

Thread: https://github.com/BookStackApp/BookStack/issues/1170#issuecomment-2994154948
Hack: https://www.bookstackapp.com/hacks/interactive-drawings/

PR #5668, opened 2025-06-22
CommentDisplayTest correct namespace

bumperbox:
Correct namespace error

Class Entity\CommentDisplayTest located in ./tests/Entity/CommentDisplayTest.php does not comply with psr-4 autoloading standard (rule: Tests\ => ./tests). Skipping.


ssddanbrown:
Thanks @bumperbox!

I just cannot get PHPStorm to understand/use the correct namespaces when it comes to tests! :facepalm:

PR #5722, opened 2025-07-18
refactor: api CRUD routes now have dedicated files

TheKiryuKha:
Moves api CRUD routes into special files for better understanding


ssddanbrown:
Thanks for offering @TheKiryuKha but I prefer the longer single list.

Issue #2662, opened 2021-03-30
Drawn entities don't work properly with dark mode enabled.


nickolasdeluca:
**Describe the bug**
Drawn entities don't work properly with dark mode enabled.

**Steps To Reproduce**
1. Turn on dark mode.
2. Create a page.
3. Edit the page.
4. Draw any entity on flowchart.
5. Save drawing.

**Expected behavior**
I believe all entities should have their colors inverted. Black turns White and White turns black. Fonts, lines and all objects.

**Screenshots**
![error](https://user-images.githubusercontent.com/17858166/113001225-fccd8e00-9146-11eb-809b-2da38d8eb3f0.png)
![error2](https://user-images.githubusercontent.com/17858166/113001242-ffc87e80-9146-11eb-8db6-9a5e54f19df0.png)

**Your Configuration (please complete the following information):**
 - Exact BookStack Version (Found in settings): v0.31.2
 - PHP Version: 7.4.9 (cli) (built: Aug  7 2020 14:29:10) ( NTS )
 - Hosting Method (Nginx/Apache/Docker): Nginx

ssddanbrown:
Hi @nickolasdeluca,

This is going to be the case for an user images that are on transparent backgrounds. I wouldn't really be comfortable inverting the color of images by default as we'd be altering user content and it's likely to have side-affects. A simple inversion is not always what would be wanted.

If this is a problem on your instance, you could set a default white background to drawings (and a border for padding) using the following in your "Custom HTML Head Content" setting:

```html
<style>
.page-content img[src*="/drawio/"] {
  background-color: #FFF;
  border: 12px solid #FFF;
}
</style>
``` 

Or if you did want inversion in dark mode:

```html
<style>
.dark-mode .page-content img[src*="/drawio/"] {
  filter: invert(1);
}
</style>
```

TechInterMezzo:
Hi @ssddanbrown,
I am a bit late for this issue but I found the "keepTheme" option in the [docs of draw.io](https://www.diagrams.net/doc/faq/embed-mode). It seems to be an optional parameter for the export action and should apply the current draw.io theme when rendering the png. So when you turn on the draw.io dark mode then the appearance of the saved PNG should match.

I think you would only need to add
keepTheme: true
to
https://github.com/BookStackApp/BookStack/blob/development/resources/js/services/drawio.js#L58

But I haven't tested it myself.

EricKotato:
[...]

PR #5745, opened 2025-08-02
Keep draw.io theme when saving a diagram

EricKotato:
Even though #2662 is marked as solved, I don't think it's solved properly since the colors in dark mode might be not the simple invert.

@TechInterMezzo has discovered "keepTheme" option (see https://github.com/BookStackApp/BookStack/issues/2662#issuecomment-1207450833 for details), and I recently rediscovered it.

I've also tested it, and can confirm that option does work properly:
изображение

### DIFF
--- a/resources/js/services/drawio.ts
+++ b/resources/js/services/drawio.ts
@@ -37,7 +37,7 @@ function drawEventExport(message: DrawioExportEventResponse) {
 
 function drawEventSave(message: DrawioSaveEventResponse) {
     drawPostMessage({
-        action: 'export', format: 'xmlpng', xml: message.xml, spin: 'Updating drawing',
+        action: 'export', format: 'xmlpng', xml: message.xml, spin: 'Updating drawing', keepTheme: true,
     });
 }
 


ssddanbrown:
Thanks for the PR @EricKotato, and sorry for the very late response.
Do you have any other option set for drawings in BookStack to use dark mode?
The default settings always open in light mode for me, regardless of browser/bookstack preference.

I'm worried that using this setting may result in further added variance in multi-user environments, so we might need to provide some level of higher-level control, but just trying to think out & test potential scenarios first.

ssddanbrown:
Since there's been no further follow-up I'll close this off for now but happy to re-open this if desired.

mo-st:
I tried looking into this but couldn't get it to work either, I think it probably needs more finetuning.
Edit: I think I'm stupid and need to figure out first how to persist the edit inside my container

I applied the proposed change in the PR plus the dark theme preference for the embedded drawio:
`DRAWIO=https://embed.diagrams.net/?embed=1&proto=json&spin=1&configure=1&ui=dark`

Thanks for your work, I'll try to gather more info when I find time

PR #5797, opened 2025-09-10
fix: empty space for the empty description field

### LINKED ISSUE #5724: Empty description field adds empty line on save
## Describe the Bug

For example, when a book's description is empty, editing and saving it will result in HTML tags being appended to the empty field. As a result, an empty `<p>` tag is rendered in the description.

Image

## Quick Fix

`app/Entities/Models/HasHtmlDescription.php`

```php
/**
 * Check if the entity has a meaningful description (not just empty HTML tags).
 */
public function hasDescription(): bool
{
    $cleanedHtml = str_replace(['<p>', '</p>', '<br>', '<br/>', "\xC2", "\xA0", " "], '', $this->descriptionHtml());
    return !empty(trim($cleanedHtml));
}
```

- `resources/views/books/show.blade.php`
- `resources/views/chapters/show.blade.php`
- `resources/views/shelves/show.blade.php`

```php
@if($book->hasDescription())
    <div class="text-muted break-text">{!! $book->descriptionHtml() !!}</div>
@endif
```

### Steps to Reproduce

1. Go to [https://demo.bookstackapp.com/books/bookstack-demo-site](https://demo.bookstackapp.com/books/bookstack-demo-site)
2. Click `Edit` on the right side.
3. Remove description content and save.
4. Click `Edit` again and save, it will add an empty line(`<p>`).

### Expected Behaviour

- No empty line.
- Trim empty description.

_No response_

### Browser Details

_No response_

### Exact BookStack Version

v25.05.2

MohammadAlhallaq:
Fix empty description rendering 
CLOSES #5724 

Centralized empty HTML trimming in HasHtmlDescription.php and updated views to prevent rendering <p> tags for empty descriptions.

### DIFF
--- a/app/Entities/Models/HtmlDescriptionTrait.php
+++ b/app/Entities/Models/HtmlDescriptionTrait.php
@@ -13,6 +13,7 @@ trait HtmlDescriptionTrait
     public function descriptionHtml(bool $raw = false): string
     {
         $html = $this->description_html ?: '<p>' . nl2br(e($this->description)) . '</p>';
+        $html = preg_replace('/<p>(\s|&nbsp;|<br\s*\/?>)*<\/p>/i', '', $html);
         if ($raw) {
             return $html;
         }


ssddanbrown:
Thanks for offering this @MohammadAlhallaq, and sorry for the very delayed response.

Currently this method would impact legitimate scenarios (empty lines between content).
Maybe we use this approach with enough regex, but I instead chose a different approach within commit 229a99ba24d7676f5608bcb662bbe047d4a3bf42.
Therefore I'm going to close this off.

PR #5820, opened 2025-10-02
Fix PWA manifest access behind authenticated proxies

tfnh621:
In environments where BookStack is served behind an authenticated proxy (such as Cloudflare Zero Trust), the browser fails to fetch `manifest.json` since credentials are not sent by default.
As a result, PWA installation and functionality do not work.

This PR resolves the issue by adding `crossorigin="use-credentials"` to the `<link rel="manifest">` tag, ensuring credentials are included when the manifest is requested.

Reference:
https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest#deploying_a_manifest


ssddanbrown:
Thanks @tfnh621!

PR #5826, opened 2025-10-08
feat(middleware): Page 404 to 301 | add redirect for old page slugs, caused by title change

oopen:
### 📝 Description  

This PR introduces a new Laravel middleware, `RedirectOldPageSlugs`, that automatically redirects requests containing **obsolete page slugs** to the page’s stable permalink (`/link/{id}`), which then redirects to the current canonical URL.

#### 🔍 Problem  
When a user changes a page’s slug in BookStack:
- The old URL becomes a 404.
- BookStack **does not store historical slugs for books or chapters** (only pages have revisions via `page_revisions`).
- External links, bookmarks, or search engine results pointing to the old URL break permanently.

This creates a poor user experience and harms SEO.

#### ✅ Solution  
The new middleware:
- Runs early in the `web` middleware stack (after `TrustProxies`, before `PageViewCounter`).
- Extracts the last path segment from the request (e.g., `/book/foo/bar` → `bar`).
- Checks if this segment exists as a **non-current slug** in `page_revisions`.
- If found, issues a **301 redirect** to `/link/{page_id}` — BookStack’s built-in, stable permalink route.
- Skips API, asset, debug, and system routes to avoid interference.

This approach is:
- **100% reliable** for pages (uses real revision data).
- **Independent of book/chapter slug history** (which BookStack doesn’t store).
- **Zero-config** and fully automatic.
- **Safe**: only redirects when an exact match is found in revisions.

#### 🧪 Testing  
1. Create a page → note its slug.
2. Edit the page and change its slug.
3. Visit any URL ending with the **old slug** (e.g., `/any/book/or/chapter/path/old-slug`).
4. ✅ Observe a 301 redirect to `/link/{id}`, then to the current page.

> Note: The middleware matches **any path ending with the old slug**, because BookStack’s routing is dynamic and doesn’t enforce strict path structures. This ensures maximum compatibility.

#### 📁 Files Changed  
- `app/Http/Middleware/RedirectOldPageSlugs.php` → new middleware  
- `app/Http/Kernel.php` → register middleware in `web` group

#### ⚙️ Middleware Placement  
Inserted after `TrustProxies` and before `PageViewCounter` to:
- Avoid counting redirected requests as page views.
- Ensure trusted proxy headers are processed first.

#### 💡 Why `/link/{id}` and not the canonical URL?  
- `/link/{id}` is BookStack’s official, stable permalink mechanism.
- It already handles final redirection to the current URL (including updated book/chapter slugs).
- This avoids duplicating routing logic and ensures consistency.

---

### ✅ Benefits  
- Restores broken links automatically.
- Improves SEO by preserving link equity via 301s.
- Requires no user intervention or configuration.
- Minimal performance impact (indexed DB lookup on `page_revisions.slug`).

---

### 🙏 Final Note  
This feature aligns with BookStack’s philosophy of usability and data integrity, and leverages existing systems (`page_revisions`, `/link/{id}`) without introducing redundancy.

Thank you for considering this contribution!

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -17,6 +17,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \BookStack\Http\Middleware\TrimStrings::class,
         \BookStack\Http\Middleware\TrustProxies::class,
+        \BookStack\Http\Middleware\RedirectOldPageSlugs::class,
         \BookStack\Http\Middleware\PreventResponseCaching::class,
     ];
 
--- a/app/Http/Middleware/RedirectOldPageSlugs.php
+++ b/app/Http/Middleware/RedirectOldPageSlugs.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace BookStack\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use BookStack\Entities\Models\PageRevision;
+
+class RedirectOldPageSlugs
+{
+    public function handle(Request $request, Closure $next)
+    {
+        // Skip system/API routes
+        if ($request->is('api/*', 'assets/*', 'dist/*', 'uploads/*', '_debugbar/*', 'link/*', 'saml2/*')) {
+            return $next($request);
+        }
+
+        $path = trim($request->path(), '/');
+        if (empty($path)) {
+            return $next($request);
+        }
+
+        // Extract last path segment (potential page slug)
+        $segments = explode('/', $path);
+        $lastSegment = end($segments);
+
+        // Look for this slug in page revisions (only if it's not the current slug)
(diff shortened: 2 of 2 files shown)


ssddanbrown:
Thanks for offering this @oopen, but this isn't the approach I'd like to take, and as mentioned in #5827 I'm awaiting on other changes before adding this kind of functionality.

Also, please avoid writing PRs (text/code) with heavy AI use for this project.
It adds a lot of extra incorrect information, and code that does not understand the full context which can result in issues or odd behaviour/decisions.

oopen:
Ok, I read your plan to update the DB first, it a good thing.

But please consider this redirection feature as a "temporary mitigation hack", 
and can be removed after when you complete the plan you described https://github.com/BookStackApp/BookStack/pull/5800

It make no change to the app/DB, just fix a big part of 404 links immediately.

PR #5847, opened 2025-10-21
Extended logic to support more than one OIDC provider for authentication

MarchoStoev:
Hi,
I tried extending the logic in the application so it can support multiple OpenID Connect OAuth providers.
Configuration changes:

1. Environmental variables changes

- New variable OIDC_PROVIDERS=provider1,provider2,provider3. Those will be the names used to map OAuth variables to specific provider.
- Other environmental variables should follow the standard naming convention with the name of the provider after the OIDC_ prefix. (Example: OIDC_PROVIDER1_CLIENT_ID=.....)
That's for the configuration part. The mapping will be done in the oidc configuration by:
1.1. Getting all providers from OIDC_PROVIDERS variable
1.2. Splitting them on every ","
1.3. Gathering the provided variables for all providers in the .env

2. Routes changes
- Only changes are to the oidc routes by adding /{provider} to the endpoints so the backend knows which provider's configuration to use. {provider} is the provider, specified in OIDC_PROVIDERS variable.

3. Controller and services changes
- All places which read the config are changed to get the configuration for the specified provider from the endpoint.

4. Session changes
- In the user session after a successful login I've saved a variable, named oidc_provider which stores the provider used to login, so I can know which provider to put in the logout button in the frontend.
- All places that save in the session something about the OIDC, I've added the provider as an affix.

5. View changes
- Dynamically adds all buttons in the login-form-oidc.blade.php
- Reads from session the provider to put in the logout button in header-user-menu.blade.php

Those are all the changes I've implemented. I tried my best to keep it as clean as I can. I hope it's useful to you. I'm open of course to feedback and all change suggestions will be much appreciated. Feel free to use my code however you like for the application. I feel like it will be an useful addition to the app.

Kind regards,
Mario Stoev

### DIFF
--- a/app/Access/Controllers/LoginController.php
+++ b/app/Access/Controllers/LoginController.php
@@ -33,6 +33,7 @@ class LoginController extends Controller
         $socialDrivers = $this->socialDriverManager->getActive();
         $authMethod = config('auth.method');
         $preventInitiation = $request->get('prevent_auto_init') === 'true';
+        $oidcProviders = array_keys(config('oidc.providers'));
 
         if ($request->has('email')) {
             session()->flashInput([
@@ -53,6 +54,7 @@ class LoginController extends Controller
         return view('auth.login', [
             'socialDrivers' => $socialDrivers,
             'authMethod'    => $authMethod,
+            'oidcProviders' => $oidcProviders
         ]);
     }
 
--- a/app/Access/Controllers/OidcController.php
+++ b/app/Access/Controllers/OidcController.php
@@ -20,17 +20,17 @@ class OidcController extends Controller
     /**
      * Start the authorization login flow via OIDC.
      */
-    public function login()
+    public function login(string $provider)
     {
         try {
-            $loginDetails = $this->oidcService->login();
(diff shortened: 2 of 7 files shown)


ssddanbrown:
Hi @MarchoStoev,

Thanks for offering this PR!
That said, I'd be hesitant to merge this as I'm not keen on adding further complexity to the existing OIDC implementation without some reasonable demand, and since I can't see any prior request for this functionality, I'm not sure it balances out to be worthwhile.

MarchoStoev:
Hello @ssddanbrown,

Thanks for your very quick response!
I understand your concern. We're integrating bookstack in a project of the company I work in and we desperately needed the functionality, so I decided to take on the challenge to upgrade it. Hoped it would be of use to you as well, so we have an easier time when new versions come.
The code structure was very clean and it was a pleasure working on it.

Thank you for the time.

ssddanbrown:
Okay, I'm therefore going to close this off as per my view/comments above, but feel free to open a feature request for this functionality and link to this, so it can either gain support and/or be a useful reference point for others wanting to customize with the same functionality.

PR #5907, opened 2025-11-20
Do not use deprecated NAMEID_EMAIL_ADDRESS as default for SAML2 logout

jdede:
For SAML 2.0 **logout**, the "`NAMEID_EMAIL_ADDRESS`" (`urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`) is used as the default value. As the value is set, it can **not** be overwritten in the `onelogin` framework for example by setting something like

    SAML2_ONELOGIN_OVERRIDES: '{"sp":{"NameIDFormat":"urn:oasis:names:tc:SAML:2.0:nameid-format:persistent"}}'

Further, the `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress` is outdated by IDMs like shibboleth.

By removing this line, the default settings of the underlying framework are being used and users can adapt the value according to their needs by using the overrides.

### DIFF
--- a/app/Access/Saml2Service.php
+++ b/app/Access/Saml2Service.php
@@ -65,8 +65,7 @@ class Saml2Service
                 [],
                 $user->email,
                 $sessionIndex,
-                true,
-                Constants::NAMEID_EMAIL_ADDRESS
+                true
             );
             $id = $toolKit->getLastRequestID();
         } catch (Error $error) {
--- a/app/Config/saml2.php
+++ b/app/Config/saml2.php
@@ -78,7 +78,7 @@ return [
             // Specifies constraints on the name identifier to be used to
             // represent the requested subject.
             // Take a look on lib/Saml2/Constants.php to see the NameIdFormat supported
-            'NameIDFormat' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress',
+            'NameIDFormat' => 'urn:oasis:names:tc:SAML:2.0:nameid-format:persistent',
 
             // Usually x509cert and privateKey of the SP are provided by files placed at
             // the certs folder. But we can also provide them with the following parameters


ssddanbrown:
Thanks for the PR @jdede.

> Do not use deprecated NAMEID_EMAIL_ADDRESS as default for SAML2 logout

Do you have any relatively official guidance as to how that's been deprecated? From searching I can't find any notice/spec defining that it's been deprecated at all.

I'm not keen on changing the defaults as per this PR, as I'm concerned this will cause breaking changes for existing users. 
Plus there are a few other parts of the code which will need to change if we start expecting non-email based IDs, since our logic does expect it to be the email address.

Generally I don't think it'd be worth supporting variation here unless there's a wider proven need, and at the moment it seems like this is a need based on the defaults of one auth provide, which can support NAMEID_EMAIL_ADDRESS with a little configuration as far as I can tell?

jdede:
Thank you for your reply @ssddanbrown 

The nameid-format Email is from SAML 1.1 and got changed to, for example, `urn:oasis:names:tc:SAML:2.0:nameid-format:transient` or `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent` according to Line 3279 in https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf

As far as I understand from the discussion, the main reason is that the mail address might be changed and also contains personal information which might not be ideal for some use cases. Further, I could be guessed. The persistent or transient identifier can be arbitrary pseudo-string.

ssddanbrown:
> The nameid-format Email is from SAML 1.1 and got changed to, for example, urn:oasis:names:tc:SAML:2.0:nameid-format:transient or urn:oasis:names:tc:SAML:2.0:nameid-format:persistent according to Line 327

To quote that line:

> Note: Several identifiers that were deprecated in SAML V1.1 have been removed for SAML V2.0.

I don't believe that means that all 1.1-originating identifiers have been deprecated. Since the `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress` identifier has not been removed from the spec, and is detailed just a few lines below without a warning, I don't think that is related to this identifier at all.

 Going by the details [found here](https://docs.oasis-open.org/security/saml/Post2.0/sstc-saml1-profiles-assertion-subject.html) these are the identifiers which were deprecated and removed:

```
urn:oasis:names:tc:SAML:1.0:assertion#emailAddress
urn:oasis:names:tc:SAML:1.0:assertion#X509SubjectName
urn:oasis:names:tc:SAML:1.0:assertion#WindowsDomainQualifiedName
```

ssddanbrown:
I'm going to go ahead and close this off since, as reflected above, The value in question is not deprecated, and I don't want to be changing defaults.

Issue #5390, opened 2025-01-05
ES256 Support in OIDC 


Adiack06:
### Describe the feature you'd like

ES256 implementation for OIDC

### Describe the benefits this would bring to existing BookStack users

It would allow the use of the far more secure and up-to-date signing format which is preferable for security especially as RS256 is generally on the way out.
It would also work better for people who use Lets Encrypt for signing certs as that is what they typically provide.

### Can the goal of this request already be achieved via other means?

No

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

Under 3 months

ssddanbrown:
ES256, [following the spec](https://www.rfc-editor.org/rfc/rfc7518.html#section-3.4), is ECDSA using P-256 and SHA-256.
Looks like it should be supported by the lib we're already using to verify signatures: https://phpseclib.com/
Would need to check/validate the process/format for certs provided via config, as well as autodiscovery.

Tricky to find any useful information out there regarding widespread use/plans/changes in ES256 use for OIDC.
The [JWA spec](https://www.rfc-editor.org/rfc/rfc7518.html#section-3.1) does mark it as `recommended+`, hinting at being required in future, so may be a good indicator at specifically supporting ES256 over any other potential algorithm, but not sure about timings around that or realistic use in the OIDC landscape.

CL107:
> ES256, [following the spec](https://www.rfc-editor.org/rfc/rfc7518.html#section-3.4), is ECDSA using P-256 and SHA-256. Looks like it should be supported by the lib we're already using to verify signatures: https://phpseclib.com/ Would need to check/validate the process/format for certs provided via config, as well as autodiscovery.
> 
> Tricky to find any useful information out there regarding widespread use/plans/changes in ES256 use for OIDC. The [JWA spec](https://www.rfc-editor.org/rfc/rfc7518.html#section-3.1) does mark it as `recommended+`, hinting at being required in future, so may be a good indicator at specifically supporting ES256 over any other potential algorithm, but not sure about timings around that or realistic use in the OIDC landscape.

Hi, I've come across this issue as its something id like to see implemented as well. Be it that its marked as `recommended+` i guess its not going to take too long for support to come about. In our use case we just need to be able to sign our own auth correctly for security - and unfortunately this is the only thing keeping us away from BookStack. 

Although once this is implemented we'll probably be switching straight over from whichever alternative we decide on.

[...]

PR #5912, opened 2025-11-23
More flexible OIDC key support

McTom234:
The goal of this PR is to allow more key algorithms for OIDC signing keys.

The concerns regarding such implementation in #5390 were considered. I try to address them by allowing a flexible, extensible, and somewhat maintainable approach, which I gratefully take feedback on and am willing to improve by the feedback provided.

Therefore, the issue with this PR is not the few lines of code changed but the architecture of how more key algorithms could be supported. For presentation reasons, I added the RS512 algorithm, which I had done before.
If the suggested approach would be considered by the maintainers, I needed some support to extend and test this with more common algorithms like ES256 as described in the issue mentioned above.

Besides that, the test suits would probably need at least a test case per key algorithm to confirm that the implementations are working, which is currently a ToDo.

### DIFF
--- a/app/Access/Oidc/OidcJwtSigningKey.php
+++ b/app/Access/Oidc/OidcJwtSigningKey.php
@@ -2,6 +2,7 @@
 
 namespace BookStack\Access\Oidc;
 
+use Illuminate\Support\Facades\Log;
 use phpseclib3\Crypt\Common\PublicKey;
 use phpseclib3\Crypt\PublicKeyLoader;
 use phpseclib3\Crypt\RSA;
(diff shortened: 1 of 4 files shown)


ssddanbrown:
Marking as blocked since this would really await a second algorithm to support, pending qualification as per my comment here: https://github.com/BookStackApp/BookStack/issues/5390#issuecomment-3567922489

ssddanbrown:

https://codeberg.org/bookstack/bookstack/pulls/5912
  
Any further discussion for this will now be done on Codeberg.
Therefore this GitHub thread will be locked.

PR #5914, opened 2025-11-23
Fix: Remove validUntil attribute from SAML SP Metadata

jdede:
This PR disables the `validUntil` attribute in the generated SAML Service Provider (SP) metadata.

## Why this is needed
Currently, the underlying `php-saml` library hardcodes the metadata validity (`TIME_VALID`) to **2 days** and caching (`TIME_CACHED`) to **1 week** (Source: [Metadata.php](https://github.com/SAML-Toolkits/php-saml/tree/master/lib/Saml2/Metadata.php)).

In many real-world scenarios, specifically with Identity Providers like Shibboleth, these default windows are too short. This causes the IdP to deny connections or require manual metadata refreshes once the hardcoded time passes.

The `getSPMetadata` function in [Settings.php](https://github.com/SAML-Toolkits/php-saml/blob/master/lib/Saml2/Settings.php) allows for an `$ignoreValidUntil` parameter.

* I have updated the `getSPMetadata` call to set `$ignoreValidUntil` to `true`.
* This removes the `validUntil` timestamp from the XML generated at `<URL>/saml2/metadata`, preventing arbitrary expiration issues.

### DIFF
--- a/app/Access/Saml2Service.php
+++ b/app/Access/Saml2Service.php
@@ -157,7 +157,7 @@ class Saml2Service
     {
         $toolKit = $this->getToolkit(true);
         $settings = $toolKit->getSettings();
-        $metadata = $settings->getSPMetadata();
+        $metadata = $settings->getSPMetadata(ignoreValidUntil: true);
         $errors = $settings->validateMetadata($metadata);
 
         if (!empty($errors)) {


ssddanbrown:
Thanks @jdede and sorry for the delay in response.

> In many real-world scenarios, specifically with Identity Providers like Shibboleth, 

Have you experienced this in other real-world scenarios outside of shibboleth? Just want to validate that statement to understand how problematic this is, since this isn't something I've seen be raised as an issue before.

Again, I'm wary of changing defaults to suit issues found on a specific identity provider.
I'm not familiar with Shibboleth though. How does it actually read BookStack's metadata in this case? Is that done via a direct call to the metadata endpoint, in a way that gets refreshed? Is is done via XML upload? If not, can it be done via XML upload?

jdede:
As far as I understand, there are two different ways of thinking:

1. The metadata is static. It is sufficient to pull it once or every couple of weeks / month to prevent a possible attack by hijacking the metadata endpoint.
2. The metadata should be pulled more or less daily / with every request and keep it short-living.

Therefore, I would like to have it configurable depending on the used identity provider.

ssddanbrown:
I'm not really sure what's stopping either approach with an IDP though.

In your original issue, how is the BookStack metadata being used? Is it being re-pulled by Shibboleth frequently? Or was this from manual upload?

ssddanbrown:
Since there's been no further follow-up, and since I'm not keen on changing defaults without established need, I'm going to close this off.

PR #5952, opened 2025-12-17
Update "Microsoft URL Rewrite Module for IIS" download link

gerundt:
The download link to the *Microsoft URL Rewrite Module for IIS* in the `web.config` file is no longer available.


ssddanbrown:
Thanks @gerundt!

PR #5969, opened 2025-12-30
 Git 2.35+ may refuse to operate on bind-mounted repos with differing ownership ("dubious ownership"), Mark /app as safe within the container.

shaoliang123456:
Local Environment:

WSL2 + Windows 11 + Docker Desktop + Git 2.43.0

After starting the application with docker compose up, the logs show the following warning:

The repository at "/app" does not have the correct ownership and git refuses to use it:
fatal: detected dubious ownership in repository at '/app'
To add an exception for this directory, call:
	git config --global --add safe.directory /app
This occurs because Git 2.35.2+ enforces ownership checks on repositories, and the bind-mounted /app directory in the container has a different UID than the container’s runtime user. While this doesn’t break the web application itself, it prevents developers from using Git commands inside the container (e.g., for debugging or version detection), and clutters the logs.

This PR automatically adds /app to Git’s safe.directory list in the app container’s entrypoint, resolving the issue without affecting functionality.

### DIFF
--- a/dev/docker/Dockerfile
+++ b/dev/docker/Dockerfile
@@ -14,6 +14,9 @@ RUN apt-get update && \
         wait-for-it && \
     rm -rf /var/lib/apt/lists/*
 
+# Mark /app as safe for Git >= 2.35.2
+RUN git config --system --add safe.directory /app
+
 # Install PHP extensions
 RUN docker-php-ext-configure ldap --with-libdir="lib/$(gcc -dumpmachine)" && \
     docker-php-ext-configure gd --with-freetype --with-jpeg && \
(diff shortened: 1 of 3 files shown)


ssddanbrown:
Thanks @shaoliang123456.
Could this instead be added to the Dockerfile for the development app container? 
We only really expect users to maybe use that container interactively for development, and since it's a global option I think this can be part of the container image, which will then take force even if the user is overriding the entry-point to jump directly into a bash shell.

shaoliang123456:
> Thanks @shaoliang123456. Could this instead be added to the Dockerfile for the development app container? We only really expect users to maybe use that container interactively for development, and since it's a global option I think this can be part of the container image, which will then take force even if the user is overriding the entry-point to jump directly into a bash shell.

Thanks for the suggestion. I've moved the git config --system --add safe.directory /app command into the Dockerfile.
This ensures that /app is recognized as a safe directory regardless of how the container is started—even if the entrypoint is overridden (e.g., when launching a shell directly for debugging).
I’ve tested the change and the “dubious ownership” warning no longer appears in the logs. Git commands now work as expected in the container.

ssddanbrown:
Thanks @shaoliang123456

Issue #5650, opened 2025-06-09
Configurable Public URL Sharing Toggle per Page Summary


s-sandeep-tech:
### Describe the feature you'd like

Add a configurable option per page to control public URL sharing:

A page property: Allow Public URL Sharing (On/Off)

If On: show a "Share Public URL" option (with a tokenized link if needed).

If Off: block attempts to share/access the URL unless proper authentication/authorization is done.

System default setting: new pages default to On/Off as per admin config.

API and export support: include this flag in page metadata.

Audit log support: log when sharing state is changed.

### Describe the benefits this would bring to existing BookStack users

Give page owners control over what content is externally shareable.

Prevent accidental leaking of sensitive or draft content.

Support workflows where only some content is intentionally public.

Align with enterprise/documentation governance policies.

### Can the goal of this request already be achieved via other means?

nope

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

Not using yet, just scoping

ssddanbrown:
Thanks for the request @s-sandeep-tech but I'm going to close this off as a duplicate of the fundemental request in #288
.

> Can the goal of this request already be achieved via other means?
> nope

You can technically make specific bits of content public via permissions, but it does require the instance to be publicly accessible by default (with permissions preventing to content by default) which does reduce a layer of security.

Issue #5105, opened 2024-07-08
Ability to share specific books with specific users.


odootek:
### Describe the feature you'd like

As the title says Ability to share specific books with specific users.

Right now there is no way to share a specific book with a specific user. The only option to share is allow view/edit all book to a user using role. 

### Describe the benefits this would bring to existing BookStack users

I use bookstack for personal documentation, sometimes I need to share a travel guide, a tutorial, a food recipe or etc. to someone, an user o a guest. That is not possible. I have to create a role for a specific user then override permission, then it get bugged and I lose access to the pages I as the owner of the book. but the guest user get access.

### Can the goal of this request already be achieved via other means?

I do not know

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

1 to 5 years

### Additional context

It will be cool in book permission to be able to add multiple users with their permission to view or edit or allow public view.

![bug](https://github.com/BookStackApp/BookStack/assets/115769694/6dcf12d5-0bd3-4cb9-9dc3-6d546653eec3)

Lost access to a page when using book permission. It is hidden and no way to recover. guest account can see it, admin account can see it. just not the owner.

guest user/ public role can se the page and not the owner.

![tv](https://github.com/BookStackApp/BookStack/assets/115769694/ad095a12-a68b-4cbb-8076-c8dd83098f0b)

ssddanbrown:
Thanks for the request @odootek, but I'd consider this already covered by existing issues #1747 and #288 so I'll therefore close this off as a duplicate.

Issue #4712, opened 2023-12-05
Temporary Links


chdcomputers:
### Describe the feature you'd like

A temporary link should be a record in a special table in the db with a global unique id, a target which will be an existing shelf, book, chapter or page, an optional description and a valid until date time.

For example:
GUID00001 - docs.example.com/books/book1/page/page1 - Offer for app developing to customer1 - 2023/12/31 00:00:00

When any NOT logged in user visits the link **docs.example.com/GUID00001** before 2023/12/31 00:00:00 should be re-directed to **docs.example.com/books/book1/page/page1**

Normally the shelf, book, chapter or page that will be used as a temporary link target will not have the view permission for the public user since we do not want it to be viewable by everyone who visits the site but rather only by the visitors who have the temporary link to it. So the security system should take that into account.

In addition to the above the temporary link visits should be audited and perhaps can trigger a webhook in order to have information on when and how many times they have been used by the visitors. Moreover, the admin should have the ability to delete or modify any temporary link.

Thank you for this great project and, of course, for your time.

### Describe the benefits this would bring to existing BookStack users

In our use case we have some books with procedures and product information that we use when we are making offers to our clients. We don't want them to be public for everyone to see, however we want them to be visible (for a limited time) by those clients that got an offer from us. Since we are talking about tens or hundred clients, creating user accounts for each one of them and provide them the login details just to see this content isn't practical. With the temporary links approach we can share the temporary link with the client so they can view it (without logging in) for the time span that the offer is valid.

### Can the goal of this request already be achieved via other means?

I don't think so

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

Under 3 months

ssddanbrown:
Thanks for your request @chdcomputers, although I'm going to close this off since I feel the fundamental request is already covered by #288.

chdcomputers:
Hi @ssddanbrown ,
Thank you for your prompt reply.
I saw #288 before adding this and on the fundamentals (sharing) it is similar so I am going to comment over there about the specifics of my idea...

Issue #4611, opened 2023-10-12
public access for one book or one page


ArtisWorld:
### Describe the feature you'd like

Hi,

it would be cool, if i can grant public access for a person, who i want to read one book or one page (maybe his customer documentation). With one sent link.
Do you think its possible?

Thanks. :-)

### Describe the benefits this would bring to existing BookStack users

every bookstack user can grant someone to access the page or book with one link (with password?).

### Can the goal of this request already be achieved via other means?

yes, i can export the page and send it to this user.

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

Under 3 months

### Additional context

Great software, thank you.

ssddanbrown:
Thanks for the request @ArtisWorld,
A request to share content via a link already exists via #288, so I'm going to close this off as a duplicate request.

There is another potential workaround via:

- Enabling public access on the instance via application settings.
- Remove all default permissions from the default guest user/role.
- Override the permissions for a single specific item, to provide the guest user role permissions to view.

But this does require enabling public access and active permission control, which can have security considerations.

PR #6004, opened 2026-01-28
Implement Share Links Feature

dev:
As requested in #288, #441, #687, #732, #979, #1215, #1279, #1551, #1802, #3479, #4611, #4712, #5105, #5650 :

This PR implements a "share links" feature that allows users to create publicly accessible shareable links for books, chapters, and pages without requiring authentication.

Features:
- Create and manage share links for books, chapters, and pages
- Public access via secure token-based URLs (`/share/{token}`)
- Navigation support within shared books and chapters
- Share link management interface in settings
- Activity logging for share link operations
- Secure attachment URLs (token required)


ssddanbrown:
Thanks for offering this PR @dev.

To be honest though, when looking to implement this myself before I've thought through this style of approach but never really been happy with it, due to the large amount of duplication, re-implementation and the wide general surface-area/impact.

Looking at this PR now, I still feel the same way.
That's not to say there is a better way, just that I'm not sure that the surface area/complexity added by this approach is worthwhile long term.

ssddanbrown:
Following my comment above, I'll go ahead and close this off.

PR #6019, opened 2026-02-13
added custom training complete notification button inside the knowledge base

shakeela-metron:
added custom training complete notification button inside the knowledge base using logic theme system


ssddanbrown:
I'm going to assume these are instance specific customizations and therefore close this off.

Issue #6021, opened 2026-02-13
OIDC: group to role mapping not working when group contains space


jrttr:
### Describe the Bug

I've migrated from LDAP to OIDC-Login.
With LDAP authentication we were able to use all groups and map them to roles.
Since the migration to OIDC the group to role mapping works for groups without spaces only

### Steps to Reproduce

1. Setup OIDC in .env and enable groups_sync
```
OIDC_USER_TO_GROUPS=true
OIDC_GROUPS_CLAIM=groups
```

2. Use "OIDC_DUMP_USER_DETAIL" to verify the groups claim is working
```
OIDC_DUMP_USER_DETAILS=true
```

The details should look something like this:
(This example just shows the necessary details)
```
{
  "sub": "f:PfJpQtSaKJ3r6TFZmgi:testuser",
  "groups": [
    "group-without-spaces",
    "group with spaces"
  ],
  "preferred_username": "testuser",
  "family_name": "Test User",
  "email": "testuser@example.org"
}
```

3. create 2 roles. One with the "External Authentication IDs" `group-without-spaces` one with `group with spaces`

4. Relogin and check the roles in the user.

### Expected Behaviour

I would expect both roles mapped to the user, but instead just the `group-without-spaces` is mapped.

### Exact BookStack Version

v25.12.3

ssddanbrown:
Hi @jrttr,

As per [our OIDC docs](https://www.bookstackapp.com/docs/admin/oidc-auth/#group-sync):

> When matching OIDC groups with role names or ‘External Authentication IDs’ values, BookStack will standardise the names of OIDC groups to be lower-cased and spaces will be replaced with hyphens. For example, to match a OIDC group named “United Kingdom” an ‘External Authentication IDs’ value of “united-kingdom” could be used.

So in the "External Authentication IDs" field you can try using `group-with-spaces` to match on it instead.

jrttr:
Hi @ssddanbrown,

a classical RTFM... Thanks for the clarification!

PR #6022, opened 2026-02-16
Fix OIDC group-to-role sync for groups containing spaces

SebastienMerv:
Fix issue #6021

Normalize spaces to hyphens in parseRoleExternalAuthId() to match the same normalization applied to incoming group names in matchGroupsToSystemsRoles(), fixing group-to-role mapping for groups whose names contain spaces.


ssddanbrown:
Thanks for offering this @SebastienMerv but I'd be wary about changing the behaviour here, just in case of potential compatibility edge cases for existing values/environments, and to be safe in scenarios where users may assume the field is space separated.

Issue #6042, opened 2026-02-27
File attachment at Book level


DoctorRetromaker:
### Describe the feature you'd like

It's possible to add file attachment to books or even chapters, and not only at pages?

I primarily produce documentation for my own use in teaching IT technologies, and most of the time, I don't just use the "books" as teaching material. They're accompanied by additional materials such as various projects, which I usually package in ZIP or 7z archives.

These files aren't linked to a specific page, but rather to a topic (chapter) or even the entire syllabus of the book.

Would it be possible to add the option to attach files to the books and/or chapters?

### Describe the benefits this would bring to existing BookStack users

* More flexibility with attached files; not all attached files are related to pages
* Best organization maintaining documentation and related files together

### Can the goal of this request already be achieved via other means?

I don't think so.

Uploading files to "someplace" and adding a link to the Book or Chapter description don't do the same magic ...

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### How long have you been using BookStack?

1 to 5 years

Zhey-on:
Hi, i've implemented this feature in https://github.com/BookStackApp/BookStack/pull/6070.

ssddanbrown:
BookStack is now [managed on Codeberg](https://codeberg.org/bookstack/bookstack).

As part of this change, we're closing all existing GitHub issues and pull requests.
You'll instead find this issue on Codeberg here: 
https://codeberg.org/bookstack/bookstack/issues/6042
  
Any further discussion for this will now be done primarily on Codeberg.
Since this thread was created relatively recently, it won't be locked just yet, and we may still provide updates/responses here on GitHub, but it may be locked in the future.

PR #6044, opened 2026-03-02
File attachment at Book level #6042

inv-hareesh:
feauret: Add file upload support for books

Extend attachment system to support book-level file uploads using polymorphic
relationships. Includes full UI for uploading, managing, and organizing
attachments on book pages with proper permission checks.

- Add polymorphic database migration for attachments
- Update backend models and services for book support
- Add book attachment routes and controller methods
- Implement book attachments UI in sidebar
- Update JavaScript components for entity type handling


ssddanbrown:
Thanks for providing a PR @inv-hareesh, but I'd want to see more support for the fundamental feature being added here before we support an expanded scope, so I'm therefore going to close this off.

PR #6056, opened 2026-03-11
Add reset MFA button for admin s on user profile edit

clauvaldez:
### Summary
This pull request adds a **Reset MFA button** to the user profile edit page, allowing administrators to easily reset a user's multi-factor authentication configuration.

### Motivation
If a user loses access to their MFA device or is unable to complete the authentication process, administrators currently have no quick way to reset MFA from the interface. This feature provides a simple and controlled way for admins to resolve such situations.

### Changes
- Added a **Reset MFA** button in the user profile edit view.
- Implemented a controller method to handle MFA reset for the selected user.
- Restricted the action to users with the appropriate **UsersManage permission**.
- The reset operation removes the user's stored MFA values, forcing them to reconfigure MFA on next login.

### Security
Only administrators with the `UsersManage` permission can perform this action.

### Result
Administrators can quickly reset MFA for users who are locked out due to lost or misconfigured authentication devices, improving support and account recovery workflows.

### DIFF
--- a/app/Users/Controllers/UserController.php
+++ b/app/Users/Controllers/UserController.php
@@ -208,4 +208,17 @@ class UserController extends Controller
 
         return redirect('/settings/users');
     }
+
+    /**
+     * Reset MFA for the specified user.
+     */
+    public function resetMfa(Request $request, int $id)
+    {
+        $this->checkPermission(Permission::UsersManage);
+        $user = $this->userRepo->getById($id);
+        // Resetear el 2FA del usuario 
+        $user->mfaValues()->delete();
+        session()->flash('success', trans('settings.users_mfa_reset_success', ['userName' => $user->name]));
+        return redirect()->back();
+    }
 }
--- a/lang/en/settings.php
+++ b/lang/en/settings.php
@@ -263,6 +263,11 @@ return [
     'users_mfa_desc' => 'Setup multi-factor authentication as an extra layer of security for your user account.',
     'users_mfa_x_methods' => ':count method configured|:count methods configured',
     'users_mfa_configure' => 'Configure Methods',
+    'users_mfa_reset' => 'Reset 2FA',
+    'users_mfa_reset_desc' => 'Reset and clear all configured MFA methods for :userName. They will be prompted to reconfigure on next login.',
+    'users_mfa_reset_confirm' => 'Are you sure you want to reset 2FA for :userName?',
+    'users_mfa_reset_success' => '2FA has been reset for :userName',
+    'users_mfa_reset_error' => 'Failed to reset 2FA for :userName',
 
     // API Tokens
     'user_api_token_create' => 'Create API Token',
--- a/resources/views/users/edit.blade.php
+++ b/resources/views/users/edit.blade.php
@@ -71,6 +71,26 @@
                 </div>
             </div>
 
+            @if(user()->hasSystemRole('admin'))
+                <div class="mt-xl">
+                    <hr class="my-m">
+                    <div class="grid half gap-xl v-center">
+                        <div>
+                            <strong class="text-neg">{{ trans('settings.users_mfa_reset') }}</strong>
+                            <p class="text-small text-muted">{{ trans('settings.users_mfa_reset_desc', ['userName' => $user->name]) }}</p>
+                        </div>
+                        <div class="text-m-right">
+                            <form action="{{ url("/settings/users/{$user->id}/reset-mfa") }}" method="POST" style="display: inline;">
+                                @csrf
+                                <button type="submit" class="button neg" 
+                                        onclick="return confirm('{{ trans('settings.users_mfa_reset_confirm', ['userName' => $user->name]) }}')">
+                                    {{ trans('settings.users_mfa_reset') }}
+                                </button>
+                            </form>
+                        </div>
+                    </div>
+                </div>
+            @endif
         </section>
 
         @if(count($activeSocialDrivers) > 0)
--- a/routes/web.php
+++ b/routes/web.php
@@ -251,6 +251,7 @@ Route::middleware('auth')->group(function () {
     Route::get('/settings/users/{id}', [UserControllers\UserController::class, 'edit']);
     Route::put('/settings/users/{id}', [UserControllers\UserController::class, 'update']);
     Route::delete('/settings/users/{id}', [UserControllers\UserController::class, 'destroy']);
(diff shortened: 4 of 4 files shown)


app/Users/Controllers/UserController.php
```diff
@@ -208,4 +208,17 @@ public function destroy(Request $request, int $id)
...
+    /**
+     * Reset MFA for the specified user.
+     */
+    public function resetMfa(Request $request, int $id)
+    {
+        $this->checkPermission(Permission::UsersManage);
+        $user = $this->userRepo->getById($id);
+        // Resetear el 2FA del usuario 
+        $user->mfaValues()->delete();
+        session()->flash('success', trans('settings.users_mfa_reset_success', ['userName' => $user->name]));
```
ssddanbrown, line 221: Please can you use the `$this->showSuccessNotification` method instead to show a notification

app/Users/Controllers/UserController.php
```diff
@@ -208,4 +208,17 @@ public function destroy(Request $request, int $id)
...
     }
+
+    /**
+     * Reset MFA for the specified user.
+     */
+    public function resetMfa(Request $request, int $id)
+    {
+        $this->checkPermission(Permission::UsersManage);
+        $user = $this->userRepo->getById($id);
+        // Resetear el 2FA del usuario
```
ssddanbrown, line 219: Remove this comment, I think it's a bit redundant based on context and I'd prefer to keep comments in one language.

lang/en/settings.php
```diff
@@ -263,6 +263,11 @@
     'users_mfa_desc' => 'Setup multi-factor authentication as an extra layer of security for your user account.',
     'users_mfa_x_methods' => ':count method configured|:count methods configured',
     'users_mfa_configure' => 'Configure Methods',
+    'users_mfa_reset' => 'Reset 2FA',
+    'users_mfa_reset_desc' => 'Reset and clear all configured MFA methods for :userName. They will be prompted to reconfigure on next login.',
+    'users_mfa_reset_confirm' => 'Are you sure you want to reset 2FA for :userName?',
+    'users_mfa_reset_success' => '2FA has been reset for :userName',
+    'users_mfa_reset_error' => 'Failed to reset 2FA for :userName',
```
ssddanbrown, line 270: Please can you use "multi-factor authentication" instead of "2FA" to keep the language used consistent in the app.

Also, avoid passing through the username here. We only do that where necessary, to keep translations simpler and avoid other issues, but it can be implied by context here so I don't think it's needed.

resources/views/users/edit.blade.php
```diff
@@ -71,6 +71,26 @@ class="button outline">{{ trans('settings.users_mfa_configure') }}</a>
                 </div>
             </div>
 
+            @if(user()->hasSystemRole('admin'))
```
ssddanbrown, line 74: This does not align with the actual permission used in the controller, and we're already checking the relevant permissions before providing the view, so a permission check here is redundant.

Instead, We should probably only show this section if the user has at least one MFA option configured.

lang/en/settings.php
```diff
@@ -263,6 +263,11 @@
     'users_mfa_desc' => 'Setup multi-factor authentication as an extra layer of security for your user account.',
     'users_mfa_x_methods' => ':count method configured|:count methods configured',
     'users_mfa_configure' => 'Configure Methods',
+    'users_mfa_reset' => 'Reset 2FA',
+    'users_mfa_reset_desc' => 'Reset and clear all configured MFA methods for :userName. They will be prompted to reconfigure on next login.',
```
ssddanbrown, line 267: This is kind of incorrect. They will only be prompted to reconfigure on login **_if_** MFA is required for their role.

app/Users/Controllers/UserController.php
```diff
@@ -208,4 +208,17 @@ public function destroy(Request $request, int $id)
...
+     * Reset MFA for the specified user.
+     */
+    public function resetMfa(Request $request, int $id)
+    {
+        $this->checkPermission(Permission::UsersManage);
+        $user = $this->userRepo->getById($id);
+        // Resetear el 2FA del usuario 
+        $user->mfaValues()->delete();
+        session()->flash('success', trans('settings.users_mfa_reset_success', ['userName' => $user->name]));
+        return redirect()->back();
```
ssddanbrown, line 222: Can we redirect directly to the user edit view?
Redirect backs like this can go wonky in a range of scenarios, so best to specify a location where known.

resources/views/users/edit.blade.php
```diff
@@ -71,6 +71,26 @@ class="button outline">{{ trans('settings.users_mfa_configure') }}</a>
...
+                    <div class="grid half gap-xl v-center">
+                        <div>
+                            <strong class="text-neg">{{ trans('settings.users_mfa_reset') }}</strong>
+                            <p class="text-small text-muted">{{ trans('settings.users_mfa_reset_desc', ['userName' => $user->name]) }}</p>
+                        </div>
+                        <div class="text-m-right">
+                            <form action="{{ url("/settings/users/{$user->id}/reset-mfa") }}" method="POST" style="display: inline;">
+                                @csrf
+                                <button type="submit" class="button neg" 
+                                        onclick="return confirm('{{ trans('settings.users_mfa_reset_confirm', ['userName' => $user->name]) }}')">
```
ssddanbrown, line 86: This won't work due to CSP (security) blocking. We avoid any inline JavaScript.

Making the initial a dropdown, with a second button for confirmation, may be better. We do this in a select few other areas I think.

ssddanbrown (changes requested):
Thanks for offering this!
I've added a range of comments regarding some changes to be made.
We will also need [testing coverage](https://github.com/BookStackApp/BookStack/blob/development/dev/docs/php-testing.md) of the core actions here. If unsure, I can add these later.

Note: If you're heavily using AI/LLM generation for this contribution please let me know, as I don't want to be spending time reviewing and providing feedback to a machine via a proxy.

If anything is unclear or you need further guidance just let me know!

clauvaldez:
Hello!
Thanks for the answers, I am working on the recommendations, it is the first time I make a contribution so there are many details to take into account.
Thank you very much for taking the time, I will work on the details you mentioned and

ssddanbrown:

https://codeberg.org/bookstack/bookstack/pulls/6056
  
ssddanbrown:
I wanted to get this into the next release so have gone ahead and made the changes needed within https://codeberg.org/bookstack/bookstack/commit/39a14cff8f64f51fa3b4be2f4c2a7b85d7c2b600, before merging.

Thanks again @clauvaldez for providing this PR.

Issue #6062, opened 2026-03-18
New revision is created but no changes were made


tortillas5:
### Describe the Bug

When you open a page with images in edit mode and save without making any change, BookStack creates a new version/revision, even tho the page is the same as before.

I guess it's linked to WYSIWYG or page complexity.

### Steps to Reproduce

1. Open a page with images
2. Go into edit mode (in classic WYSIWYG)
3. Don't touch anything, just save
4. Go into revisions
5. You can see a new revision as been created

### Expected Behaviour

No new version/revision should be created since the page did not change.

### Screenshots or Additional Context

It's not always the case, but in long pages I'm able to reproduce it consistently.

### Browser Details

Firefox 148.0.2 (64 bits), windows 11 Build 26200.8037

### Exact BookStack Version

v26.03.1

virtadpt:
Confirmed, been seeing this since I started using Bookstack a few years back.

tortillas5:
@ssddanbrown I fixed it, and tested it locally, hope it's ok for you.
I created a PR for you, I can't assign people on it so I'm pinging you :)

ssddanbrown:
BookStack is now [managed on Codeberg](https://codeberg.org/bookstack/bookstack).

As part of this change, we're closing all existing GitHub issues and pull requests.
You'll instead find this issue on Codeberg here: 
https://codeberg.org/bookstack/bookstack/issues/6062
  
Any further discussion for this will now be done primarily on Codeberg.
Since this thread was created relatively recently, it won't be locked just yet, and we may still provide updates/responses here on GitHub, but it may be locked in the future.

ssddanbrown:
Copying from the Codeberg thread. Please provide responses via Codeberg rather than here.
https://codeberg.org/bookstack/bookstack/issues/6062

---

Did assign this to the next feature release but the scope looks to be bigger than expected so skipping for now.

From scenarios which I could replicate, the issue arises due to the cycle of parsing on both the server/client:

- On save, HTML taken from client editor and sent to server.
- Server makes changes as needed (applying IDs, some other reformating/extracting) before saving.
- On next edit, client editor loads server-altered HTML.
- Client editor than applies its own formatting changes.
    - _Quite often at these stage these are little changes like attribute ordering or whitespace._
- On save, client editor sends altered HTML.
- Server detects as change, creates revision.

Typically the content should even out and become stable after the first non-content-changing save, but there could be scenarios where the two sides continue making changes.

[...]

PR #6063, opened 2026-03-19
Fix: New revision is created but no changes were made #6062

tortillas5:
With this change, if there is no changes on the page, no revision is created.


ssddanbrown:

https://codeberg.org/bookstack/bookstack/pulls/6063
  

Issue #2715, opened 2021-04-27
Mixed Login: LDAP or user registration


pbordon:
Exist the possibility to login in a mixed method: LDAP or simple user registration, defined by user?

ssddanbrown:
Hi @pbordon,
Would you be able to provide insight into the environment where you'd want this within and the benefits this would bring?

pbordon:
In my organization, we have internal users, connected to an AD and external users, which are many and vary over time. Therefore I wanted to implement the login via LDAP for internal users and for external users to register separately.

ssddanbrown:
Updating this to be generic to methods, and merging similar issues into this.

abulgatz:
Any plans to add this to your roadmap or implement this? Internal SSO with guest access seems pretty common.

ssddanbrown:
@abulgatz Probably not anytime too soon, to be totally honest. It's high-risk, low demand, low target audience, high support & maintenance. Therefore it doesn't look worth including at this time.

Fabsky:
I'm in the same case, I mean I've contents for internal users (azure), and content for customers (self register)

mfatfhg:
Hi, I have opened #4401 and because it was closed, I would like to continue the discussion here.

One of the reasons why we would like to see this features was: 

> So, I understand that its not a high level feature request on your roadmap. In our opinion, the ability to authenticate with different types of identities (local db users, oidc, ldap) is somethink like industry standard and should be possible.
> 
> On of the reasons is the following: Typically, OIDC providers are services in the cloud (if you use SAAS IDPs it might be auth0, or if you self host a IDP, it might be located on a remote site of your company).
> 
> If you only allow OIDC at the same time, you cant login to bookstack anymore if you dont have WAN/Internet connectivity anymore. And because we would like to use bookstack as documentation system for emergency manuals too, we would like to have the possibility to login with different types of accounts ( local db accounts or maybe ldap accounts from a local Active directory) as fallback method.
> 
> A admin should always have the possibility to access a system in case of technical problems (bad WAN/Internet connectivity).

This was answered by @ssddanbrown with the following argument:

> Okay. Could always flip the auth method in an emergency. 

This would be like changing the electronic door locks against bearded locks in case of power failure before entering the building.

And changing a config only to access your documentation system is not what you want to do in an emergency situation. And not everyone who need access to the system in such a situation has the ability or possibility to do this.

GitTH:
[...]

PR #6068, opened 2026-03-24
Add configurable multi-auth support for local and OIDC logins

holger-dev:
This PR extends BookStack authentication so multiple login methods can be enabled at the same time, instead of forcing a single global auth method.

The main use case is running OIDC alongside local BookStack accounts. With this change, admins can keep OIDC as the primary login method while still creating internal users with local passwords who can sign in through the normal login form.

Key changes:

- Adds support for AUTH_METHODS as a list of enabled auth methods
- Adds AUTH_PRIMARY_METHOD for preferred UI/redirect behavior
- Keeps backward compatibility with existing single-method setups via AUTH_METHOD
- Updates login/logout flow to track the actual auth method used for the session
- Allows standard/local login to coexist with OIDC, LDAP, or SAML
- Keeps registration, password reset, and password change available whenever standard auth is enabled
- Updates admin user handling so local users can be created without requiring an external_auth_id when standard auth is active
- Hardens install/deploy behavior by clearing stale Laravel package/service cache files before Composer install, avoiding issues with removed dev providers such as Clockwork

Example configuration:
AUTH_METHOD=standard
AUTH_METHODS=standard,oidc
AUTH_PRIMARY_METHOD=oidc
AUTH_AUTO_INITIATE=false

This allows:

- OIDC login for externally managed users
- local BookStack users with password-based login
- admin-managed fallback/internal accounts without external identity mapping


ssddanbrown:
Thanks @holger-dev for offering this PR.

Unfortunately this still isn't something I'd look to maintain/support at this time, for pretty much the same reasons as [I stated on the issue](https://github.com/BookStackApp/BookStack/issues/2715#issuecomment-1418245067), although I appreciate demand has increased since.
This is especially true at this time, after a bunch of recent security releases which have put me on edge (more than usual) when it comes to security critical elements.

That's not to say such a feature won't eventually be added, but I just can't see it being included within the next year or so. Especially as there are some other elements (a revamp to login sessions) which I'd probably like to do before something like this.

Thanks again for offering this code though, but due to the above, I'm going to go ahead and close this off.

holger-dev:
He @ssddanbrown !

Thanks for the reply and for the transparent explanation. 

I understand your security concerns and the current priorities (e.g. login session revamp). That said, I don’t agree with the decision: the code is already running in production, many users need this feature, and the reasoning doesn’t feel sufficient to me. 

I’d appreciate it if you could keep the topic open or suggest an alternative path forward. For us the feature is a game changer in the use of bookstack!

ssddanbrown:
Hi @holger-dev,
I am keeping the original request/topic open (#6068), and I'm not discounting this being added in the future.
It's just that I know I'm not keen to work on this over the next year or so, so I don't see reason to keep it open.

> the code is already running in production

I don't consider this to be a factor when considering what features are supported. 
Implementation effort is commonly the lesser concern.

> many users need this feature

Sure, I don't doubt that, but it's still for a narrow audience relative to maintenance impact right now from my view.
I generally focus on things that will improve the platform for existing users, rather than things that will meet a requirement/need to use BookStack.

 > the reasoning doesn’t feel sufficient to me

Okay, sorry about that, but I have to draw lines with what I'm comfortable with including, based on my experience of maintaining the platform over the last decade, otherwise the same reasons could be used to justify almost any other feature, way beyond what is reasonable maintainable.

Langlais115:
Hi @ssddanbrown,

As I'm using Bookstack in my enterprise, I've been waiting for this feature for many years, which makes me sad that this PR was rejected.

But on the other hand, I understand your point of view regarding the security and maintainability.

I really hope this feature will be integrated in the future.

Keep the good work, Bookstack is marvelous.

Best regards,

holger-dev:
> Hi @ssddanbrown,
> 
> As I'm using Bookstack in my enterprise, I've been waiting for this feature for many years, which makes me sad that this PR was rejected.
> 
> But on the other hand, I understand your point of view regarding the security and maintainability.
> 
> I really hope this feature will be integrated in the future.
> 
> Keep the good work, Bookstack is marvelous.
> 
> Best regards,

[You can use my repo](https://github.com/holger-dev/BookStack) - i try to keep it up to date...

PR #6069, opened 2026-03-25
Fix PDF export heading font fallback

alexwoo-awso:
PDF exports used the heading font var which in DOMPDF fell back to base fonts lacking Polish glyphs, causing ? in titles. This PR adds a PDF-only heading font fallback to DejaVu Sans for h1-h6. Tested by exporting a page with a Polish title; characters render correctly.

Before
image
image

After
image
image

### DIFF
--- a/resources/sass/export-styles.scss
+++ b/resources/sass/export-styles.scss
@@ -64,6 +64,11 @@ body.export-format-pdf {
   font-size: 14px;
   line-height: 1.2;
 
+  // Ensure heading glyph coverage for PDF engines that don't handle CSS vars well.
+  h1, h2, h3, h4, h5, h6 {
+    font-family: 'DejaVu Sans', -apple-system, BlinkMacSystemFont, "Segoe UI", "Oxygen", "Ubuntu", "Roboto", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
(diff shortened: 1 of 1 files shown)


janusz-net:
The problem lies elsewhere. The last version that didn't have a problem with Polish characters in headers was v25.02.2. After updating to v25.02.3, "zażółć gęśli jaźń" -> "za?ół?? g??l? ja??.". The only change that causes this error is at the very beginning of the `public/dist/export-styles.css` file: 
```css
.dummy-style{color:red}:root{--font-body:...
```
Removing `.dummy-style{color:red}` makes Polish characters in PDF headers correct again.

alexwoo-awso:
@janusz-net Removing .dummy-style{color:red} would likely reintroduce the earlier BOM-related issue tracked in https://github.com/sass/dart-sass/issues/472. 

Your observation is still useful though, since it shows the PDF export path is sensitive to the first rule in the inlined export-styles.css; this PR aims to fix that without regressing the earlier workaround.

ssddanbrown:
Thanks @alexwoo-awso, now merged. I did make a slight change afterwards in 4f370ccddb4f5fd8eddff5915a9413786ae2a2c7 to set this for all exports so that it's aligned with what we're doing for fonts on other text.

PR #6071, opened 2026-03-26
Introduce image and CSS CSP controls

### LINKED ISSUE #6033: Introduce image & CSS CSP controls
### Description

As an extra step to harden security.
Compared to JS/iframes, this'll probably be relatively permissive by default to prevent breaking changes, since there's a lower risk for these elements, but we can then have options and guidance on how to tighten rules where desired.

Zhey-on:
## Summary
This PR introduces CSP controls for image and CSS sources.

In line with the issue intent, the defaults are kept relatively permissive to prevent breaking changes on existing instances, while still providing clear options to tighten policies where needed.

## Changes
- Added `img-src` and `style-src` directives to CSP handling.
- Added environment/config options:
  - `ALLOWED_IMAGE_SOURCES`
  - `ALLOWED_CSS_SOURCES`
- Kept permissive defaults when these options are not set, to reduce rollout risk.
- Added tests in `SecurityHeaderTest` for:
  - default behavior
  - custom override behavior
- Added documentation:
  - `.env.example.complete` entries and examples
  - development docs section describing controls and hardening guidance
  - README pointer to the CSP docs section

## Why this approach
The implementation is intentionally conservative by default, especially compared to JS/iframe controls, so instances do not unexpectedly break after upgrade. At the same time, admins can now explicitly restrict allowed sources as part of their hardening process.

## Testing
- `docker compose run --rm app php artisan test tests/SecurityHeaderTest.php --filter="style src|img src|csp"`

Closes #6033.

### DIFF
--- a/.env.example.complete
+++ b/.env.example.complete
@@ -395,6 +395,18 @@ ALLOWED_IFRAME_HOSTS=null
 # Current host and source for the "DRAWIO" setting will be auto-appended to the sources configured.
 ALLOWED_IFRAME_SOURCES="https://*.draw.io https://*.youtube.com https://*.youtube-nocookie.com https://*.vimeo.com"
 
+# A list of sources/hostnames that can be loaded as CSS styles within BookStack.
+# Space separated if multiple. BookStack host domain is auto-inferred.
+# Defaults to a permissive set if not provided.
+# Example: ALLOWED_CSS_SOURCES="https://fonts.googleapis.com"
+ALLOWED_CSS_SOURCES=null
+
+# A list of sources/hostnames that can be loaded as image content within BookStack.
+# Space separated if multiple. BookStack host domain is auto-inferred.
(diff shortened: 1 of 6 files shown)


ssddanbrown (changes requested):
Thanks for offering this @Zhey-on,
Please can you remove the readme and dev/docs changes though. Information on something this specific isn't relevant for these areas. I'll document this CSP use in our docs site as part of the future release changes.

ssddanbrown:

https://codeberg.org/bookstack/bookstack/pulls/6071
  
ssddanbrown:
Thanks again @Zhey-on. I followed this up with https://codeberg.org/bookstack/bookstack/commit/0cd773a5d3d07b1299660c22a3f65456b6aa28d0, and it's now merged to be part of the next release.

Issue #5631, opened 2025-05-28
New WYSIWYG Editor: Beta Testing Feedback


ssddanbrown:
This is a thread specifically to collect feedback regarding issues and limitations of the new WYSIWYG editor, introduced in BookStack v24.10, during its beta phase which starts as of BookStack v25.05. This follows on from thread #5245.

This below list will be continuously updated from feedback to form a full listing to be worked on.
Please only report issues where not already in this list.
Generally our goal is to match the existing TinyMCE based editor as close as possible.

### Known Issues

- [x] Switching editor to new editor with bold text can lead to double-bold. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-2948634591).
- [x] Text formats can carry across to below non-formatted existing lines. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-2951427455)
- Videos/Embeds
  - [x] Can't re-edit embed code. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-2952273371)
  - [x] Can't control dimensions after insert. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-2952273371)
- [x] Inline Highlight format changes text color instead of background. [ref](https://github.com/BookStackApp/BookStack/issues/5715)
- [x] Table cell background color does not retain consistently.
- [x] Add way to focus in new empty paragraph if edit area clicked below a last block consuming element (drawing, table, details). [ref](https://github.com/BookStackApp/BookStack/issues/5716)
- [x] Header formats via keyboard shortcuts are all a size too big. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-3052112532)
- [x] Link protocols are allow-listed, old editor did not seem to allow-list. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-3068881135)
- [x] Source code editor is much smaller than TinyMCE editor. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-3078820079)
- [x] Source code output shows all as a single line, should be multi-line ideally. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-3078820079).
- [x] Inline highlight color should ideally have a different name to background color to result in better translations. [ref](https://github.com/BookStackApp/BookStack/issues/5631#issuecomment-3094381182).
[...]

PR #6119, opened 2026-04-26
Lexical: May 2026 Fixes

ssddanbrown:
- Adds keyCode based fallback for shortcuts to allow them to be used for keyboard layouts such as Cyrillic, where the latin key representation may differ to the actual key value.

### Todo

- [ ] For keyboard shortcuts, convert built-in editor shortcuts to go through our system to work with our fallback system. [Relevant lines](https://github.com/BookStackApp/BookStack/blob/9bf9ae9c371904c770f1ecdbab6c024aec4a9937/resources/js/wysiwyg/lexical/core/LexicalEvents.ts#L1067-L1075)

For #5631


ssddanbrown:
BookStack is now managed on Codeberg.
https://codeberg.org/bookstack/bookstack/pulls/6119
  
Since this thread was created relatively recently, it won't be locked just yet and we may still provide updates/responses here on GitHub, but it may be locked in the future.

